PR #32, opened 2021-10-01
Added tests for IoPathFileLister and IoPathFileLoader

Nayef211:
- Added unit tests for `IoPathFileLister` and `IoPathFileLoader`
- Added decorator to skip tests if `iopath` is not installed
- Note that tests need to be modified once data can be hosted on online storage systems (gdrive/s3/onedrive)

### DIFF
--- a/test/test_datapipe.py
+++ b/test/test_datapipe.py
@@ -785,7 +785,7 @@ class TestDataPipeWithIO(expecttest.TestCase):
         with self.assertRaisesRegex(TypeError, "instance doesn't have valid length"):
             len(xz_reader_dp)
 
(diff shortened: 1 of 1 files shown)


test/test_datapipe.py
```diff
@@ -776,8 +785,25 @@ def test_xz_archive_reader_iterdatapipe(self):
         with self.assertRaisesRegex(TypeError, "instance doesn't have valid length"):
             len(xz_reader_dp)
 
+    # TODO (erjia): this test currently only covers reading from local
```
ejguan, line 788: Super nit: Could you please change it to ejguan as the GH username. 😄

ejguan (approved):
LGTM

PR #35, opened 2021-10-01
Adding option to close stream of ArchiveReaders

NivekT:
* #36
* __->__ #35

Differential Revision: [D31355312](https://our.internmc.facebook.com/intern/diff/D31355312)


NivekT:
@nivekt has imported this pull request.  If you are a Facebook employee, you can view this diff [on Phabricator](https://www.internalfb.com/diff/D31355312).

torchdata/datapipes/iter/util/_archivereader.py
```diff
@@ -0,0 +1,29 @@
...
+from typing import DefaultDict, List, Tuple, Union, IO
+
+from torch.utils.data import IterDataPipe
+
+
+class _ArchiveReaderIterDataPipe(IterDataPipe[Tuple[str, BufferedIOBase]]):
+    def __init__(self):
+        super().__init__()
+        self.open_source_streams: DefaultDict[str, List[Union[BufferedIOBase, IO[bytes]]]] = defaultdict(list)
+        self.open_archive_streams: DefaultDict[str, List[Union[BufferedIOBase, IO[bytes]]]] = defaultdict(list)
```
ejguan, line 13: This has another potential issue that `open_streams` are appended inside each iterator. What if two iterators are constructed, there would be duplicated filename. If I only want to close the opened streams inside one iterator, all the streams are closed.

ejguan: Is there any way for us to check if the stream has been depleted?

NivekT: I can't think of a good way without trying call `read()` or something similar directly. There is also the possibility of the user rewinding the stream if they wanted to.

Therefore, I prefer to give users the ability to close the stream rather than detecting if the stream has ended and close it for them, especially true for `FileLoader` streams that is used by multiple DataPipes.

ejguan: Could we rely on https://docs.python.org/3/library/gc.html#gc.callbacks?

ejguan: gc should know the reference for the opened file handle is out of scope. The warning came from gc when the file is opened and not used.
If we have a way to register the opened file handle to gc to run close for us, we may be able to suppress the warning.

But, we need to make sure the file handle registered to gc don't increment the ref_count. Otherwise, the opened file handle would never be closed even with the callback.

torchdata/datapipes/iter/util/_archivereader.py
```diff
@@ -0,0 +1,21 @@
...
+        super().__init__()
+        self._open_source_streams: DefaultDict[str, List[Union[BufferedIOBase, IO[bytes]]]] = defaultdict(list)
+
+    def _close_all_source_streams(self):
+        for _name, streams_list in self._open_source_streams.items():
+            for stream in streams_list:
+                stream.close()
+        self._open_source_streams = defaultdict(list)
+
+    def __del__(self):
```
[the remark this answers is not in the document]
NivekT: Also, we can store them in a `List` instead of a `Dict` as well.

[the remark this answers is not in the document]
ejguan: Then, users still need to explicitly `del datapipe` with this API.

I was thinking for each stream yielded from `__iter__` can be wrapped by a Wrapper class like:
```py
class StreamWrapper:
    def __init__(self, fd):
        self.fd = fd

    def __getattr__(self, name):  # Delegate methods to fd
        return getattr(self.fd, name)

    def __del__(self):
        self.fd.close()
```
We could yield stream wrapper here: https://github.com/pytorch/pytorch/blob/d85948896cde158012ed758879be88a2afdc29d7/torch/utils/data/datapipes/utils/common.py#L55
```py
for pathname in pathnames:
    yield pathname, StreamWrapper(open(pathname, mode))
```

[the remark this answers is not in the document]
ejguan: Do you think this would work?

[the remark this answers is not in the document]
NivekT: I think `__del__` is called when `gc` is invoked as well.

PR #40, opened 2021-10-04
[WIP] Adding a test that creates an arbitrarily large .tar file and read from it via HTTP

NivekT:
* __->__ #40

The goal here is to create an arbitrarily large .tar file and read from it via HTTP.

Currently, it seems to work if we cache the file first and then read from it. However, an error is raised when we attempt to read directly from the `HTTPReader` because its stream does not support the operation`seek`:
```
  File "/Users/ktse/data/test/test_stream.py", line 66, in <module>
    for fname, stream in tar_dp:
  File "/Users/ktse/data/torchdata/datapipes/iter/util/tararchivereader.py", line 62, in __iter__
    raise e
  File "/Users/ktse/data/torchdata/datapipes/iter/util/tararchivereader.py", line 48, in __iter__
    tar = tarfile.open(fileobj=cast(Optional[IO[bytes]], data_stream), mode=self.mode)
  File "/Users/ktse/miniconda3/envs/pytorch/lib/python3.9/tarfile.py", line 1609, in open
    saved_pos = fileobj.tell()
io.UnsupportedOperation: seek
```


test/test_stream.py
```diff
@@ -0,0 +1,112 @@
...
+        self.temp_dir.cleanup()
+        print("Tear down has completed")
+
+    def running_server(self):
+        httpd = socketserver.TCPServer(("", self.port), handler_from(self.temp_dir_path))
+        print(f"Starting TCP server on port {self.port}...")
+        httpd.serve_forever()
+        while True:
+            if self.stop_server:  # TODO: This is not closing
+                httpd.server_close()
```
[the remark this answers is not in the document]
ejguan: You should use thread event to terminate the loop.

test/test_stream.py
```diff
@@ -0,0 +1,112 @@
...
+    def setUp(self) -> None:
+        # Temporary Directory abd Server Setup
+        print("\nSetup is running...")
+        self.temp_dir = tempfile.TemporaryDirectory()  # noqa: P201
+        self.temp_dir_path = self.temp_dir.name
+        self.port = 8006
+        self.stop_server = False
+        self.server_thread = threading.Thread(
+            target=self.running_server
+        )  # TestStream.start_test_server(self.temp_dir_path, self.port)
```
ejguan, line 37: This is not going to work. You need to pass these variables to thread during construction.

test/test_stream.py
```diff
@@ -0,0 +1,112 @@
...
+        self.server_thread.start()
+
+    def tearDown(self) -> None:
+        print("Tear down is running...")
+        self.stop_server = True
```
ejguan, line 42: Trigger threading event here.

test/test_stream.py
```diff
@@ -0,0 +1,112 @@
...
+            return "test_tar.tar.xz", "w:xz", "r|xz"
+        else:
+            raise ValueError(f"file_type must be 'tar', 'gz', or 'xz', but instead found {file_type}.")
+
+    @staticmethod
+    def add_to_tar_archive(file_path, write_mode: str, file_names: List[str]) -> None:
+        print("Adding files to Tar archive...")
+        with tarfile.open(file_path, write_mode) as tar:
+            for fname in file_names:
+                tar.add(fname)
```
ejguan, line 85: You may need to specify `arcname` for `tar.add` to override the file name, because `tar.add` would preserve the hierarchy of each file. Let's say your archive is `/tmp/foo/1.tar`, and file is `/tmp/foo/1.csv`. If you don't specify the `arcname`, the file is going to be `/tmp/foo/1.tar/tmp/foo/1.cdv` when it's extracted from tar.

test/test_stream.py
NivekT: It seems to work for now since `running_server` has access to `self` variable. But I will keep an eye out to see if there is any bug related to this.

NivekT: But I think it would be better to refactor it and take those as arguments as you suggested.

NivekT:
No longer need it since we have other benchmark and remote testing

PR #51, opened 2021-10-08
Change HTTPReader to use requests module and enable streaming, minor change to test case

NivekT:
* (to be filled)

### DIFF
--- a/torchdata/datapipes/utils/common.py
+++ b/torchdata/datapipes/utils/common.py
@@ -6,6 +6,7 @@ import warnings
 
 from io import BufferedIOBase
 from typing import Iterable, List, Tuple, Union
+from urllib3.response import HTTPResponse
 
 
 def match_masks(name: str, masks: Union[str, List[str]]) -> bool:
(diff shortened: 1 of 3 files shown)


torchdata/datapipes/utils/common.py
```diff
@@ -63,14 +64,14 @@ def _default_filepath_fn(data):
     return os.path.normpath(data)
 
 
-def validate_pathname_binary_tuple(data: Tuple[str, BufferedIOBase]):
+def validate_pathname_binary_tuple(data: Tuple[str, Union[BufferedIOBase, HTTPResponse]]):
```
ejguan, line 67: These two classes are sharing `io.IOBase`. Is there any concern that we have to use union of these two types?

ejguan:
BTW, could you please also add a corresponding PR to PyTorch Core?

NivekT:
> BTW, could you please also add a corresponding PR to PyTorch Core?

Sure. In general, do we still intend to keep the duplicate code within PyTorch Core up-to-date even though they are deprecated?

PR #59, opened 2021-10-14
Add fakedata and split tests

ejguan:
* #61
* #65
* #60
* __->__ #59

Split tests into three files based on the utilities.
Add _fakedata. Then, we can rely on GH to test remote accessing folders and tar files.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -6,6 +6,8 @@ on:
   pull_request:
     branches:
       - main
+      # For PR created by ghstack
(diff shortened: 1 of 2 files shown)


ejguan:
> > Super annoying for this PR. `ghimport` doesn't support binary files. It has to be normal PR without using ghstack.
> 
> You can switch the base branch to `main` here and hit `Import to fbsource`?

Unfortunately, it doesn't work. Thanks for the advice anyway.

PR #67, opened 2021-10-18
Fakedata

ejguan:
Replace https://github.com/pytorch/data/pull/59 because `ghimport` doesn't work with binary files.


NivekT (approved):
LGTM

PR #72, opened 2021-10-20
Fix TorchText examples

ejguan:
* #74
* #73
* __->__ #72

There is a bug that `HTTPResponse` doesn't have correct `read` method. We need to use `__iter__` to get data. Then, the bytes are also needed to be decoded as string.

cc: @Nayef211

Differential Revision: [D31797766](https://our.internmc.facebook.com/intern/diff/D31797766)


ejguan:
@ejguan has imported this pull request.  If you are a Facebook employee, you can view this diff [on Phabricator](https://www.internalfb.com/diff/D31797766).

PR #93, opened 2021-10-28
Closing streams for DataPipe that doesn't yield stream

### LINKED ISSUE #86: Close all streams within DataPipe after reading (if DataPipe's output isn't a stream)
## 🚀 Feature

Close all streams within DataPipe after reading (if DataPipe's output isn't a stream). This is applicable to DataPipes such as `CSVParserIterDataPipe` and `JsonParserIterDataPipe`.

## Motivation

This features prevents streams that has been exhausted from remaining open. Since the end of the stream has been reached, the user must reset or reopen the stream before they can be reused. It makes sense to close them and the users can re-open them outside the DataPipe if desired.

## Alternatives

One alternative is to only close streams that are not seek-able. The users still have to re-open/reset those streams outside of the DataPipe.

## Additional context

Note that if the DataPipe returns a stream (e.g. `TarArchiveReader`), the behavior will be different (it won't be closed) because it is uncertain when that output stream will be read.

cc @ejguan @VitalyFedyunin

NivekT:
* __->__ #93

Fixes #86 

The goal of this change is to close opened streams within `DataPipes` that read from streams but do not yield streams. #86 describes the issue in more details.

Aside from these ones, I don't believe any other `IterDataPipes` require this change.

Differential Revision: [D32152883](https://our.internmc.facebook.com/intern/diff/D32152883)


NivekT:
@nivekt has imported this pull request.  If you are a Facebook employee, you can view this diff [on Phabricator](https://www.internalfb.com/diff/D32152883).

ejguan (approved):
LGTM. Wondering does our StreamWrapper guarantee each file would be closed in such cases?

NivekT:
@nivekt has imported this pull request.  If you are a Facebook employee, you can view this diff [on Phabricator](https://www.internalfb.com/diff/D32152883).

NivekT:
> LGTM. Wondering does our StreamWrapper guarantee each file would be closed in such cases?

If the previous DataPipe in the chain produces `StreamWrapper`, it should get closed at some point (but not immediately). This covers the case when `StreamWrapper` wasn't used and free up the resource sooner. Moreover, I believe `IOStream` can be closed multiple times without issues.

PR #94, opened 2021-10-29
Refactor IoPath FileLister and FileLoader

### LINKED ISSUE #52: Use IoPath to download HTTP Url
Since IoPath has features to download different sources: HTTP, S3, Google, and etc, we should add a `IoPathDownloader` to handle all different sources.

ejguan:
Fixes #52

This PR is going to add all existing `PathHandler`s from `iopath` to make sure it can handle the following data sources:
- Local File Path
- HTTP Url
- OneDrive Url
- S3 Url (will be included in `iopath` next release)

Before this PR, these two DataPipes only work with `NativePathHandler`, which would handle the local file.

### DIFF
--- a/torchdata/datapipes/iter/load/iopath.py
+++ b/torchdata/datapipes/iter/load/iopath.py
@@ -1,72 +1,107 @@
 # Copyright (c) Facebook, Inc. and its affiliates.
 import os
 
+from typing import Iterator, List, Tuple, Union
+
+from torch.utils.data.datapipes.utils.common import match_masks
+
 from torchdata.datapipes import functional_datapipe
 from torchdata.datapipes.iter import IterDataPipe
 from torchdata.datapipes.utils import StreamWrapper
-from typing import Iterator, Tuple
+
+try:
+    import iopath
+
+except ImportError:
+    iopath = None
+
+
+def _create_default_pathmanager():
+    from iopath.common.file_io import HTTPURLHandler, OneDrivePathHandler, PathManager
+    pathmgr = PathManager()
+    pathmgr.register_handler(HTTPURLHandler(), allow_override=True)
+    pathmgr.register_handler(OneDrivePathHandler(), allow_override=True)
+    # S3PathHandler is not included in 0.1.8
+    try:
+        from iopath.common.s3 import S3PathHandler
+
+        pathmgr.register_handler(S3PathHandler(), allow_override=True)
+    except ImportError:
+        pass
+    return pathmgr
 
 
 class IoPathFileListerIterDataPipe(IterDataPipe[str]):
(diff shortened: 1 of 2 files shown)


torchdata/datapipes/iter/load/iopath.py
```diff
@@ -1,73 +1,108 @@
...
                 "Package `iopath` is required to be installed to use this "
-                "datapipe. Please use `pip install iopath` or `conda install "
-                "iopath`"
-                "to install the package"
+                "datapipe. Please use `pip install iopath` to install the package"
             )
 
         self.root: str = root
-        self.pathmgr = g_pathmgr
+        self.pathmgr = _create_default_pathmanager()
```
NivekT, line 61: Should we give users the option to pass in custom `PathManager` as an argument with None as the default?

Then if `None`, call `_create_default_pathmanager()`

torchdata/datapipes/iter/load/iopath.py
```diff
@@ -6,30 +6,43 @@
...
-        try:
-            from iopath.common.file_io import g_pathmgr
-        except ImportError:
+        if PathManagerFactory is None:
             raise ModuleNotFoundError(
                 "Package `iopath` is required to be installed to use this "
-                "datapipe. Please use `pip install iopath` or `conda install "
-                "iopath`"
-                "to install the package"
+                "datapipe. Please use `pip install iopath` to install the package"
```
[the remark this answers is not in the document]
NivekT: I am out of the loop but what is the issue with conda-forge?

torchdata/datapipes/iter/load/iopath.py
```diff
@@ -1,73 +1,108 @@
...
             )
 
         self.root: str = root
-        self.pathmgr = g_pathmgr
+        self.pathmgr = _create_default_pathmanager()
+        self.masks = masks
+
+    def register_handler(self, handler, allow_override=False):
```
NivekT, line 64: Do the existing test cases cover these change? Or do we plan to have a separate PR to write a complete set of test for these DataPipes?

torchdata/datapipes/iter/load/iopath.py
ejguan: Makes sense

NivekT (commented):
Have some questions but overall looks good

torchdata/datapipes/iter/load/iopath.py
ejguan: It requires to have remote data sources. Can't really any test for now.

ejguan: We only have test for local files.

ejguan:
Re-create PR to use ghstack.

PR #98, opened 2021-11-05
Improve ci

ejguan:
- Remove `mypy` for tests
- Add `3.8`
- Remove requirements

### DIFF
--- a/mypy.ini
+++ b/mypy.ini
@@ -6,11 +6,9 @@ show_column_numbers = True
 check_untyped_defs = True
 pretty = True
 
-files =
-    torchdata,
-    test
+files = torchdata
 
-exclude = examples
+exclude = examples, test
 
 # Minimum version supported - variable annotations were introduced
 # in Python 3.6
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,3 +1,1 @@
-expecttest
 requests
-torch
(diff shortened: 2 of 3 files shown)


mypy.ini
```diff
@@ -6,11 +6,9 @@ show_column_numbers = True
 check_untyped_defs = True
 pretty = True
 
-files =
-    torchdata,
-    test
+files = torchdata
 
-exclude = examples
+exclude = examples, test
```
NivekT, line 11: That means `mypy` won't scan through any test files right?

ejguan: Yeah. It's way too hard to maintain `mypy` for tests if we want to do some hacks.
Actually, that part is redundant as I only set `files = torchdata`. None of other folds will carry out mypy.

requirements.txt
```diff
@@ -1,3 +1 @@
-expecttest
 requests
-torch
```
NivekT, line 3: I know we are still installing PyTorch nightly for testing with CI. Does the removal of `torch` from `requirements.txt` impact anything else such as installation?

mypy.ini
ejguan: Second thought, let me keep the `exclude` here to make everything explicitly.

requirements.txt
ejguan: It won't because we explicitly have `torch` here https://github.com/pytorch/data/blob/012d8f279e313e4b21a387cc9bf10449ce7b3f4e/setup.py#L55

NivekT (approved):
LGTM

PR #99, opened 2021-11-05
Fix bug for GDriveReader

ejguan:
`self.source_datapip` -> `self.source_datapipe`

### DIFF
--- a/torchdata/datapipes/iter/load/online.py
+++ b/torchdata/datapipes/iter/load/online.py
@@ -84,9 +84,10 @@ class GDriveReaderDataPipe(IterDataPipe[Tuple[str, StreamWrapper]]):
     Args:
         source_datapipe: a DataPipe that contains URLs to GDrive files
     """
+    source_datapipe: IterDataPipe[str]
 
     def __init__(self, source_datapipe: IterDataPipe[str]) -> None:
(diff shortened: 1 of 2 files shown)


torchdata/datapipes/iter/load/online.py
```diff
@@ -84,9 +84,10 @@ class GDriveReaderDataPipe(IterDataPipe[Tuple[str, StreamWrapper]]):
     Args:
         source_datapipe: a DataPipe that contains URLs to GDrive files
     """
+    source_datapipe: IterDataPipe[str]
 
     def __init__(self, source_datapipe: IterDataPipe[str]) -> None:
-        self.source_datapip: IterDataPipe[str] = source_datapipe
```
ejguan, line 89: Lol

ejguan:
Failing tests are caused by limited times to access Google Drive.

torchdata/datapipes/iter/load/online.py
NivekT: It has an error because an 'e' is missing?

NivekT (approved):
LGTM

torchdata/datapipes/iter/load/online.py
ejguan: Yeah

PR #102, opened 2021-11-08
Draft of example integration with Lhotse

pzelasko:
As discussed with @ejguan, I'm uploading my example code for integration of Lhotse and DataPipes.

Some context: Lhotse is a library for working with audio/speech data. It has a common representation (manifest) format called  `Cut` and `CutSet`, and recipes for generating these manifests for 30+ popular speech corpora. Lhotse provides Dataset and Sampler classes that simplify working with speech data. 

The reason I'm interested in DataPipes is that I find the "old" Dataset/Sampler setup lacks some flexibility that is helpful in speech/audio. This PR shows how to:
- take Lhotse manifests (lightweight meta-data objects), 
- construct a metadata-level batch with a dynamic size (`DurationBatcher`: code copied from one of Lhotse's samplers),
- perform I/O for each batch element in parallel
- recombine the batch manifests + tensors into a single batch again

What is missing is: collation, transforms, etc. -- these should be simple to add later.

I'd like to get your feedback on this usage of DataPipes. I enjoyed writing this example and I think Lhotse could gain a lot by adopting this type of data pipeline.

### DIFF
--- a/examples/audio/lhotse_example.py
+++ b/examples/audio/lhotse_example.py
@@ -1,0 +1,235 @@
+#!/usr/bin/env python
+
+"""
+This is a working example of using a new PyTorch dataloading API called DataPipes
+with Lhotse. It's not the final version and probably a lot of things will change.
+
+Lhotse is a speech data representation library that brings various corpora into
+a single format and interfaces well with PyTorch. See: https://github.com/lhotse-speech/lhotse
+
+Prepare the env like this first:
+
+   $ conda create -n torchdata python=3.8
+   $ conda activate torchdata
+   $ conda install numpy
+   $ pip install --pre torch -f https://download.pytorch.org/whl/nightly/cpu/torch_nightly.html
+   $ pip install --user "git+https://github.com/pytorch/data.git"
+
+Optionally, if not in this Lhotse branch already:
+
+   $ pip install git+https://github.com/lhotse-speech/lhotse@feature/datapipes-prototyping
+
+"""
+
+
+import warnings
+from collections import deque, defaultdict
+from functools import partial
+from pathlib import Path
+from typing import Optional
+
+from lhotse import CutSet, load_manifest
+from lhotse.utils import Seconds
+from lhotse.recipes import download_librispeech, prepare_librispeech
+
+import torchdata
+import torch
+import torchdata.datapipes.iter as dp  # Only use Iter-style DataPipe to illustrate the pipeline
+
+from torch.utils.data.datapipes.iter import Multiplexer, Demultiplexer
+from torch.utils.data.communication.eventloop import SpawnProcessForDataPipeline
+from torch.utils.data.communication.protocol import IterDataPipeQueueProtocolClient
+from torch.utils.data.communication.iter import QueueWrapper
+
+
+class CutsReader(dp.IterDataPipe):
+    def __init__(self, path) -> None:
+        self.cuts = CutSet.from_jsonl_lazy(path)
+
+    def __iter__(self):
+        yield from self.cuts
+
+
+class DurationBatcher(dp.IterDataPipe):
+    def __init__(
+        self,
+        datapipe,
+        max_frames: int = None,
+        max_samples: int = None,
+        max_duration: Seconds = None,
+        max_cuts: Optional[int] = None,
+        drop_last: bool = False,
+    ):
+        from lhotse.dataset.sampling.base import SamplingDiagnostics, TimeConstraint
+
+        self.datapipe = datapipe
+        self.reuse_cuts_buffer = deque()
+        self.drop_last = drop_last
+        self.max_cuts = max_cuts
+        self.diagnostics = SamplingDiagnostics()
+        self.time_constraint = TimeConstraint(
+            max_duration=max_duration, max_frames=max_frames, max_samples=max_samples
+        )
+
+    def __iter__(self):
+        self.datapipe = iter(self.datapipe)
+        while True:
+            yield self._collect_batch()
+
+    def _collect_batch(self):
+        self.time_constraint.reset()
+        cuts = []
+        while True:
+            # Check that we have not reached the end of the dataset.
+            try:
+                if self.reuse_cuts_buffer:
+                    next_cut = self.reuse_cuts_buffer.popleft()
+                else:
+                    # If this doesn't raise (typical case), it's not the end: keep processing.
+                    next_cut = next(self.datapipe)
+            except StopIteration:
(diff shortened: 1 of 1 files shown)


examples/audio/lhotse_example.py
```diff
@@ -0,0 +1,235 @@
...
+        self.reuse_cuts_buffer = deque()
+        self.drop_last = drop_last
+        self.max_cuts = max_cuts
+        self.diagnostics = SamplingDiagnostics()
+        self.time_constraint = TimeConstraint(
+            max_duration=max_duration, max_frames=max_frames, max_samples=max_samples
+        )
+
+    def __iter__(self):
+        self.datapipe = iter(self.datapipe)
```
ejguan, line 75: Please do not remove the reference to prior/source datapipe here. Since we want to let `DataLoaderV2` have `graph` mode execution of data pipeline (run parallel, automatic sharding, etc.), we need to keep reference graph of each DataPipe instance.

examples/audio/lhotse_example.py
```diff
@@ -0,0 +1,235 @@
...
+        while True:
+            yield self._collect_batch()
```
ejguan, line 77: We should reset the `iter` to None, because generator instance is not serializable (pickable).
IIUC, this is going to iterate forever. Is this intentionally?

ejguan: Simply speaking.
```suggestion
    def __iter__(self):
        self.datapipe_iter = iter(self.datapipe)
        ...
        self.datapipe_iter = None
```

examples/audio/lhotse_example.py
ejguan: This would also prevent you do second iteration.

examples/audio/lhotse_example.py
```diff
@@ -0,0 +1,235 @@
...
+        return CutSet.from_cuts(cuts)
+
+
+class IODataPipe(dp.IterDataPipe):
+    def __init__(self, datapipe):
+        self.datapipe = datapipe
+
+    def __iter__(self):
+        for cut_idx, batch_idx, batch_size, cut in self.datapipe:
+            yield cut_idx, batch_idx, batch_size, cut.load_audio(), cut
```
ejguan, line 143: Could you use `Mapper` to replace this DataPipe?
You should be able to use functional API as:
```py
source_datapipe # Referring to self.datapipe
source_datapipe.map(fn=lambda x: x.load_audio(), input_col=3, output_col=-1)
```
The result is going to be `cut_idx, batch_idx, batch_size, cut, cut.load_audio()`

examples/audio/lhotse_example.py
```diff
@@ -0,0 +1,235 @@
...
+
+
+class RecombineBatchAfterIO(dp.IterDataPipe):
```
ejguan, line 146: This is an interesting use case.  It's better to have a buffer size IMO. Otherwise, it potentially blows up memory.

examples/audio/lhotse_example.py
```diff
@@ -0,0 +1,235 @@
...
+
+    # The following Multiprocessing API may change, and multiprocessing should be added and handled by DataLoader
+    # demux and mux would be called by DataLoader. But, I want to explicitly call them to illustrate the functionality
+
+    # Routed prior datapipe to ``num_jobs`` datapipes in round-robin manner
+    datapipes = datapipe.demux(
+        num_jobs, classifier_fn=partial(classifier_fn, num_jobs=num_jobs)
+    )
+    for i in range(len(datapipes)):
+        datapipes[i] = IODataPipe(datapipes[i])
```
ejguan, line 197: Thank you so much. This is a great example to show the functionality to distribute extensive operation to multiprocessing.

examples/audio/lhotse_example.py
```diff
@@ -0,0 +1,235 @@
...
+
+
+if __name__ == "__main__":
+    cuts_path = Path(
+        "./workspace/LibriSpeech/manifests/mini-libri-cuts-train-clean-5.jsonl.gz"
+    )
+
+    if not cuts_path.exists():
+        print("Downloading Mini LibriSpeech.")
+        download_librispeech("./workspace", dataset_parts="mini_librispeech")
```
ejguan, line 223: This could potentially replaced by the `on_disk_cache` [WIP for refactoring] with download and hash check.

ejguan (commented):
Thank you so much for providing early feedback on this project. This is extremely useful.
A couple of comments below.

examples/audio/lhotse_example.py
pzelasko: Thanks. DataLoaderV2 sounds interesting -- is there a description somewhere that I could read to learn more?

examples/audio/lhotse_example.py
pzelasko: > IIUC, this is going to iterate forever. Is this intentionally?

This should stop once `self._collect_batch()` raises `StopIteration`. Maybe the sampling code in `DurationBatcher` can be made simpler -- as you can see there is a lot of branching. I'll give it more thought at a later point.

examples/audio/lhotse_example.py
pzelasko: This looks interesting. Will the lambda work well with pickling / multiprocessing though?

pzelasko: Also: is it possible to work with dicts rather than tuples when using the functional API? E.g.:

```python
yield {
  'cut_idx': cut_idx,
  'batch_idx': batch_idx,
  ...
}
```

examples/audio/lhotse_example.py
pzelasko: How do you envision it? Let's say buffer size is 500 items but for some reason, I sampled a batch that has 600 items. Wouldn't that hang indefinitely?

I might not understand the new dataloading machinery yet -- my assumption was that there is some mechanism to make it prepare a `prefetch_factor` number of batches, and if the consumer stops fetching them (e.g., because it's training), then the whole pipeline would "stop" and wait. If it doesn't stop, I can see how the memory might blow up though.

examples/audio/lhotse_example.py
ejguan: > Will the lambda work well with pickling / multiprocessing though?

We want to support `lambda` as it's a common feature for functional programming. So, as long as `dill` is installed, we would let multiprocessing with lambda works.

> Also: is it possible to work with dicts rather than tuples when using the functional API?

Yes. It's supposed to work if you give `input_col='cut_idx'`.

PR #121, opened 2021-12-10
Autogenerated issues for TODO lines

VitalyFedyunin:


ejguan:
There are duplicated autogenerated Issues, like
- https://github.com/pytorch/data/issues/129
- https://github.com/pytorch/data/issues/136
The link from the former one is also somehow incorrect.

PR #145, opened 2021-12-21
Upstreaming S3 datapipes

ydaiming:
Fixes #{issue number}

### Changes

  - Added `S3FileLister` and `S3FileLoader` IterDataPipes.
  - Added pybind11 build for s3 io cpp files and python scripts.

### TODO

- [ ] clean up setup files and link pybind11 in CMAKE_PREFIX automatically.
- [ ] remove aws cpp sdk dependency at build & pop exceptions when missing dependencies at usage.
- [ ] new api changes for `list_files`.
- [ ] expose timeouts, regions, credential settings.
- [ ] benchmarks
- [ ] thorough tests

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,0 +1,43 @@
+cmake_minimum_required(VERSION 3.13)
+project(_torchdata)
+
+set(CMAKE_CXX_STANDARD 17)
+SET(TORCH_MIN_VERSION "1.5.1")
+
+find_package(Python3 COMPONENTS Interpreter Development)
+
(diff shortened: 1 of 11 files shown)


chauhang:
@ydaiming Thanks for this contribution. Please add a readme describing how user's can use this feature.

yqhu:
Thanks @ydaiming.  It seems the definition of `test_s3_lister_iterdatapipe()` in `test_remote_io.py` is indented one level too much so it is not run during tests.  Also `file_url = "s3://pt-s3plugin-test-data-west2/images/test"` is inaccessible to me; if I change it to `file_url = "s3://ai2-public-datasets/charades"` the test runs fine for me.

PR #153, opened 2021-12-30
Adding timeout option for GDriveReader

### LINKED ISSUE #132: [TODO] can this also have a timeout?
https://github.com/pytorch/data/blob/f102d25f9f444de3380c6d49bf7aaf52c213bb1f/build/lib/torchdata/datapipes/iter/load/online.py#L113

### LINKED ISSUE #137: [TODO] Add timeout to GDrive part of OnlineReaderIterDataPipe
https://github.com/pytorch/data/blob/f102d25f9f444de3380c6d49bf7aaf52c213bb1f/torchdata/datapipes/iter/load/online.py#L120

NivekT:
* __->__ #153

Fixes #132 
Fixes #137

Differential Revision: [D33370399](https://our.internmc.facebook.com/intern/diff/D33370399)


NivekT:
@nivekt has imported this pull request.  If you are a Facebook employee, you can view this diff [on Phabricator](https://www.internalfb.com/diff/D33370399).

torchdata/datapipes/iter/load/online.py
```diff
@@ -49,30 +49,36 @@ def __len__(self) -> int:
...
-        raise RuntimeError("Internal error: headers don't contain content-disposition.")
-
-    filename = re.findall('filename="(.+)"', response.headers["content-disposition"])
-    if filename is None:
-        raise RuntimeError("Filename could not be autodetected")
+    with requests.Session() as session:
+        if timeout is None:
+            response = session.get(url, stream=True)
+        else:
+            response = session.get(url, timeout=timeout, stream=True)
```
ejguan, line 58: I believe the default value for `timeout` is None. So, no need to do if-else.

torchdata/datapipes/iter/load/online.py
```diff
@@ -49,30 +49,36 @@ def __len__(self) -> int:
...
-
-    response = session.get(url, stream=True)
-
-    if "content-disposition" not in response.headers:
-        raise RuntimeError("Internal error: headers don't contain content-disposition.")
-
-    filename = re.findall('filename="(.+)"', response.headers["content-disposition"])
-    if filename is None:
-        raise RuntimeError("Filename could not be autodetected")
+    with requests.Session() as session:
```
ejguan, line 54: Quick question here. As we do streaming session here, could using `with` context manager would destroy the response before we actually read data.

Let's add the label `ciflow/slow` to test them.

ejguan (commented):
Looks good in general. I do have two questions below.

torchdata/datapipes/iter/load/online.py
NivekT: I think you are right, I would think the `with` statement will close the session before the `response.raw` byte stream is fully read. I will work on testing it out.

NivekT: @ejguan The test seems fine. Do you think we should still remove the `with` statements? I believe `response.raw` should be a byte stream but I'm unsure if still has a network connection and if it does, it will be terminated by the closing of `session` (I think it should).

NivekT:
@nivekt has imported this pull request.  If you are a Facebook employee, you can view this diff [on Phabricator](https://www.internalfb.com/diff/D33370399).

torchdata/datapipes/iter/load/online.py
ejguan: I made a mistake, the corresponding label is `ciflow/period` as https://github.com/pytorch/data/blob/7899c86a12613748d14a4fbb352edfb3409afa46/test/test_period.py#L13 is in `test_period.py` file.

Besides, could we add `reset` tests for gdrive and other online datapipe?

NivekT: I think the existing reset test should be testing the behavior?

First, the functional test builds an iterator and reads the first line from `GDriveReader` to confirm it has content.
https://github.com/pytorch/data/blob/7899c86a12613748d14a4fbb352edfb3409afa46/test/test_period.py#L20-L25

After that, we call `HashChecker` on `GDriveReader` and check the full hash of the downloaded file.

https://github.com/pytorch/data/blob/7899c86a12613748d14a4fbb352edfb3409afa46/test/test_period.py#L27-L32

As part of `HashChecker` I believe this line here should reset the `source_datapipe` (equivalent to creating a new iterator) so the content will be read from the beginning again?

https://github.com/pytorch/data/blob/7899c86a12613748d14a4fbb352edfb3409afa46/torchdata/datapipes/iter/util/hashchecker.py#L49


Therefore, the behavior is that one line from `GDriveReader` was read, it gets reset and the full hash is checked. Does that make sense or am I still missing something?

I also just tried running this and the test still passes:
```python
check_cache_dp = gdrive_reader_dp.check_hash({expected_file_name: expected_MD5_hash}, "md5", rewind=False)
it = iter(check_cache_dp)
path, stream = next(it)
self.assertEqual(expected_file_name, os.path.basename(path))
self.assertTrue(io.BufferedReader, type(stream))
it = iter(check_cache_dp)
path, stream = next(it)
self.assertEqual(expected_file_name, os.path.basename(path))
self.assertTrue(io.BufferedReader, type(stream))
```

NivekT:
@nivekt has imported this pull request.  If you are a Facebook employee, you can view this diff [on Phabricator](https://www.internalfb.com/diff/D33370399).

torchdata/datapipes/iter/load/online.py
ejguan: Thank! Interesting use case. I will approve the PR as it's still working with the context manager

ejguan (approved):
Thanks, LGTM

PR #165, opened 2022-01-12
S3 datapipes

ydaiming:
### Changes
- Added S3FileLister and S3FileLoader IterDataPipes.
- Added pybind11 build for s3 io cpp files and python scripts.
### TODO
- [x] clean up setup files and link pybind11 in CMAKE_PREFIX automatically.
- [x] remove `aws-cpp-sdk` dependency at build with `BUILD_S3` env var & pop exceptions when missing dependencies at usage.
- [x] new api changes for list_files.
- [x] clean up cpp files (naming, new structure, new logic etc.)
- [x] expose timeouts, regions.
- [x] thorough tests
  - [x] different correct usage: bucket (with or without / at last), folder (with or without / at last), prefix, item.
  - [x] different incorrect usage: non-existing files, wrong s3 urls, etc.
  - [x] region changes
  - [x] choice of public datasets
- [x] benchmarks
  - [x] performance test
- [x] README.md
  - [x] user guide & recommendations
  - [x] dependencies


josephevans:
Can we get this into the torch 1.11.0 release?

setup.py
```diff
@@ -70,4 +94,12 @@ def _export_version(version, sha):
     # Package Info
     packages=find_packages(exclude=["test*", "examples*"]),
     zip_safe=False,
+    extras_require={
+        "scipy": ["scipy"],
+    },
```
ejguan, line 99: Do we need `scipy` here?

ejguan (commented):
Could you please add `ninja cmake` into `pip install` on our CI here https://github.com/pytorch/data/blob/e53afb723de8ce761c0fabc8c8f5c066e83cb50a/.github/workflows/ci.yml#L43-L46?

ejguan:
> Can we get this into the torch 1.11.0 release?

TorchData is going to do a release after torch 1.11 (March maybe). If this PR is landed before that, we can squeeze it into TorchData release

setup.py
ydaiming: I removed this dependency. It was copied from S3 Plugin initially, and I don't think it's required in S3 IO.

ydaiming:
Note that, to use S3 IO, we'll need pybind11 and "aws-sdk-cpp:s3;transfer" installed. We also need the BUILD_S3 env var set to "ON" to trigger build.

ejguan:
> Note that, to build torchdata with S3 IO, we'll need pybind11 and "aws-sdk-cpp:s3;transfer" installed. We also need the BUILD_S3 env var set to "ON" to trigger build.

If so, we need to add `pip install pybind11` as well to run CI correctly. But, I am not sure how we are gonna install "aws-sdk-cpp:s3;transfer"?

ydaiming:
> how we are gonna install "aws-sdk-cpp:s3;transfer"

I've written a README on how to install dependencies and build and use S3 IO datapipes. [torchdata/datapipes/iter/load/README.md](https://github.com/ydaiming/data/blob/s3-datapipes/torchdata/datapipes/iter/load/README.md)

ejguan:
> git clone --recurse-submodules https://github.com/aws/aws-sdk-cpp
> cd aws-sdk-cpp/
> mkdir sdk-build
> cd sdk-build
> cmake .. -DCMAKE_BUILD_TYPE=Release -DBUILD_ONLY="s3;transfer"
> make
> make install # may need sudo

Add a separate step before building `torchdata` should be fine. https://github.com/pytorch/data/blob/e53afb723de8ce761c0fabc8c8f5c066e83cb50a/.github/workflows/ci.yml#L47-L48

ydaiming:
> > git clone --recurse-submodules https://github.com/aws/aws-sdk-cpp
> > cd aws-sdk-cpp/
> > mkdir sdk-build
> > cd sdk-build
> > cmake .. -DCMAKE_BUILD_TYPE=Release -DBUILD_ONLY="s3;transfer"
> > make
> > make install # may need sudo
> 
> Add a separate step before building `torchdata` should be fine.
> 
> https://github.com/pytorch/data/blob/e53afb723de8ce761c0fabc8c8f5c066e83cb50a/.github/workflows/ci.yml#L47-L48

Added another step to install aws-sdk-cpp.

I wonder why install test requirements before building instead of after building and before tests?

ejguan:
> Added another step to install aws-sdk-cpp.
> 
> I wonder why install test requirements before building instead of after building and before tests?

You need to compile dependency before building TorchData, am I right? Except, you want to do a more fancy feature to compile aws and provide S3 IO whenever such DataPipe is used. (I think it's doable but needs to do some research, we should land this PR first then figure it out).

.github/workflows/ci.yml
```diff
@@ -44,9 +44,21 @@ jobs:
...
+          git clone --recurse-submodules https://github.com/aws/aws-sdk-cpp
+          cd aws-sdk-cpp/
+          mkdir sdk-build
+          cd sdk-build
+          cmake .. -DCMAKE_BUILD_TYPE=Release -DBUILD_ONLY="s3;transfer"
+          make
+          make install
       - name: Build TorchData
+        env: BUILD_S3=1
         run: python setup.py develop
```
ejguan, line 62:
```suggestion
        run: python setup.py develop
        env:
          BUILD_S3=1
```

ydaiming:
> > Added another step to install aws-sdk-cpp.
> > I wonder why install test requirements before building instead of after building and before tests?
> 
> You need to compile dependency before building TorchData, am I right?

I meant that I noticed the following step (already existing in ci) is before building torchdata:
```
        run: pip3 install expecttest fsspec iopath==0.1.9 numpy pytest rarfile
```
Should we move it to after building torchdata and before testing?

ydaiming:
> CMake Error: CMake was unable to find a build program corresponding to "Ninja".  CMAKE_MAKE_PROGRAM is not set.  You probably need to select a different build tool.
> CMake Error: CMAKE_C_COMPILER not set, after EnableLanguage
> CMake Error: CMAKE_CXX_COMPILER not set, after EnableLanguage

Looks like the CI still doesn't install `ninja` before installing `torchdata`. Any idea? @ejguan?

.github/workflows/ci.yml
```diff
@@ -44,10 +44,23 @@ jobs:
         run: |
           pip3 install -r requirements.txt
           pip3 install --pre torch -f https://download.pytorch.org/whl/nightly/cpu/torch_nightly.html
+          pip3 install cmake ninja
+          pip3 install pybind11
       - name: Install test requirements
```
ejguan, line 49: You may want to add a step to put `ninja` in the `PATH`. 
```
- name: xxxx
  run: echo "/home/runner/.local/bin" >> $GITHUB_PATH
  shell: bash
```

CMakeLists.txt
```diff
@@ -0,0 +1,61 @@
+cmake_minimum_required(VERSION 3.18 FATAL_ERROR)
```
VitalyFedyunin, line 1: Any specific reasons to require 3.18 ? (it is not default in Ubuntu/Debian LTS)

VitalyFedyunin:
Without SDK installed any usage of S3 pipes fails with 
```
#17 [13/13] RUN python test.py
#17 sha256:a0ceb2208da477f7ae261e437253fd7fe4763562b03c7bcce76b1d46b57163b5
#17 1.951 /usr/local/lib/python3.8/dist-packages/torchdata-0.3.0a0+2627793-py3.8-linux-x86_64.egg/torchdata/_extension.py:30: UserWarning: torchdata C++ extension is not available.
#17 1.951   warnings.warn("torchdata C++ extension is not available.")
#17 1.952 Traceback (most recent call last):
#17 1.952   File "test.py", line 8, in <module>
#17 1.952     s3_lister_dp = S3FileLister(IterableWrapper(input), region="us-west-2")
#17 1.952   File "/usr/local/lib/python3.8/dist-packages/torchdata-0.3.0a0+2627793-py3.8-linux-x86_64.egg/torchdata/datapipes/iter/load/s3io.py", line 29, in __init__
#17 1.952     self.handler = torchdata._torchdata.S3Handler(request_timeout_ms, region)
#17 1.952 AttributeError: module 'torchdata' has no attribute '_torchdata'
#17 ERROR: executor failed running [/bin/sh -c python test.py]: exit code: 1
```

VitalyFedyunin:
Should be nice error instead

CMakeLists.txt
ydaiming: Not much reasons.
[For PyTorch, it does require 3.10](https://github.com/pytorch/pytorch/blob/master/CMakeLists.txt#L1), while having a series of logic to deal with different possible versions. We may set it to 3.13, the minimum version that allows ninja.

VitalyFedyunin: I think 3.13 should be fine as I have seen 3.16 as default versions (AFAIR).

josephevans:
> Should be nice error instead

Hi @VitalyFedyunin, would you recommend we raise a `ModuleNotFoundError` exception, similar to what is done [here](https://github.com/pytorch/data/blob/main/torchdata/datapipes/iter/load/iopath.py#L60-L64)?

Something like this?

```
--- a/torchdata/datapipes/iter/load/s3io.py
+++ b/torchdata/datapipes/iter/load/s3io.py
@@ -24,6 +24,11 @@ class S3FileListerIterDataPipe(IterDataPipe[str]):
     """

     def __init__(self, source_datapipe: IterDataPipe[str], length: int = -1, request_timeout_ms=-1, region="") -> None:
+        if not hasattr(torchdata, '_torchdata'):
+            raise ModuleNotFoundError(
+                "Torchdata must be built with BUILD_S3=1 to use this datapipe."
+            )
+
         self.source_datapipe: IterDataPipe[str] = source_datapipe
         self.length: int = length
         self.handler = torchdata._torchdata.S3Handler(request_timeout_ms, region)
```

CMakeLists.txt
josephevans: Updated to 3.13.

VitalyFedyunin (commented):
Code looks good, we would need to check final binary size and what happens if we install python egg on the system without AWS library.

ydaiming:
@ejguan , let's make the mergins process faster and catch PyTorch 1.11.0 release. 

The current issues are:
1. Windows requires a different set of commands to install aws-cpp-sdk (the final step also requires administrative privileges). How should we do that?
2. The style lint is failing for weird reasons un-related to torchdata, and mypy lint is complaining about the inevitable `torchdata._torchdata` usage. Shall we set to ignore the `mypy` lint with some `mypy.ini` settings?
3. Building fails for MacOS py37/38, Ubuntu py37, which I can't reproduce. I tried to build in Ubuntu py37 env, and my build and test both succeeded. Could this be a resource issue?
4. The test failures for the remaining MacOS py39, Ubuntu py38/39 are also weird, because I tested locally and the ValueError should be raised to pass the tests. I changed it to BaseException to see if it catches the error.

Please help us speed up the process, as we all want to include S3 IO datapipes in the coming torchdata release. Thank you!

.github/workflows/ci.yml
```diff
@@ -46,10 +46,23 @@ jobs:
...
         run: pip3 install expecttest fsspec iopath==0.1.9 numpy pytest rarfile
+      - name: Install AWS-SDK-CPP for S3 IO datapipes
+        run: |
+          git clone --recurse-submodules https://github.com/aws/aws-sdk-cpp
+          cd aws-sdk-cpp/
+          mkdir sdk-build
+          cd sdk-build
+          cmake .. -DCMAKE_BUILD_TYPE=Release -DBUILD_ONLY="s3;transfer"
+          make
+          sudo make install
```
ejguan, line 61: Add `shell: bash` here

.github/workflows/ci.yml
```diff
@@ -35,7 +35,7 @@ jobs:
         run: |
           sudo add-apt-repository multiverse
           sudo apt update
-          sudo apt install rar unrar
+          sudo apt install rar unrar libssl-dev libcurl4-openssl-dev zlib1g-dev
       - name: Setup Python ${{ matrix.python-version }}
         uses: actions/setup-python@v2
         with:
```
ejguan, line 41: In order to let `cmake` to find the right Python version, we may need to add right Python liib to `$PATH`

ejguan: Or, you can follow this instruction https://github.com/actions/setup-python/issues/186#issuecomment-1010828971

ydaiming:
@ejguan Could you please trigger the CI again to check if the mypy error is cleared, and the correct python versions are found at building? Thanks.

ydaiming:
> > The test failures for the remaining MacOS py39, Ubuntu py38/39 are also weird, because I tested locally and the ValueError should be raised to pass the tests. I changed it to BaseException to see if it catches the error.
>
> Can I have a code pointer which Error you are expecting?

The logic for error is:
1. passing in the input url (incorrect in this test) to list the files:
https://github.com/ydaiming/data/blob/s3-datapipes/torchdata/csrc/pybind/S3Handler/pybind.cpp#L22
https://github.com/ydaiming/data/blob/s3-datapipes/torchdata/csrc/pybind/S3Handler/S3Handler.cpp#L431
2. when parsing the incorrect input, `invalid_argument` will be thrown.
https://github.com/ydaiming/data/blob/s3-datapipes/torchdata/csrc/pybind/S3Handler/S3Handler.cpp#L435
https://github.com/ydaiming/data/blob/s3-datapipes/torchdata/csrc/pybind/S3Handler/S3Handler.cpp#L139
3. Then `pybind11` will throw `ValueError` on the PyThon side.
https://pybind11.readthedocs.io/en/stable/advanced/exceptions.html

The tests passed in my environment, I'm not sure what's the different in the CI system.

.github/workflows/ci.yml
ydaiming: > Or, you can follow this instruction [actions/setup-python#186 (comment)](https://github.com/actions/setup-python/issues/186#issuecomment-1010828971)

Added.

ydaiming:
> > * Windows requires a different set of commands to install aws-cpp-sdk (the final step also requires administrative privileges). How should we do that?
> 
> What do you mean different set of commands? Can you list the command you want to use? IIRC, you can specify `shell: bash` in the workflow to run the command
> 

For Ubuntu and MacOS, the commands are as following, and already added:
```bash
          git clone --recurse-submodules https://github.com/aws/aws-sdk-cpp
          cd aws-sdk-cpp/
          mkdir sdk-build
          cd sdk-build
          cmake .. -DCMAKE_BUILD_TYPE=Release -DBUILD_ONLY="s3;transfer"
          make
          make install
```
while on Windows, the following commands are used:
```bash
git clone --recurse-submodules https://github.com/aws/aws-sdk-cpp
mkdir sdk_build
cd sdk_build
cmake ..\aws-sdk-cpp -DCMAKE_BUILD_TYPE=Release -DBUILD_ONLY="s3:transfer"
"C:\Program Files (x86)\Microsoft Visual Studio\2019\Community\MSBuild\Current\Bin\MSBuild.exe" ALL_BUILD.vcxproj  -p:Configuration=Release # depending on where MSBuild is
msbuild INSTALL.vcxproj -p:Configuration=Release # with administrative privilege 
```

reference: https://docs.aws.amazon.com/sdk-for-cpp/v1/developer-guide/setup-windows.html

mypy.ini
```diff
@@ -12,6 +12,13 @@ exclude = examples, test
 
 python_version = 3.7
 
+#
+# Extension modules without stubs.
+#
+
+[mypy-torchdata._torchdata]
```
ydaiming, line 19: Should this be in a different format? @ejguan

.github/workflows/ci.yml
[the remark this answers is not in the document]
ydaiming: Got it, will apply

mypy.ini
ejguan: My bad. Let me push a commit to your branch.

ydaiming: Are you working on pushing right now?

For a quick fix on python version, starting from CMake 3.19, we can specify versions directly. https://cmake.org/cmake/help/latest/command/find_package.html#id5
> The [version] argument requests a version with which the package found should be compatible. There are two possible forms in which it may be specified:
> 
> A single version with the format major[.minor[.patch[.tweak]]], where each component is a numeric value.

.github/workflows/ci.yml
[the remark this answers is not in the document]
ydaiming: Edited and pushed with separate windows installation commands for AWS-SDK-CPP

ydaiming:
My bad, looks like `sudo` is still required. I checked early commits and saw successful builds. Could you re-trigger the ci again? If possible, we can stop the ci for commit fe63369.

.github/workflows/ci.yml
```diff
@@ -46,10 +46,33 @@ jobs:
         run: |
           pip3 install -r requirements.txt
           pip3 install --pre torch -f https://download.pytorch.org/whl/nightly/cpu/torch_nightly.html
+          pip3 install cmake ninja pybind11 pybind11-global
+          echo "/home/runner/.local/bin" >> $GITHUB_PATH
       - name: Install test requirements
         run: pip3 install expecttest fsspec iopath==0.1.9 numpy pytest rarfile
+      - name: Install AWS-SDK-CPP on Windows for S3 IO datapipes
+        if: ${{ matrix.os }} == 'windows-latest'
```
ejguan, line 54:
```suggestion
if: matrix.os == 'windows-latest'
```
I never get used to the syntax of github actions.

ydaiming: Edited and pushed

ydaiming:
The correct python version isn't picked still. Any other ideas?

ydaiming:
@ejguan My bad, the test for incorrect filenames should take `input` not `file_urls`. The test should pass now.

And if the test passes now, the only thing remaining is about the CI.

ejguan:
For windows, you may add another step to install msbuild via https://github.com/microsoft/setup-msbuild

ydaiming:
> For windows, you may add another step to install msbuild via https://github.com/microsoft/setup-msbuild

Added the step. The last step for installation still requires administrative privilege. We may need a way for that.

Also, the ubuntu py38/39 are passing. We've got us a solid checkpoint.

ejguan:
@ydaiming Could you send me your email via slack then I will see how to let you run the CI without my manual approval.

ydaiming:
@ejguan Could you trigger the CI now to see if the runtime Python is actually indicated as the same for setup-python? Thanks.

tools/setup_helpers/extension.py
```diff
@@ -0,0 +1,106 @@
...
+
+        cmake_args = [
+            f"-DCMAKE_BUILD_TYPE={cfg}",  # not used on MSVC, but no harm
+            f"-DCMAKE_PREFIX_PATH={torch.utils.cmake_prefix_path}",
+            f"-DCMAKE_INSTALL_PREFIX={extdir}",
+            f"-DCMAKE_LIBRARY_OUTPUT_DIRECTORY={extdir}",
+            f"-DPYTHON_EXECUTABLE={sys.executable}",
+            "-DCMAKE_CXX_FLAGS=-fPIC",
+            f"-DBUILD_S3:BOOL={'ON' if _BUILD_S3 else 'OFF'}",
+        ]
```
ejguan, line 74: Could we directly provide `Python_EXECUTABLE`, `Python_INCLUDE_DIRS`, `Python_LIBRARIES` here to bypass `find_package(python3)`

ydaiming: Thanks. That's what I'm trying as well. Just pushed a commit, check if the paths are correct.

ydaiming: I was also looking at this from pybind11, but somehow the `PYTHON_EXECUTABLE` wasn't used in the CMake. Maybe we don't even need the find_package(Python3) at all.

ejguan: Yeah. It's probably true that `PYTHON_EXECUTABLE` is not used as executable won't be used for compilation.

torchdata/datapipes/iter/load/s3io.py
```diff
@@ -0,0 +1,87 @@
...
+
+    Args:
+        source_datapipe: a DataPipe that contains URLs/URL prefixes to s3 files
+        length: Nominal length of the datapipe
+        requestTimeoutMs: optional, overwrite the default timeout setting for this datapipe
+        region: optional, overwrite the default region inferred from credentials for this datapipe
+
+    Note:
+        AWS_CPP_SDK is necessary to use the S3 DataPipe(s).
+    """
```
NivekT, line 24:
```suggestion
    r"""
    Iterable DataPipe that lists URLs with the given prefixes (functional name: ``list_file_by_s3``).

    Args:
        source_datapipe: a DataPipe that contains URLs/URL prefixes to s3 files
        length: Nominal length of the datapipe
        requestTimeoutMs: optional, overwrite the default timeout setting for this datapipe
        region: optional, overwrite the default region inferred from credentials for this datapipe

    Note:
        AWS_CPP_SDK is necessary to use the S3 DataPipe(s).
     
    Example:
        >>> # Add usage example here
    """
```

Minor comment on docstring to ensure the rendered documentation will look consistent.

NivekT (changes requested):

torchdata/datapipes/iter/load/s3io.py
ydaiming: Updated docstring for both functions with functional names, and updated examples

ydaiming:
@ejguan Looks like 6 Domain tests failed in [MacOS, Windows] x [py37, 38, 39]. Samples can't be drawn for [eurosat]. Is this issue related to my S3 IO PR in any way?

.github/workflows/ci.yml
```diff
@@ -35,21 +35,52 @@ jobs:
...
+          mkdir sdk-build
+          cd sdk-build
+          cmake .. -DCMAKE_BUILD_TYPE=Release -DBUILD_ONLY="s3;transfer" -DAUTORUN_UNIT_TESTS=OFF
+          make
+          sudo make install
       - name: Build TorchData
         run: python setup.py develop
+        env:
+          BUILD_S3: 1
+          BUILD_PYTHON_VERSION: ${{ matrix.python-version }}
```
ejguan, line 83: Could we add test when `BUILD_S3` was disabled?

ejguan (commented):
I don't think the Error from domain is caused by your PR.

.github/workflows/ci.yml
ydaiming: I think I've added a test for missing `_torchdata_` as well as for missing `aws-sdk-cpp`.

To run those tests, should we build two binaries? and also have two separate test stages?

ejguan: What you can do is to add another matrix option here https://github.com/pytorch/data/blob/19cf4530084820c54264141f22b82ba0e2997cfd/.github/workflows/ci.yml#L23-L31
```
with-s3:
  - 1
  - 0
```
Then, using this option to determine if we want to build with S3.
```
env:
    BUILD_S3: ${{ matrix.with-s3 }}
```

ydaiming: updated the CI for a new matrix variable for BUILD_S3, and related tests have passed.

.github/workflows/ci.yml
```diff
@@ -29,27 +29,61 @@ jobs:
           - 3.7
           - 3.8
           - 3.9
+        with-s3:
+          - 1
+          - 0
```
ejguan, line 34: What if AWS SDK was not installed but we compile torchdata with `BUILD_S3=1`?

ejguan (approved):
In general, looks good to me with one comment on the expected behavior.

.github/workflows/ci.yml
ydaiming: Compiling with `BUILD_S3=1` and without AWS SDK will result in compilation failure.

torchdata/csrc/CMakeLists.txt
```diff
@@ -0,0 +1,19 @@
+if(BUILD_S3)
+  message(STATUS "Building S3 IO functionality")
+  find_package(AWSSDK REQUIRED COMPONENTS s3 transfer)
```
ejguan, line 3: I was asking the question because we may need to specify the path for AWSSDK here.

ejguan: I really hope AWSSDK can provide static library for users to install rather than installing from source code.

.github/workflows/ci.yml
```diff
@@ -29,27 +29,61 @@ jobs:
...
+      - name: Install AWS-SDK-CPP on Non-Windows for S3 IO datapipes
+        if: matrix.with-s3 == 1 && matrix.os != 'windows-latest'
+        run: |
+          git clone --recurse-submodules https://github.com/aws/aws-sdk-cpp
+          cd aws-sdk-cpp/
+          mkdir sdk-build
+          cd sdk-build
+          cmake .. -DCMAKE_BUILD_TYPE=Release -DBUILD_ONLY="s3;transfer" -DAUTORUN_UNIT_TESTS=OFF
+          make
+          sudo make install
```
ejguan, line 81: Another follow up in my mind, does AWS provide docker images for such SDK then we don't have to download / compile it all the time?

torchdata/csrc/CMakeLists.txt
ydaiming: I'm looking into building AWS SDK CPP static libraries, and doing some experiments now. Therefore, if official `torchdata` is built with S3 IO functionality, users can just install `torchdata` directly.

.github/workflows/ci.yml
ydaiming: I don't think there's docker image for aws-sdk-cpp. It's too large and can be built and installed in very customized way on different systems.

torchdata/csrc/CMakeLists.txt
ejguan: Do you mind adding another environment variable like `S3_PATH` to cmake. And use it as the `PATH` for `find_package`.
And, please also do a rebase. Then, I think this PR is good to go.

ydaiming: Where would you like the `S3_PATH` pointing? 
`aws-sdk-cpp` will build and install as shared/dynamically-linked libraries by default, and the libraries are put at default places. If the anyone cares to build and install in a custom manner, the developer should know to add the path to env `PATH`.

ydaiming: I just did a test with static libraries. Adding `-DBUILD_SHARED_LIBS=OFF` when cmaking `aws-sdk-cpp` allows `torchdata` to include static libraries when building the binary. The S3 IO datapipes in the new binary work on my instance without aws-sdk-cpp installed and with the nightly-torch.

ydaiming: Rebasing now.

VitalyFedyunin: What is the size footprint in the case of statical linking?

ydaiming: @VitalyFedyunin the size will be from 128K to 2.0M with static libraries.

ydaiming:
@ejguan I rebased my branch based on the main branch, which causes a diverge (non-fast-forward). I will need to create a new branch and a new PR https://github.com/pytorch/data/pull/318 for the rebased code.

torchdata/csrc/CMakeLists.txt
VitalyFedyunin: Awesome, that means we can prepackage it into the official distro as soon as we are manylinux compatible,

ejguan:
> @ejguan I rebased my branch based on the main branch, which causes a diverge (non-fast-forward). I will need to create a new branch and a new PR #318 for the rebased code.

I did rebase for you and forced update your branch. LMK if anything wrong with the rebase with your change.

NivekT (approved):

ydaiming:
@ejguan
Thanks for rebasing the branch! A small suggestion is removing the `setup.cfg` which looks redundantly added.

torchdata/_extension.py
```diff
@@ -0,0 +1,42 @@
...
+        warnings.warn("torchdata C++ extension is not available.")
+        return
+
+    _load_lib("libtorchdata")
+    try:
+        # This import is for initializing the methods registered via PyBind11
+        # This has to happen after the base library is loaded
+        from torchdata import _torchdata  # noqa
+    except ImportError as e:
+        warnings.warn(f"torchdata C++ extension unable to load: {e}")
```
ejguan, line 39: Can we remove this part? This warning can be annoying for users who don't want to use pybind.

ydaiming: Updated

tools/setup_helpers/extension.py
```diff
@@ -0,0 +1,109 @@
...
+
+
+def get_ext_modules():
+    if _BUILD_S3:
+        return [Extension(name="torchdata._torchdata", sources=[])]
+    else:
+        return []
+
+
+# Based off of pybiind cmake_example
```
ejguan, line 45: nit: pybind

setup.py
```diff
@@ -85,5 +106,11 @@ def _export_version(version, sha):
         # Package Info
         packages=find_packages(exclude=["test*", "examples*"]),
```
ejguan, line 107: Could you add `"tools*"` in `exclude`?

ejguan (commented):
Could you fix lint?

torchdata/_extension.py
```diff
@@ -0,0 +1,45 @@
+import os
+import warnings
+from pathlib import Path
+
+import torch
+from torchdata._internal import module_utils as _mod_utils  # noqa: F401
+
+_LIB_DIR = Path(__file__).parent / "lib"
```
ejguan, line 8: Removing `"lib"` here may solve the issue. But, it's kind interesting that why it's working on linux/macos.

tools/setup_helpers/extension.py
```diff
@@ -0,0 +1,109 @@
...
+        extdir = os.path.abspath(os.path.dirname(self.get_ext_fullpath(ext.name)))
+
+        # required for auto-detection of auxiliary "native" libs
+        if not extdir.endswith(os.path.sep):
+            extdir += os.path.sep
+
+        debug = int(os.environ.get("DEBUG", 0)) if self.debug is None else self.debug
+        cfg = "Debug" if debug else "Release"
+
+        sdk_dir = "C:\\Program Files (x86)\\aws-cpp-sdk-all"
```
ejguan, line 67: Wait. This seems super buggy.

torchdata/_extension.py
```diff
@@ -0,0 +1,63 @@
...
+    import os
+
+    import torch
+
+    # load the custom_op_library and register the custom ops
+    lib_dir = os.path.dirname(__file__)
+    loader_details = (importlib.machinery.ExtensionFileLoader, importlib.machinery.EXTENSION_SUFFIXES)
+
+    extfinder = importlib.machinery.FileFinder(lib_dir, loader_details)
+    ext_specs = extfinder.find_spec("_torchdata")
```
ejguan, line 56: We are trying to load `_torchdata` here. But, except `_torchdata.so` or `_torchdata.pyd` compiled files. We do have a folder called `_torchdata`. It can be confusing for `importlib`

torchdata/_torchdata/__init__.pyi
```diff
@@ -0,0 +1,12 @@
...
+from typing import List, Tuple
+
+# TODO: Add pyi generate script
+class S3Handler:
+    def __init__(self, request_timeout_ms: int, region: str) -> None: ...
+    def s3_read(self, url: str) -> bytes: ...
+    def list_files(self, prefix: str) -> List[str]: ...
+    def set_buffer_size(self, buffer_size: int) -> None: ...
+    def set_multi_part_download(self, multi_part_download: bool) -> None: ...
+    def clear_marker(self) -> None: ...
```
ejguan, line 12: Let's remove the pyi file and `_torchdata` folder for now to see if this would help `python` find the library

ydaiming: Removed

torchdata/_extension.py
```diff
@@ -0,0 +1,63 @@
...
+    if ext_specs is None:
+        raise ImportError("torchdata C++ Extension is not found.")
```
ejguan, line 58: For `BUILD_S3=0`, we should ignore this part.

ydaiming: Yes, we shouldn't. But I think we should first focus on get the things work when BUILD_S3=1, then disable these errors when BUILD_S3=0?

setup.py
```diff
@@ -85,5 +106,11 @@ def _export_version(version, sha):
...
         zip_safe=False,
+        # C++ Extension Modules
+        ext_modules=setup_helpers.get_ext_modules(),
+        cmdclass={
+            "build_ext": setup_helpers.CMakeBuild,
+            "clean": clean,
+        },
     )
     gen_pyi()
```
ejguan, line 116: Let's remove `gen_pyi` and import.

torchdata/_torchdata/__init__.pyi
ejguan: And, for the pyi generation, this should happen during `cmake` time. This feature can be placed into a separate PR.

torchdata/_extension.py
```diff
@@ -0,0 +1,63 @@
...
+    torch.ops.load_library(ext_specs.origin)
+    torch.classes.load_library(ext_specs.origin)
```
ejguan, line 60: Adding this part into https://github.com/pytorch/vision/blob/9f48c6274f66f71dd0cabe6a48bde84c8f239abe/torchvision/extension.py#L77-L88

ydaiming: added, please review.

torchdata/datapipes/iter/load/README.md
```diff
@@ -0,0 +1,76 @@
...
+
+```bash
+git clone --recurse-submodules https://github.com/aws/aws-sdk-cpp
+cd aws-sdk-cpp/
+mkdir sdk-build
+cd sdk-build
+cmake .. -DCMAKE_BUILD_TYPE=Release -DBUILD_ONLY="s3;transfer"
+make
+make install # may need sudo
+```
```
ejguan, line 16: We need to add a few comments here to tell users to build static library on Windows.

torchdata/datapipes/iter/load/s3io.py
```diff
@@ -0,0 +1,107 @@
...
+        self.length: int = length
+        self.handler = torchdata._torchdata.S3Handler(request_timeout_ms, region)
+
+    def __iter__(self) -> Iterator[str]:
+        for prefix in self.source_datapipe:
+            while True:
+                urls = self.handler.list_files(prefix)
+                yield from urls
+                if not urls:
+                    break
```
ejguan, line 50: Please correct me if I am wrong. I am not sure how this `while` loop exits. When `urls` becomes None?
Does that mean `handler` understand what `urls` have been yielded?

ejguan (commented):
I have looked through this PR again. I don't see the problem. @VitalyFedyunin Do you mind taking an extra look at this PR

torchdata/datapipes/iter/load/s3io.py
ydaiming: Yes, when `urls` becomes None, the loop exits. `handler` outputs at most 1000 items each time, and has a marker on where to continue in the next loop, until all items given the prefix are yieled.

torchdata/datapipes/iter/load/README.md
```diff
@@ -0,0 +1,76 @@
...
+
+`S3FileLister` accepts a list of S3 prefixes and iterates all matching s3 urls. The functional API is `list_file_by_s3`.
+Acceptable prefixes include `s3://bucket-name`, `s3://bucket-name/`, `s3://bucket-name/folder`,
+`s3://bucket-name/folder/`, and `s3://bucket-name/prefix`. You may also set `length`, `request_timeout_ms` (default 3000
+ms in aws-sdk-cpp), and `region`. Note that:
+
+1. Input **must** be a list and direct S3 URLs are skipped.
+2. `length` is `-1` by default, and any call to `__len__()` is invalid, because the length is unknown until all files
+   are iterated.
+3. `request_timeout_ms` and `region` will overwrite settings in the configuration file or environment variables.
```
ejguan, line 46: Could you add this part of note into the inline-doc of S3FileLister? Then these doc can be shown in https://pytorch.org/data

torchdata/datapipes/iter/load/README.md
```diff
@@ -0,0 +1,76 @@
...
+
+### S3FileLoader
+
+`S3FileLoader` accepts a list of S3 URLs and iterates all files in `BytesIO` format with `(url, BytesIO)` tuples. The
+functional API is `load_file_by_s3`. You may also set `request_timeout_ms` (default 3000 ms in aws-sdk-cpp), `region`,
+`buffer_size` (default 120Mb), and `multi_part_download` (default to use multi-part downloading). Note that:
+
+1. Input **must** be a list and S3 URLs must be valid.
+2. `request_timeout_ms` and `region` will overwrite settings in the configuration file or environment variables.
```
ejguan, line 55: Ditto

torchdata/datapipes/iter/load/s3io.py
ejguan: I see. It makes sense, thanks for explaining.

torchdata/datapipes/iter/load/README.md
ydaiming: What's the format to do so? Append to the Note section of the datapipes?

ejguan: LIke https://github.com/pytorch/data/blob/71a16a3b63966af0f60af74d3044b3c963cfc5fb/torchdata/datapipes/iter/util/combining.py#L14-L46

Put it in the in-line doc

ydaiming: updated in the lates commit, please review.

torchdata/datapipes/iter/load/README.md
ydaiming: updated in the lates commit, please review.

torchdata/datapipes/iter/load/README.md
ydaiming: Addressed in the latest commit.

VitalyFedyunin (approved):

PR #166, opened 2022-01-14
Fixing HeaderIterDP's __len__ function

### LINKED ISSUE #123: [TODO] __len__ Test: returns the length of source when it is less than the limit
https://github.com/pytorch/data/blob/f102d25f9f444de3380c6d49bf7aaf52c213bb1f/test/test_datapipe.py#L301

### LINKED ISSUE #134: [TODO] Fix the case that the length of source_datapipe is shorter than limit
https://github.com/pytorch/data/blob/f102d25f9f444de3380c6d49bf7aaf52c213bb1f/torchdata/datapipes/iter/util/header.py#L29

NivekT:
* __->__ #166

The previous implementation simply return the limit as the length for `HeaderIterDataPipe`. This updated implementation takes it a step further to account for the different possible scenarios. 

Fixes #123
Fixes #134

Differential Revision: [D33589168](https://our.internmc.facebook.com/intern/diff/D33589168)

### DIFF
--- a/test/test_datapipe.py
+++ b/test/test_datapipe.py
@@ -307,9 +307,24 @@ class TestDataPipe(expecttest.TestCase):
         # __len__ Test: returns the limit when it is less than the length of source
         self.assertEqual(5, len(header_dp))
 
-        # TODO(123): __len__ Test: returns the length of source when it is less than the limit
-        # header_dp = source_dp.header(30)
-        # self.assertEqual(20, len(header_dp))
+        # __len__ Test: returns the length of source when it is less than the limit
+        header_dp = source_dp.header(30)
+        self.assertEqual(20, len(header_dp))
+
+        # __len__ Test: returns limit if source doesn't have length
+        source_dp_NoLen = IDP_NoLen(list(range(20)))
+        header_dp = source_dp_NoLen.header(30)
+        with warnings.catch_warnings(record=True) as wa:
+            self.assertEqual(30, len(header_dp))
+            self.assertEqual(len(wa), 1)
+            self.assertRegex(
+                str(wa[0].message), r"length of this HeaderIterDataPipe is inferred to be equal to its limit"
+            )
+
+        # __len__ Test: returns limit if source doesn't have length, but it has been iterated through once
+        for _ in header_dp:
+            pass
+        self.assertEqual(20, len(header_dp))
 
     def test_enumerator_iterdatapipe(self) -> None:
         letters = "abcde"
--- a/torchdata/datapipes/iter/util/header.py
+++ b/torchdata/datapipes/iter/util/header.py
@@ -1,5 +1,6 @@
 # Copyright (c) Facebook, Inc. and its affiliates.
 from typing import Iterator, TypeVar
+from warnings import warn
 
 from torchdata.datapipes import functional_datapipe
 from torchdata.datapipes.iter import IterDataPipe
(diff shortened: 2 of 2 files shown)


torchdata/datapipes/iter/util/header.py
```diff
@@ -20,14 +21,26 @@ class HeaderIterDataPipe(IterDataPipe[T_co]):
...
     def __len__(self) -> int:
-        return self.limit
+        if self.length != -1:
+            return self.length
+        try:
+            source_len = len(self.source_datapipe)
+            self.length = min(source_len, self.limit)
+            return self.length
+        except TypeError:
+            warn(
```
NivekT, line 42: Is raising a warning and returning the best guess (i.e. `self.limit`) the right decision here? It can be triggered by things like `list(header_dp)` if `header_dp` hasn't been fully traversed once yet, which might confuse the users. The alternative is to raise an exception and not return anything.

test/test_datapipe.py
```diff
@@ -307,9 +307,24 @@ def test_header_iterdatapipe(self) -> None:
         # __len__ Test: returns the limit when it is less than the length of source
         self.assertEqual(5, len(header_dp))
 
-        # TODO(123): __len__ Test: returns the length of source when it is less than the limit
-        # header_dp = source_dp.header(30)
-        # self.assertEqual(20, len(header_dp))
+        # __len__ Test: returns the length of source when it is less than the limit
```
NivekT, line 310: These four tests here show what the expected behavior will be

torchdata/datapipes/iter/util/header.py
```diff
@@ -20,14 +21,26 @@ class HeaderIterDataPipe(IterDataPipe[T_co]):
...
         self.limit: int = limit
+        self.length = -1
 
     def __iter__(self) -> Iterator[T_co]:
         for i, value in enumerate(self.source_datapipe):
             if i < self.limit:
                 yield value
             else:
                 break
+        self.length = min(i + 1, self.limit)  # We know length with certainty when we reach here
```
NivekT, line 32: `i + 1` for the case when `source_datapipe` has a length smaller than the limit, so the loop exits with `i = len(source_datapipe) - 1`.

ejguan: There is a minor bug that `i` would be unavailable if `self.source_datapipe` is empty. Let's declare a variable before the iteration to represent `i`.

torchdata/datapipes/iter/util/header.py
ejguan: This looks good to me.

ejguan (approved):
LGTM except one nit comment.

NivekT:
@nivekt has imported this pull request.  If you are a Facebook employee, you can view this diff [on Phabricator](https://www.internalfb.com/diff/D33589168).

torchdata/datapipes/iter/util/header.py
ejguan: Could we do it in a pythonic way
```suggestion
i = 0
for value in self.source_datapipe:
    i += 1
    if i < self.limit:
        yield value
    else:
        break
self.length = min(i, self.limit)
```

NivekT:
@nivekt has imported this pull request.  If you are a Facebook employee, you can view this diff [on Phabricator](https://www.internalfb.com/diff/D33589168).

PR #169, opened 2022-01-17
fixing amazonreviewpolarity example

parmeet:
This PR fixes couple of things:
1. It introduces caching for extracted files. It will help avoid repetitive extractions with every iteration
2. It fixed the parsed CSV content

### DIFF
--- a/examples/text/amazonreviewpolarity.py
+++ b/examples/text/amazonreviewpolarity.py
@@ -37,25 +37,13 @@ def AmazonReviewPolarity(root, split):
     """Demonstrating caching, extraction and sanity check pipelines."""
 
     url_dp = IterableWrapper([URL])
-    # cache data on-disk with sanity check
-    cache_dp = url_dp.on_disk_cache(
+    cache_compressed_dp = url_dp.on_disk_cache(
         filepath_fn=lambda x: os.path.join(root, _PATH), hash_dict={os.path.join(root, _PATH): MD5}, hash_type="md5"
     )
-    cache_dp = GDriveReader(cache_dp).end_caching(mode="wb", same_filepath_fn=True)
-
-    cache_dp = FileOpener(cache_dp, mode="b")
-
-    # stack TAR extractor on top of loader DP
-    extracted_files = cache_dp.read_from_tar()
-
-    # filter files as necessary
-    filter_extracted_files = extracted_files.filter(lambda x: split in x[0])
-
-    # stack sanity checker on top of extracted files
-    check_filter_extracted_files = filter_extracted_files.check_hash(
-        {os.path.normpath(os.path.join(root, _EXTRACTED_FILES[split])): _EXTRACTED_FILES_MD5[split]},
-        "md5",
-    )
-
-    # stack CSV reader and do some mapping
-    return check_filter_extracted_files.parse_csv().map(fn=lambda t: (int(t[0]), t[1]))
+    cache_compressed_dp = GDriveReader(cache_compressed_dp).end_caching(mode="wb", same_filepath_fn=True)
+    cache_decompressed_dp = cache_compressed_dp.on_disk_cache(
+        filepath_fn=lambda x: os.path.join(root, os.path.dirname(_EXTRACTED_FILES[split]), os.path.basename(x)))
+    cache_decompressed_dp = FileOpener(cache_decompressed_dp, mode="b").read_from_tar()
+    cache_compressed_dp = cache_decompressed_dp.end_caching(mode="wb", same_filepath_fn=True)
+    data_dp = FileOpener(cache_decompressed_dp.filter(lambda x: _EXTRACTED_FILES[split] in x[0]).map(lambda x: x[0]), mode='b')
+    return data_dp.parse_csv().map(fn=lambda t: (int(t[0]), ' '.join(t[1:])))
--- a/examples/text/amazonreviewpolarity.py
+++ b/examples/text/amazonreviewpolarity.py
@@ -18,13 +18,8 @@ NUM_LINES = {
 _PATH = "amazon_review_polarity_csv.tar.gz"
 
 _EXTRACTED_FILES = {
-    "train": os.path.join(_PATH, "amazon_review_polarity_csv", "train.csv"),
-    "test": os.path.join(_PATH, "amazon_review_polarity_csv", "test.csv"),
-}
-
-_EXTRACTED_FILES_MD5 = {
-    "train": "520937107c39a2d1d1f66cd410e9ed9e",
-    "test": "f4c8bded2ecbde5f996b675db6228f16",
+    "train": os.path.join("amazon_review_polarity_csv", "train.csv"),
+    "test": os.path.join("amazon_review_polarity_csv", "test.csv"),
 }
 
 DATASET_NAME = "AmazonReviewPolarity"
@@ -41,9 +36,25 @@ def AmazonReviewPolarity(root, split):
         filepath_fn=lambda x: os.path.join(root, _PATH), hash_dict={os.path.join(root, _PATH): MD5}, hash_type="md5"
     )
     cache_compressed_dp = GDriveReader(cache_compressed_dp).end_caching(mode="wb", same_filepath_fn=True)
+
+    def extracted_filepath_fn(x):
+        file_path = os.path.join(root, _EXTRACTED_FILES[split])
+        dir_path = os.path.dirname(file_path)
+        if not os.path.exists(dir_path):
+            os.makedirs(dir_path)
+        return file_path
+
     cache_decompressed_dp = cache_compressed_dp.on_disk_cache(
-        filepath_fn=lambda x: os.path.join(root, os.path.dirname(_EXTRACTED_FILES[split]), os.path.basename(x)))
-    cache_decompressed_dp = FileOpener(cache_decompressed_dp, mode="b").read_from_tar()
-    cache_compressed_dp = cache_decompressed_dp.end_caching(mode="wb", same_filepath_fn=True)
-    data_dp = FileOpener(cache_decompressed_dp.filter(lambda x: _EXTRACTED_FILES[split] in x[0]).map(lambda x: x[0]), mode='b')
+        filepath_fn=extracted_filepath_fn)
+    cache_decompressed_dp = FileOpener(cache_decompressed_dp, mode="b").\
+                                            read_from_tar().\
+                                            filter(lambda x: _EXTRACTED_FILES[split] in x[0]).\
+                                            map(lambda x: (x[0].replace('_PATH' + '/', ''), x[1]))
+    cache_decompressed_dp = cache_decompressed_dp.end_caching(mode="wb", same_filepath_fn=True)
+    data_dp = FileOpener(cache_decompressed_dp, mode='b')
+
+    # data_dp = FileOpener(cache_compressed_dp, mode='b')
+    # data_dp = data_dp.read_from_tar()
+    # data_dp = data_dp.filter(lambda x: _EXTRACTED_FILES[split] in x[0])
+
     return data_dp.parse_csv().map(fn=lambda t: (int(t[0]), ' '.join(t[1:])))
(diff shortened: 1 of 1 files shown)


examples/text/amazonreviewpolarity.py
```diff
@@ -37,25 +37,13 @@ def AmazonReviewPolarity(root, split):
...
-    )
-
-    # stack CSV reader and do some mapping
-    return check_filter_extracted_files.parse_csv().map(fn=lambda t: (int(t[0]), t[1]))
+    cache_compressed_dp = GDriveReader(cache_compressed_dp).end_caching(mode="wb", same_filepath_fn=True)
+    cache_decompressed_dp = cache_compressed_dp.on_disk_cache(
+        filepath_fn=lambda x: os.path.join(root, os.path.dirname(_EXTRACTED_FILES[split]), os.path.basename(x)))
+    cache_decompressed_dp = FileOpener(cache_decompressed_dp, mode="b").read_from_tar()
+    cache_compressed_dp = cache_decompressed_dp.end_caching(mode="wb", same_filepath_fn=True)
+    data_dp = FileOpener(cache_decompressed_dp.filter(lambda x: _EXTRACTED_FILES[split] in x[0]).map(lambda x: x[0]), mode='b')
```
parmeet, line 48: Hi @ejguan, it seems like the end_caching for decompressed files are not saved onto the disk. I am getting following error: 
`NotADirectoryError: [Errno 20] Not a directory: '/Users/parmeetbhatia/.torchtext/cache/AmazonReviewPolarity/amazon_review_polarity_csv.tar.gz/amazon_review_polarity_csv/train.csv'` 

not sure, what i am doing wrong here. could you please help investigate this?

ejguan: After `read_from_tar`, the data becomes "(decompressed file path, file handle)" where decompressed file path becomes `amazon_review_polarity_csv/amazon_review_polarity_csv/train.csv`. But, as you used `same_filepath_fn=True`, the same function from `on_disk_cache` would be applied to each data, then you would get each file to `amazon_review_polarity_csv.tar.gz/amazon_review_polarity_csv/train.csv`.

You can use `end_caching(mode="wb")` without specifying `same_filepath_fn` to get `amazon_review_polarity_csv/amazon_review_polarity_csv/train.csv` for each decompressed file.

examples/text/amazonreviewpolarity.py
```diff
@@ -37,25 +37,13 @@ def AmazonReviewPolarity(root, split):
...
-    check_filter_extracted_files = filter_extracted_files.check_hash(
-        {os.path.normpath(os.path.join(root, _EXTRACTED_FILES[split])): _EXTRACTED_FILES_MD5[split]},
-        "md5",
-    )
-
-    # stack CSV reader and do some mapping
-    return check_filter_extracted_files.parse_csv().map(fn=lambda t: (int(t[0]), t[1]))
+    cache_compressed_dp = GDriveReader(cache_compressed_dp).end_caching(mode="wb", same_filepath_fn=True)
+    cache_decompressed_dp = cache_compressed_dp.on_disk_cache(
+        filepath_fn=lambda x: os.path.join(root, os.path.dirname(_EXTRACTED_FILES[split]), os.path.basename(x)))
```
ejguan, line 45: I think there is a problem. When we do the cache check at this stage, we want to check if decompressed files existing on local file system. Then, we need to use a generator function as the filepath_fn:
```py
def decompressed_file_fn(x):
    for f in ["train.csv", "test.csv", "readme.txt"]:
        yield os.path.join(root, os.path.dirname(_EXTRACTED_FILES[split]), ..., f)
cache_decompressed_dp = cache_compressed_dp.on_disk_cache(filepath_fn=decompressed_file_fn)
...
```

examples/text/amazonreviewpolarity.py
```diff
@@ -37,25 +32,29 @@ def AmazonReviewPolarity(root, split):
...
+        if not os.path.exists(dir_path):
+            os.makedirs(dir_path)
+        return file_path
+
+    cache_decompressed_dp = cache_compressed_dp.on_disk_cache(
+        filepath_fn=extracted_filepath_fn)
+    cache_decompressed_dp = FileOpener(cache_decompressed_dp, mode="b").\
+                                            read_from_tar().\
+                                            filter(lambda x: _EXTRACTED_FILES[split] in x[0]).\
+                                            map(lambda x: (x[0].replace('_PATH' + '/', ''), x[1]))
```
parmeet, line 52: @ejguan   I think the path returned by read_from_tar is still `amazon_review_polarity_csv.tar.gz/amazon_review_polarity_csv/train.csv`. I have to remove `amazon_review_polarity_csv.tar.gz` to make it work properly.

examples/text/amazonreviewpolarity.py
```diff
@@ -37,25 +32,29 @@ def AmazonReviewPolarity(root, split):
...
-        "md5",
-    )
-
-    # stack CSV reader and do some mapping
-    return check_filter_extracted_files.parse_csv().map(fn=lambda t: (int(t[0]), t[1]))
+    cache_compressed_dp = GDriveReader(cache_compressed_dp).end_caching(mode="wb", same_filepath_fn=True)
+
+    def extracted_filepath_fn(x):
+        file_path = os.path.join(root, _EXTRACTED_FILES[split])
+        dir_path = os.path.dirname(file_path)
```
parmeet, line 42: @ejguan Another thing I realized while working is end_caching won't be able to save file to disk if the corresponding parent directory doesn't exist. I have to explicitly create it. Do you think it would make sense to upstream directory creation in torchdata. For instance `tar` do the same and create appropriate directories when extracting files, I guess it would be non-intuitive for users to explicitly create directories for end_caching?

examples/text/amazonreviewpolarity.py
```diff
@@ -37,25 +32,29 @@ def AmazonReviewPolarity(root, split):
...
-    check_filter_extracted_files = filter_extracted_files.check_hash(
-        {os.path.normpath(os.path.join(root, _EXTRACTED_FILES[split])): _EXTRACTED_FILES_MD5[split]},
-        "md5",
-    )
-
-    # stack CSV reader and do some mapping
-    return check_filter_extracted_files.parse_csv().map(fn=lambda t: (int(t[0]), t[1]))
+    cache_compressed_dp = GDriveReader(cache_compressed_dp).end_caching(mode="wb", same_filepath_fn=True)
+
+    def extracted_filepath_fn(x):
```
parmeet, line 40: @ejguan I am not sure why is it necessary to pass all the files to generator as per your comment here (https://github.com/pytorch/data/pull/169#discussion_r787049096). I guess it is sufficient to only check for file(s) we are concerned with, right? not sure if I miss anything here?

examples/text/amazonreviewpolarity.py
parmeet: Ahh, actually ignore the part of "make it work properly". By using same_filepath_fn=True in end_caching, this part is irrelevant, thanks to @Nayef211 for catching that I am not using `_PATH` properly. That said, the extracted filename still contains `amazon_review_polarity_csv.tar.gz`, this is not really a concern though :)

examples/text/amazonreviewpolarity.py
ejguan: As you can see from the code https://github.com/pytorch/data/blob/c06066ae360fc6054fb826ae041b1cb0c09b2f3b/torchdata/datapipes/iter/util/saver.py#L39-L40, `Saver` would create corresponding parent directory.

examples/text/amazonreviewpolarity.py
ejguan: The second `on_disk_cache` is used to cache the decompressed files from an archive. So, for `on_disk_cache`, it only gets the input archive file path as `xxx.tar.gz` but we want to check if all decompressed files existing on your file system. This is a 1-to-N operation. So, in order to check all of the decompressed files, we have to accept a generator function as the `filepath_fn`.

examples/text/amazonreviewpolarity.py
parmeet: I think we can get rid of the whole map operation here.

examples/text/amazonreviewpolarity.py
```diff
@@ -37,25 +32,29 @@ def AmazonReviewPolarity(root, split):
...
+    cache_decompressed_dp = FileOpener(cache_decompressed_dp, mode="b").\
+                                            read_from_tar().\
+                                            filter(lambda x: _EXTRACTED_FILES[split] in x[0]).\
+                                            map(lambda x: (x[0].replace('_PATH' + '/', ''), x[1]))
+    cache_decompressed_dp = cache_decompressed_dp.end_caching(mode="wb", same_filepath_fn=True)
+    data_dp = FileOpener(cache_decompressed_dp, mode='b')
+
+    # data_dp = FileOpener(cache_compressed_dp, mode='b')
+    # data_dp = data_dp.read_from_tar()
+    # data_dp = data_dp.filter(lambda x: _EXTRACTED_FILES[split] in x[0])
```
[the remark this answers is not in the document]
parmeet: Not really, I was benchmarking and seems like forgot to remove this other piece of  code :)

examples/text/amazonreviewpolarity.py
parmeet: got it, I might be something wrong. Just checked-in the new code, and yes it does not require to explicitly create the directory.

examples/text/amazonreviewpolarity.py
parmeet: I see, got it. Ya, In this case, I am only need to check the existence of single file (even though there are more in the archive). I am guess it would be OK to specify only files we are concerned with and filter out remaining ones in the pipe?

parmeet:
BTW, I think now this PR is ready for final review :) @ejguan

ejguan:
@parmeet Thanks a lot. Could you please add a commit to incorporate lint requirements. You can use pre-commit here. https://github.com/pytorch/data/blob/main/CONTRIBUTING.md#code-style

It would automatically add lint changes when you use `git commit`

parmeet:
> @parmeet Thanks a lot. Could you please add a commit to incorporate lint requirements. You can use pre-commit here. https://github.com/pytorch/data/blob/main/CONTRIBUTING.md#code-style
> 
> It would automatically add lint changes when you use `git commit`

done! @Nayef211, @abhinavarora  this think looks cool, we should probably do this for torchtext?

ejguan (approved):
Thanks, LGTM

Nayef211:
> > @parmeet Thanks a lot. Could you please add a commit to incorporate lint requirements. You can use pre-commit here. https://github.com/pytorch/data/blob/main/CONTRIBUTING.md#code-style
> > It would automatically add lint changes when you use `git commit`
> 
> done! @Nayef211, @abhinavarora this thing looks cool, we should probably do this for torchtext?

This does look really cool. @ejguan do you know if can set this up with the formatter of your choice (i.e. black or autopep8)?

parmeet:
> This does look really cool. @ejguan do you know if can set this up with the formatter of your choice (i.e. black or autopep8)?

I guess this might help answer some https://github.com/pytorch/data/blob/main/.pre-commit-config.yaml

ejguan:
> This does look really cool. @ejguan do you know if can set this up with the formatter of your choice (i.e. black or autopep8)?

Here is the configuration file https://github.com/pytorch/data/blob/main/.pre-commit-config.yaml and `ufmt` can be used for black. I believe you can add your own formatter to the configuration file.

Credit belongs to @pmeier, who helped us to achieve this auto-formatter in this PR https://github.com/pytorch/data/pull/147.

pmeier:
Let me know if another PyTorch repository needs help setting this up.

parmeet:
> Let me know if another PyTorch repository needs help setting this up.

would be happy to get contributions to torchtext :)

Nayef211:
@ejguan just following up to see whether this PR is good to be merged?

ejguan (approved):

PR #177, opened 2022-01-21
Remove duplicated DataPipe reference from bucketbatcher

### LINKED ISSUE #136: [TODO] Verify _datapippe is not going to be serialized twice and is able to reconstruct
https://github.com/pytorch/data/blob/f102d25f9f444de3380c6d49bf7aaf52c213bb1f/torchdata/datapipes/iter/transform/bucketbatcher.py#L60

ejguan:
Fixes #136 

* **#177 Remove duplicated DataPipe reference from bucketbatcher**

Differential Revision: [D33719657](https://our.internmc.facebook.com/intern/diff/D33719657)

### DIFF
--- a/torchdata/datapipes/iter/transform/bucketbatcher.py
+++ b/torchdata/datapipes/iter/transform/bucketbatcher.py
@@ -1,6 +1,6 @@
 # Copyright (c) Facebook, Inc. and its affiliates.
 import random
-from typing import Callable, Iterator, Optional, Sized, TypeVar
+from typing import Callable, Optional, TypeVar
 
 from torch.utils.data import DataChunk
 
@@ -42,8 +42,8 @@ class BucketBatcherIterDataPipe(IterDataPipe[DataChunk[T_co]]):
     in_batch_shuffle: bool
     length: Optional[int]
 
-    def __init__(
-        self,
+    def __new__(
+        cls,
         datapipe: IterDataPipe[T_co],
         batch_size: int,
         drop_last: bool = False,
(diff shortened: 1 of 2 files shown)


ejguan:
@ejguan has imported this pull request.  If you are a Facebook employee, you can view this diff [on Phabricator](https://www.internalfb.com/diff/D33719657).

torchdata/datapipes/iter/transform/bucketbatcher.py
```diff
@@ -42,44 +42,32 @@ class BucketBatcherIterDataPipe(IterDataPipe[DataChunk[T_co]]):
...
         in_batch_shuffle: bool = True,
-    ) -> None:
+    ):
         assert batch_size > 0, "Batch size is required to be larger than 0!"
         assert batch_num > 0, "Number of batches is required to be larger than 0!"
         assert bucket_num > 0, "Number of buckets is required to be larger than 0!"
-        super().__init__()
 
-        # TODO(136): Verify _datapippe is not going to be serialized twice and is able to reconstruct
-        self._datapipe: IterDataPipe[T_co] = datapipe
```
ejguan, line 61: `self._datapipe` is only used to get `__len__`. And, having it would break the expected DataLoader graph. As `IterDataPipe` doesn't guarantee to have length, removing it and `__len__` function from `bucketbacher`.

ejguan: We may be able to rely on graph to generate proper length in the future.

ejguan: Yeah. But, I believe this is not a mandatory feature as we should not let users expecting that `len` is always available using `IterDataPipe`.

PR #179, opened 2022-01-22
adds flatmap datapipe.

### LINKED ISSUE #178: Add a flatmap datapipe
### 🚀 The feature

Add a `FlatMapDataPipe` which simulataneously flattens nested `IterDataPipe`s and applies a function to nested pipes.

### Motivation, pitch

I have tarballs containing tarballs. When dealing with these types of recursive data, it's useful to have a mechanism to both flatten the structure and apply a function to it so I only need to consider the "inner-most" data.

### Alternatives

Domain libraries can define their own impls.

### Additional context

Migrating torchtext to datapipes via torchdata would benefit from this feature when dealing with [IWSLT data](https://github.com/pytorch/text/blob/main/torchtext/datasets/iwslt2016.py#L231-L241).

erip:
Fixes #178

### Changes

- Adds a `FlatMapDataPipe` which applies a function to each nested data pipe and yields from them to flatten lazily.

### DIFF
--- a/test/test_datapipe.py
+++ b/test/test_datapipe.py
@@ -4,6 +4,7 @@ import unittest
 import warnings
 
 from collections import defaultdict
+from functools import partial
 from typing import Dict
 
 import expecttest
@@ -566,6 +567,16 @@ class TestDataPipe(expecttest.TestCase):
         with self.assertRaises(TypeError):
             len(batch_dp)
 
+    def test_flatmap_datapipe(self):
+        l = [1,2,3,4,5,6]
+        first, second = l[:len(l)//2], l[len(l)//2:]
+        nested_wrapper = IterableWrapper([IterableWrapper(first), IterableWrapper(second)])
+        fn = partial(int.__add__, 1)
+        expected_it = list(map(fn, l))
+        flatmapped_dp = list(nested_wrapper.flatmap(fn))
+        #for flatmap_element, pred_element in zip(expected_it, flatmapped_dp):
+        #    self.assertEqual(flatmap_element, pred_element)
+        self.assertEqual(expected_it, flatmapped_dp)
 
 if __name__ == "__main__":
     unittest.main()
--- a/test/test_datapipe.py
+++ b/test/test_datapipe.py
@@ -576,6 +576,8 @@ class TestDataPipe(expecttest.TestCase):
 
         flatmapped_dp = source_dp.flatmap(fn)
         expected_list = list(itertools.chain(*[(e, e*10) for e in source_dp]))
+        flatmapped_dp_list = list(flatmapped_dp)
+        self.assertEqual(expected_list, flatmapped_dp_list)
 
         # Reset Test: reset the DataPipe after reading part of it
         n_elements_before_reset = 5
--- a/test/test_datapipe.py
+++ b/test/test_datapipe.py
@@ -568,13 +568,17 @@ class TestDataPipe(expecttest.TestCase):
             len(batch_dp)
 
     def test_flatmap_datapipe(self):
-        l = [1,2,3,4,5,6]
-        first, second = l[:len(l)//2], l[len(l)//2:]
-        nested_wrapper = IterableWrapper([IterableWrapper(first), IterableWrapper(second)])
-        fn = partial(int.__add__, 1)
-        expected_it = list(map(fn, l))
-        flatmapped_dp = list(nested_wrapper.flatmap(fn))
-        self.assertEqual(expected_it, flatmapped_dp)
+        source_dp = IterableWrapper(list(range(20)))
+
+        def fn(e):
+            return [e, e*10]
+
+        flatmapped_dp = nested_wrapper.flatmap(fn)
+        list_flatmapped_dp = list(flatmapped_dp)
+        expected_list = list(itertools.chain([(e, e*10) for e in source_dp]))
+        self.assertEqual(expected_list, list_flatmapped_dp)
+
+        self.assertEqual(2*len(source_dp), len(flatmapped_dp))
 
 if __name__ == "__main__":
     unittest.main()
--- a/test/test_datapipe.py
+++ b/test/test_datapipe.py
@@ -586,7 +586,7 @@ class TestDataPipe(expecttest.TestCase):
 
 
         # __len__ Test: length should be len(source_dp)*len(fn->out_shape) which we can't know
-        with self.assertRaisesRegex(TypeError, "length relies on the output of its function.")
+        with self.assertRaisesRegex(TypeError, "length relies on the output of its function."):
             len(flatmapped_dp)
 
 if __name__ == "__main__":
--- a/test/test_datapipe.py
+++ b/test/test_datapipe.py
@@ -576,9 +576,16 @@ class TestDataPipe(expecttest.TestCase):
         flatmapped_dp = nested_wrapper.flatmap(fn)
         list_flatmapped_dp = list(flatmapped_dp)
         expected_list = list(itertools.chain([(e, e*10) for e in source_dp]))
-        self.assertEqual(expected_list, list_flatmapped_dp)
 
-        self.assertEqual(2*len(source_dp), len(flatmapped_dp))
+        # Reset Test: reset the DataPipe after reading part of it
+        n_elements_before_reset = 5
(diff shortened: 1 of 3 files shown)


torchdata/datapipes/iter/transform/flatmap.py
```diff
@@ -0,0 +1,28 @@
...
+from torch.utils.data import DataChunk
+from torch.utils.data import functional_datapipe, IterDataPipe
+
+from typing import TypeVar
+
+T_co = TypeVar("T_co", covariant=True)
+
+
+@functional_datapipe("flatmap")
+class FlatMapDataPipe(IterDataPipe[IterDataPipe[DataChunk[T_co]]]):
```
ejguan, line 11: Could you change the name to `FlatMapperIterDataPipe`? We want to name each DataPipe class as `op-er` and functional API as the `op`.

torchdata/datapipes/iter/transform/flatmap.py
```diff
@@ -0,0 +1,28 @@
...
+        datapipe: Iterable datapipe containing iterable datapipes to which the function is applied
+        fn: the function to be applied to each of the ``inner" datapipes
+    """
+    def __init__(self, datapipe, fn):
+        self.datapipe = datapipe
+        self.fn = fn
+
+    def __iter__(self):
+        for e in self.datapipe:
+            yield from e.map(self.fn)
```
ejguan, line 28: Let's not assume each data inside the `self.datapipe` is an `IterDataPipe`. We can do:
```suggestion
        for e in self.datapipe:
            yield from self.fn(e)
```

torchdata/datapipes/iter/transform/flatmap.py
```diff
@@ -0,0 +1,28 @@
...
+    Iterable DataPipe which applies a function to an IterableDataPipe containing
+    IterableDataPipes and flattens to a single unnested IterableDataPipe.
+
+    Args:
+        datapipe: Iterable datapipe containing iterable datapipes to which the function is applied
+        fn: the function to be applied to each of the ``inner" datapipes
+    """
+    def __init__(self, datapipe, fn):
+        self.datapipe = datapipe
+        self.fn = fn
```
ejguan, line 24: Could you also add a check lambda function here? https://github.com/pytorch/pytorch/blob/b372be4211d40be92ffbb025cac0038ee3445141/torch/utils/data/datapipes/iter/callable.py#L47

`lambda` function can not be pickled.

test/test_datapipe.py
```diff
@@ -569,6 +570,14 @@ def _return_self(x):
         # __len__ Test: returns the number of batches
         self.assertEqual(3, len(batch_dp))
 
+    def test_flatmap_datapipe(self):
```
ejguan, line 573: Could you please add a more comprehensive tests?
We at least need to add validation on `reset` using `reset_after_n_next_calls` shown from the other test cases, and `__len__` test.

ejguan (commented):
Thanks for adding this PR, we do want to implement `flatmap`.

Could you please also take a reference from `MapperIterDataPipe` here. https://github.com/pytorch/pytorch/blob/b372be4211d40be92ffbb025cac0038ee3445141/torch/utils/data/datapipes/iter/callable.py#L105-L131

Since we want to make sure each DataPipe instance is serializable for the future snapshotting, could you please add the similar `__getstate__` and `__setstate__` to make sure `self.fn` is serialized if `dill` is available.

test/test_datapipe.py
erip: Yes, I'll add some tests momentarily.

erip:
I've made some changes to the test. One that I'm a bit sketchy on is the `__len__` test since the length is a function of the size of the output of the `fn` (explained in-line in the test).

As an aside, I can't seem to test this locally? I've installed pytorch nightly and see this error when I run `pytest test/test_datapipe.py`: 

> ImportError: cannot import name 'DILL_AVAILABLE' from 'torch.utils.data.datapipes.utils.common'

ejguan:
> I've made some changes to the test. One that I'm a bit sketchy on is the `__len__` test since the length is a function of the size of the output of the `fn` (explained in-line in the test).

Then, we should expect `__len__` not implemented in this case. It's allowed for `IterDataPipe`. And, a test to validate an Error would be raised would be sufficient.

> As an aside, I can't seem to test this locally? I've installed pytorch nightly and see this error when I run `pytest test/test_datapipe.py`:
> 
> > ImportError: cannot import name 'DILL_AVAILABLE' from 'torch.utils.data.datapipes.utils.common'
> 
> It seems like CI only runs for certain people, so I don't know if this is OK 😞

This is due to PyTorch nightly release for both mac and windows has been stalled for 10 days... 😢

erip:
That's fine! So I should remove the __len__ test and we just need to wait?

ejguan:
> That's fine! So I should remove the `__len__` test and we just need to wait?

You can do `self.assertRaises` for `__len__` test.

erip:
OK, great!

torchdata/datapipes/iter/__init__.py
```diff
@@ -41,6 +41,7 @@
     OnlineReaderIterDataPipe as OnlineReader,
 )
 from torchdata.datapipes.iter.transform.bucketbatcher import BucketBatcherIterDataPipe as BucketBatcher
+from torchdata.datapipes.iter.transform.flatmap import FlatMapDataPipe as FlatMap
```
ejguan, line 44: nit: name needs to be changed here.

test/test_datapipe.py
```diff
@@ -569,6 +570,27 @@ def _return_self(x):
...
+        source_dp = IterableWrapper(list(range(20)))
+
+        def fn(e):
+            return [e, e*10]
+
+        flatmapped_dp = nested_wrapper.flatmap(fn)
```
ejguan, line 579: nit: nested_wrapper -> source_dp

ejguan (approved):
The PyTorch nightly is updated. Could you please address the lint error? Then, I can import the PR.

erip:
OK, @ejguan -- I can run tests locally now and it seems good now.

test/test_datapipe.py
```diff
@@ -569,6 +570,27 @@ def _return_self(x):
...
+        list_flatmapped_dp = list(flatmapped_dp)
+        expected_list = list(itertools.chain([(e, e*10) for e in source_dp]))
```
ejguan, line 581: Should we do another `self.assertEqual(list_flatmapped_dp, expected_list)` here?

torchdata/datapipes/iter/transform/flatmap.py
```diff
@@ -0,0 +1,62 @@
+# Copyright (c) Facebook, Inc. and its affiliates.
+from torch.utils.data import DataChunk
+from torch.utils.data import functional_datapipe, IterDataPipe
+from torch.utils.data.datapipes.utils.common import DILL_AVAILABLE, check_lambda_fn
+
+from typing import TypeVar
+
+if DILL_AVAILABLE:
+    import dill
```
ejguan, line 9: The Error from `mypy` can be ignored here. I have another PR that should solve it.

ejguan:
Could you please do a rebase onto main branch?

erip:
Rebased!

test/test_datapipe.py
```diff
@@ -1,9 +1,11 @@
 # Copyright (c) Facebook, Inc. and its affiliates.
 import io
+import itertools
 import unittest
 import warnings
 
 from collections import defaultdict
+from functools import partial
```
ejguan, line 8: Last nit: Could you please also remove `partial` here, as it's not used?

erip: Done! Not sure how linting missed that, sorry.

erip:
@ejguan I can't see what the Meta-internal issues are, but I'm happy to try to resolve any warnings if it makes sense for me to do that. Otherwise, I'm excited to see this land. 😎

ejguan:
> @ejguan I can't see what the Meta-internal issues are, but I'm happy to try to resolve any warnings if it makes sense for me to do that. Otherwise, I'm excited to see this land. 😎

No worries. I did check these Error and they are not related to this PR. I have triggered shipment of this PR and it will be merged shortly.
Thanks a lot!! @erip

PR #182, opened 2022-01-27
Fix reference test from torchtext

ejguan:
SST is moved from prototype to datasets in https://github.com/pytorch/text/pull/1538. The reference test in TorchData needs to be moved correspondingly.

cc: @Nayef211 Let me know when you have unit test for each Dataset. Then, I can directly use the test inside TorchText.


NivekT (approved):
LGTM! The CI failures look unrelated to this PR since they are about TorchVision.

Nayef211:
> cc: @Nayef211 Let me know when you have unit test for each Dataset. Then, I can directly use the test inside TorchText.

Will do! We're working on adding these tests over the next week or so!

PR #194, opened 2022-02-02
Create .nojekyll

brianjo:
Adds .nojekyll to gh-pages as the site is not jekyll.


NivekT (approved):
Thanks for adding this

Issue #103, opened 2021-11-08
Switch `FileLoader` to `FileOpener`


ejguan:
Should we change the name of `FileLoader` to `FileOpener`?

We split the file-loading functionality into three steps:
- List file names in a directory: `FileLister`
- Open file handles: `FileLoader` (I personally feel like the name is incorrect.)
- Read files: `dp.map(fn=lambda x: x.read(), input_col=1)`

cc: @VitalyFedyunin @NivekT

NivekT:
From a purely naming perspective, I think `FileOpener` is more accurate but `FileLoader` isn't wildly misleading either.

I had a look at `numpy.load`, `json.load`, `pickle.load`, these seem to either read or parse through files that have already been opened and return some structured data. In that sense, `FileLoader` definitely behaves differently that other modules. Renaming it is likely better.

Perhaps we can rename it, but still leave `FileLoader` functional with deprecation warning? I would imagine most people who started using `IterDataPipe` in PyTorch Core would be using `FileLoader` and it would be a BC-breaking change.

ejguan:
> I had a look at `numpy.load`, `json.load`, `pickle.load`, these seem to either read or parse through files that have already been opened and return some structured data. In that sense, `FileLoader` definitely behaves differently that other modules. Renaming it is likely better.

Thanks for digging into it. You are right. And, the most concerning part is the functional API since we are currently using `load_file_...` to for each `FileLoader`. I don't want users to complain the inconsistent behavior across libs.

> Perhaps we can rename it, but still leave `FileLoader` functional with deprecation warning? I would imagine most people who started using `IterDataPipe` in PyTorch Core would be using `FileLoader` and it would be a BC-breaking change.

Thank you for pointing out. If we plan to move, we should definitely add deprecation warning at least before official release.

ejguan:
Also want to gather some insights from domains since this is going to be BC breaking.
cc: @pmeier @Nayef211

Nayef211:
> Also want to gather some insights from domains since this is going to be BC breaking. cc: @pmeier @Nayef211

I agree that `FileLoader` sounds misleading and that `FileOpener` would be a better name for what the datapipe is actually doing. I also think adding a deprecation warning is a good idea, so that users have time to migrate to the new datapipe.

pmeier:
+1 for `FileOpener`. 

[...]

PR #197, opened 2022-02-04
Prevent RarArchiveLoader reading nested RAR explicitly

### LINKED ISSUE #189: `RarArchiveLoader` fails for nested rar archives
### 🐛 Describe the bug

For this example, I'm using [this archive]().

---

For a single archive, `RarArchiveLoader` works fine:

```py
import pathlib

from torchdata.datapipes.iter import RarArchiveLoader, IterableWrapper, FileOpener

dp = IterableWrapper(["/home/philip/datasets/hmdb51/hmdb51_org.rar"])
dp = FileOpener(dp, mode="rb")
dp = RarArchiveLoader(dp)
print([pathlib.Path(data[0]).name for data in dp])
```

```
['shoot_gun.rar', 'sit.rar', 'situp.rar', 'smile.rar', 'smoke.rar', 'somersault.rar', 'stand.rar', 'swing_baseball.rar', 'sword.rar', 'sword_exercise.rar', 'talk.rar', 'throw.rar', 'turn.rar', 'walk.rar', 'wave.rar', 'brush_hair.rar', 'cartwheel.rar', 'catch.rar', 'chew.rar', 'clap.rar', 'climb.rar', 'climb_stairs.rar', 'dive.rar', 'draw_sword.rar', 'dribble.rar', 'drink.rar', 'eat.rar', 'fall_floor.rar', 'fencing.rar', 'flic_flac.rar', 'golf.rar', 'handstand.rar', 'hit.rar', 'hug.rar', 'jump.rar', 'kick.rar', 'kick_ball.rar', 'kiss.rar', 'laugh.rar', 'pick.rar', 'pour.rar', 'pullup.rar', 'punch.rar', 'push.rar', 'pushup.rar', 'ride_bike.rar', 'ride_horse.rar', 'run.rar', 'shake_hands.rar', 'shoot_ball.rar', 'shoot_bow.rar']
```

As you can see from the output, the archive is actually a rar of rars. Thus, we need to use `RarArchiveLoader` again. Trying to iterate now leads to 

```py
dp = RarArchiveLoader(dp)
next(iter(dp))
```

```
ValueError: whence value 7 unsupported
```

pointing to 

https://github.com/pytorch/data/blob/fb29c810e6e8881e5a3558743d9573cb82134cd6/torchdata/datapipes/iter/util/rar_archive_loader.py#L22

@ejguan could you have a look?

### Versions

Current `main`.

ejguan:
Fixes #189

### Changes

- ~Rename `RarArchiveLoader` to `RarArchvieReader` and corresponding functional API~
- Raise Error when reading nested RAR
- Add tests for nested RAR and nested TAR.

cc: @pmeier There is a BC-breaking naming change.

### DIFF
--- a/test/test_local_io.py
+++ b/test/test_local_io.py
@@ -592,13 +592,27 @@ class TestDataPipeLocalIO(expecttest.TestCase):
 
     def _write_test_rar_files(self):
         # `rarfile` can only read but not write .rar archives so we use to system utilities
-        subprocess.run(("rar", "a", os.path.join(self.temp_dir.name, "test_rar.rar"), *self.temp_files), check=True)
+        rar_archive_name = os.path.join(self.temp_dir.name, "test_rar")
+        subprocess.run(("rar", "a", rar_archive_name + ".rar", *self.temp_files), check=True)
+
+        # Nested RAR
+        subprocess.run(("rar", "a", rar_archive_name + "1.rar", self.temp_files[0]), check=True)
+        subprocess.run(("rar", "a", rar_archive_name + "2.rar", *self.temp_files[1:]), check=True)
+        subprocess.run(
+            ("rar", "a", rar_archive_name + "_nested.rar", rar_archive_name + "1.rar", rar_archive_name + "2.rar"),
+            check=True,
+        )
+
+        # Nested RAR in TAR
+        with tarfile.open(rar_archive_name + "_nested.tar", "w:tar") as tar:
+            tar.add(rar_archive_name + "1.rar")
+            tar.add(rar_archive_name + "2.rar")
 
     @skipIfNoRarTools
     def test_rar_archive_loader(self):
         self._write_test_rar_files()
 
-        datapipe1 = FileLister(self.temp_dir.name, "*.rar")
+        datapipe1 = IterableWrapper([os.path.join(self.temp_dir.name, "test_rar.rar")])
         datapipe2 = FileOpener(datapipe1, mode="b")
         rar_loader_dp = RarArchiveLoader(datapipe2)
 
@@ -611,7 +625,7 @@ class TestDataPipeLocalIO(expecttest.TestCase):
 
         # Reset Test: reset the DataPipe after reading part of it
         rar_loader_dp = datapipe2.load_from_rar()
-        n_elements_before_reset = 1
+        n_elements_before_reset = 2
         res_before_reset, res_after_reset = reset_after_n_next_calls(rar_loader_dp, n_elements_before_reset)
         # Check the results accumulated before reset
         self._unordered_compressed_files_comparison_helper(self.temp_files[:n_elements_before_reset], res_before_reset)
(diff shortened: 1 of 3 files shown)


test/test_local_io.py
```diff
@@ -622,6 +636,28 @@ def test_rar_archive_loader(self):
...
+        datapipe2 = FileOpener(datapipe1, mode="b")
+        tar_loader_dp = TarArchiveReader(datapipe2)
+        rar_loader_dp = RarArchiveReader(tar_loader_dp)
+
+        # Functional Test: read extracted files before reaching the end of the rarfile
+        self._unordered_compressed_files_comparison_helper(self.temp_files, rar_loader_dp, check_length=False)
+
+        # Functional Test: read extracted files after reaching the end of the rarfile
+        data_refs = list(rar_loader_dp)
+        self._unordered_compressed_files_comparison_helper(self.temp_files, data_refs)
```
ejguan, line 659: These two examples shows the difference between nested RAR in RAR and in TAR.

pmeier:
Re naming change: Recall, that we explicitly went for "Loader" in #41. Unless something has changed in the mean time, you should probably rename the other "Reader"'s to "Loader"'s.

ejguan:
> Re naming change: Recall, that we explicitly went for "Loader" in #41. Unless something has changed in the mean time, you should probably rename the other "Reader"'s to "Loader"'s.

Yes. But, it seems we reach another agreement #103 later.
cc: @NivekT

ejguan:
> Since you haven't replied to that, I was under the impression that was the way to go. Anyway, `Reader` seems misleading for these, since no reading is performed.

I made a mistake of naming, and thanks for pointing out. `Reader` is not appropriate here.

NivekT:
> In [#41 (comment)](https://github.com/pytorch/data/pull/41#issuecomment-968591908) I wrote:
> 
> > True, but that means neither `Lister` nor `Opener` is a good [f]it here, since both actions are performed. Maybe `RarArchiveLoader`?
> 
> Since you haven't replied to that, I was under the impression that was the way to go. Anyway, `Reader` seems misleading for these, since no reading is performed.

Personally, I think `Reader` might be the best given that we have other DataPipes that perform decompression and are named `Reader` (i.e. `tar`, `xz`, `zip`). Reading a compressed file and returning a stream seems okay to me?

@pmeier do you also think `tar`, `xz`, `zip` readers should be renamed?

ejguan:
I removed renaming from this PR. We can decide which is the best name for all archive loader/opener/reader in a separate PR.

ejguan:
Failing tests on windows and mac are caused by stalled nightly release again.

pmeier:
@NivekT 

Two issues here:

1. `.xz`, `.gz`, and so on are completely different concepts from `.tar`, `.zip`, and so on. The former is compression while the latter is archiving. You can have both separately. The current name of `Extractor` is misleading since it doesn't extract anything, but rather decompresses it. For example, it would turn `foo.tar.gz` into `foo.tar` without ever touching the internals of the archive. 
2. The activity "read" is usually associated with getting data from a file. Thus, `LineReader` is a good description, because you actually read the data. In contrast, `TarArchiveReader` is not since we never touch the data of the files inside the archive. 

    There are more inconsistencies like that throughout `torchdata` and this is ok for the prototype phase. This needs to be resolved before we commit to BC. My suggestion was the following taxonomie:

    - list: Take a single path and return all included paths. For example, `FileLister`
    - open: Take a single path and return the path plus the open file handle. For example, `FileOpener`
    - load: Combine list and open. For example, `RarFileLoader`, `TarFileReader`, `ZipFileReader` 
    - read: Anything that actually returns data in whatever form of a file handle. For example `LineReader`, `CSVParser`, `CSVDictParser`

    Of course you don't need to use this, but whatever you decide on, it should be consistent throughout the code base.

NivekT:
@pmeier I agree `load` is marginally better for these archive files (not that `read` will make or break things, I had a look at the `tarfile` documentation and it uses the word 'read' pretty liberally).

I also concur that `extract` is confusing in that people may expect things to be written to file systme. I think we should rename that to `decompress` or `in_memory_extract`?

NivekT (approved):
LGTM, what is the reasoning behind not supporting nested RAR? Might be good to explicit include the reasoning somwhere (as a comment or error message)

ejguan:
> LGTM, what is the reasoning behind not supporting nested RAR? Might be good to explicit include the reasoning somwhere (as a comment or error message)

Since we are relying on third-party `rarfile` library and RAR is a close-source format, I can't really tell the reason why it can not be supported

pmeier:
> I had a look at the `tarfile` documentation and it uses the word 'read' pretty liberally

They only have to have consistent use of "read" throughout this module so they may get away with a more relaxed definition. Unfortunately, this doesn't apply to us.

> I also concur that `extract` is confusing in that people may expect things to be written to file systme. I think we should rename that to `decompress` or `in_memory_extract`?

+1 for `decompress` given that is the most precise. `in_memory_extract` has again the "extract" keyword that might be misleading. For example `IterableWrapper(["foo.tar.gz"]).in_memory_extract()` would give you the file handle corresponding to `foo.tar` rather than the contents of `foo.tar`.

ejguan:
Then, let's make a decision here:
`ArchiveLoader`: Taking charge of listing and opening inner files
`Decompressor`: Taking charge of decompressing the compressed files/archives like `file.gz` or `a.tar.gz`

PR #205, opened 2022-02-08
Adding documentation for utility functions

NivekT:
Stack from [ghstack](https://github.com/ezyang/ghstack):
* #212
* #211
* #209
* #208
* #207
* #206
* __->__ #205

Setup documentation for utility function and added some links to PyTorch Core documentation.

Let me know what else should be added here.

Differential Revision: [D34093684](https://our.internmc.facebook.com/intern/diff/D34093684)

PR #206, opened 2022-02-08
Override upper-case conversion of titles

NivekT:
* #212
* #211
* #209
* #208
* #207
* __->__ #206
* #205

Adding CSS file to override the default setting where all titles are in UPPER CASE.

Differential Revision: [D34093683](https://our.internmc.facebook.com/intern/diff/D34093683)


NivekT:
@nivekt has imported this pull request.  If you are a Facebook employee, you can view this diff [on Phabricator](https://www.internalfb.com/diff/D34093683).

NivekT:
> Do you mean making titles not in upper-case? If so, please do not do it. We need to have an aligned format across different repos under PyTorch https://pytorch.org/docs/stable/index.html.

Yep, it makes the title not in upper case. It looks very odd and less readable for the DataPipe break out pages.

For example, this one:
https://ejguan.github.io/generated/torchdata.datapipes.iter.FSSpecFileLister.html#torchdata.datapipes.iter.FSSpecFileLister

Please render this PR to have a look and let me know what you think of the visual difference.

I am open to other suggestion. I can also check with our docs team to see if that is acceptable.

NivekT:
@ejguan I just checked with the docs team lead, he said we can do as we wish. I share your concern and we should make sure that this is acceptable.

Here is the before:
Screen Shot 2022-02-09 at 11 48 49 AM


Here is the after:
Screen Shot 2022-02-09 at 11 50 42 AM

cc: @dongreenberg to double check

ejguan:
Yeah, it looks weird. Just want to have a unified doc theme across PyTorch packages. 
- PyTorch Core: https://pytorch.org/docs/stable/generated/torch.is_tensor.html#torch.is_tensor


Another way maybe doing the similar approach like TorchVision. https://pytorch.org/vision/stable/datasets.html
We can have all DataPipe in the same page (iter, map may split into two pages). Then, the title is going to be `TORCHDATA.DATAPIPES.ITER` or `TORCHDATA.DATAPIPES.MAP`, and all classes are rendered as the expected format.

VitalyFedyunin (approved):

VitalyFedyunin:
Alternatively we can all agree to use CAPS in titles, when we put them manually. Is it going to work?

NivekT:
> Yeah, it looks weird. Just want to have a unified doc theme across PyTorch packages.
> 
> * PyTorch Core: https://pytorch.org/docs/stable/generated/torch.is_tensor.html#torch.is_tensor

I think we should ask for approval to have this small difference, since it clearly looks better.

> Another way maybe doing the similar approach like TorchVision. https://pytorch.org/vision/stable/datasets.html We can have all DataPipe in the same page (iter, map may split into two pages). Then, the title is going to be `TORCHDATA.DATAPIPES.ITER` or `TORCHDATA.DATAPIPES.MAP`, and all classes are rendered as the expected format.

That is because all titles are automatically set to be UPPER CASE, and non-titles are not affected as far as I can tell.

TorchVision is actually moving away from having everything on the same page. See the latest doc:
https://pytorch.org/vision/main/datasets.html

NivekT:
Note that I confirmed that this is acceptable with @brianjo. Let me know if anyone still has any concern.

NivekT:
@nivekt has imported this pull request.  If you are a Facebook employee, you can view this diff [on Phabricator](https://www.internalfb.com/diff/D34093683).

ejguan (approved):

PR #207, opened 2022-02-08
Adding more signature replacements

NivekT:
Stack from [ghstack](https://github.com/ezyang/ghstack):
* #212
* #211
* #209
* #208
* __->__ #207
* #206
* #205

Replacing long signature types with more readable ones. Using regular expression should be better here but I left it as a TODO.

Differential Revision: [D34093685](https://our.internmc.facebook.com/intern/diff/D34093685)

PR #208, opened 2022-02-08
Fixing flatmap signature and documentation

NivekT:
* #212
* #211
* #209
* __->__ #208
* #207
* #206
* #205

Fixing `flatmap` signature and documentation because it is more flexible than the previous documentation indicates.

Differential Revision: [D34093682](https://our.internmc.facebook.com/intern/diff/D34093682)


NivekT:
@nivekt has imported this pull request.  If you are a Facebook employee, you can view this diff [on Phabricator](https://www.internalfb.com/diff/D34093682).

torchdata/datapipes/iter/transform/flatmap.py
```diff
@@ -13,18 +13,17 @@
...
-    flattens to a single unnested IterableDataPipe.
+class FlatMapperIterDataPipe(IterDataPipe[T_co]):
+    r"""
+    Applies a function over each item from the source DataPipe, then
+    flattens the outputs to a single, unnested IterDataPipe (functional name: ``flatmap``).
 
     Args:
-        datapipe: Iterable datapipe containing iterable datapipes to which the function is applied
-        fn: the function to be applied to each of the "inner" datapipes
+        datapipe: Source Iterable DataPipe
```
ejguan, line 22: nit: Source IterDataPipe

torchdata/datapipes/iter/transform/flatmap.py
```diff
@@ -13,18 +13,17 @@
...
+        fn: the function to be applied to each element in the DataPipe
```
ejguan, line 23: Can we mention we expect `fn` returns a Sequence (List or Tuple)?

ejguan (approved):

torchdata/datapipes/iter/transform/flatmap.py
NivekT: This is the part that I feel unsure about.

In the programming language Scala (and probably Java?), `flatmap` actually allows you to use any `fn`, not just a function that return `Sequence` (i.e. no structural change is necessary). Effectively, it is `.map(fn).flatten()`. 

Here is what [Apache Beam does](https://beam.apache.org/documentation/transforms/python/elementwise/flatmap/#example-3-flatmap-with-a-lambda-function). It also doesn't require structural change from `fn`.

I think that may be the case for `pyspark.RDD.flatMap` as well.

On the other hand, [TensorFlow](https://www.tensorflow.org/api_docs/python/tf/data/Dataset#flat_map) seems to want a structural changing `fn`. Not sure if it is enforced.

Here is what our current code does:
```python
dp = IterableWrapper(["abc"])
dp = dp.flatmap(lambda x: x)
print(list(dp))
# ['a', 'b', 'c']
```
There is no enforcement that `fn` have to return a Sequence. This behavior might be unexpected but it seems to match what some other programming languages do. 

To be clear, I think most users will pass in a structural changing `fn` (e.g. `string.split()`) and we should show examples of those when we add examples in our docstring, but technically is not a requirement.

I think the options are:

1. Do not enforce a structural changing `fn` and leave the doc as it is because that is how our implementation current behaves. We will add examples with structural changing `fn` showing users that is how `flatmap` is typically used.

2. Enforce having a structural changing `fn` and change the doc as well.

Thoughts?

ejguan: `fn` doesn't need to be structural transform to return `Sequence`. I believe `str` is still Sequence in Python. 
Since it is error out in the following case because the output can not be `yield from`. We should mention about it in docstring.
```
dp = IterableWrapper([123])
dp = dp.flatmap(lambda x: x)
print(list(dp))  # Error out
```

NivekT: I added that to the docstring.

NivekT:
@nivekt has imported this pull request.  If you are a Facebook employee, you can view this diff [on Phabricator](https://www.internalfb.com/diff/D34093682).

PR #223, opened 2022-02-14
Update docs.yml

### LINKED ISSUE #222: Workflow to generate documentation should not deploy for a PR
### 🐛 Describe the bug

My PR #221 should not update the `gh-pages` branch in the workflow to generate documentation. https://github.com/pytorch/data/tree/gh-pages

We can test if doc generation is correct or not on PRs. And, we should only deploy changes to gh-pages when it's a push on `main` branch.

### Versions

main

brianjo:
Changes docs build to run on merge to main.

Fixes #222

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -1,6 +1,8 @@
 name: Generate documentation
-on: [push]
-
+on:
+  push:
+    branches:
+      - main
 jobs:
   build_docs_job:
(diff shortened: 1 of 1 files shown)


ejguan (commented):
Thanks for fixing it.
QQ: Is there anything to incorporate in `gh-pages` branch? Do we need to revert a few commits in `gh-pages` that comes from open PR but changes doc files?

PR #228, opened 2022-02-17
Disable RC release on main branch

ejguan:
Due to the limitation of current GHA, scheduled workflow can't run on release branch. I have to switch the workflow to manually trigger for release branch. Please check: https://github.community/t/scheduled-builds-of-non-default-branch/16306

The right workflow will be created in release branch rather than main branch.


NivekT (approved):
LGTM

PR #257, opened 2022-02-25
Editing list of available versions

NivekT:
Adding 0.3.0 to this page - https://pytorch.org/data/versions.html

The default page that users reach when they visit pytorch.org/data will still be the documentation for the`main` branch. I recommend keeping it that way unless you have other thoughts.


ejguan (commented):
And, can you please update https://github.com/pytorch/data/blob/gh-pages/index.html to point to 0.3.0?

NivekT:
> And, can you please update https://github.com/pytorch/data/blob/gh-pages/index.html to point to 0.3.0?

That's the part I am unsure about, it depends how much our API/doc will change in the very near future, but I can do that.

NivekT:
It is done. The landing page will now redirect to `beta` which redirects to `0.3.0`.

beta
```diff
@@ -0,0 +1 @@
+0.3.0
```
ejguan, line 1: Symbolic link

index.html
```diff
@@ -4,9 +4,8 @@
         <title>TorchData</title>
     </head>
     <body>
-        <p>The documentation is <a href="/data/main/index.html">here</a></p>
+        <p>The documentation is <a href="/data/beta/index.html">here</a></p>
```
ejguan, line 7: If you want to do so, you need to change `beta` as a symbolic link to `0.3.0`

ejguan (changes requested):
I don't think this is going to work. Can you please do a test?

beta
NivekT: Updated

NivekT:
> I don't think this is going to work. Can you please do a test?

Updated as symbolic link

ejguan (approved):

PR #262, opened 2022-02-25
README: Add version compatibility table

ejguan:


NivekT (approved):

setup.py
```diff
@@ -59,8 +59,10 @@ def get_parser():
     return parser
 
 
-pytorch_package_dep = "torch"
+# Use new version of torch on main branch
+pytorch_package_dep = "torch>1.11.0"
```
ejguan, line 63: Just found the version should be updated to `torch>1.12`

PR #266, opened 2022-02-28
Fix bugs for setup and CI (Dep & PyTorch Version)

ejguan:
### Changes

- Fix dependency requirements for setup (urlliib3)
- Add condition to prevent workflow executed on PRs to upload wheel
- Add env to specify PyTorch version

### DIFF
--- a/.github/workflows/nightly_release.yml
+++ b/.github/workflows/nightly_release.yml
@@ -19,6 +19,7 @@ jobs:
       branch: ""
       pip_path: https://download.pytorch.org/whl/nightly/cpu/torch_nightly.html
       pre_dev_release: true
+      pytorch_version: ""
     secrets:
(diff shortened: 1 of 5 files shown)


.github/workflows/nightly_release.yml
```diff
@@ -19,6 +19,7 @@ jobs:
       branch: ""
       pip_path: https://download.pytorch.org/whl/nightly/cpu/torch_nightly.html
       pre_dev_release: true
+      pytorch_version: ""
```
NivekT, line 22: What's the purpose of adding these `""`?

NivekT (approved):

.github/workflows/nightly_release.yml
ejguan: empty string is mainly used by nightly because we don't know the nightly version of pytorch before we start to run the workflow.
But, for release/rc-release, we would add a hard code value to it like `1.11.0` to enforce our dependency is fixed.

PR #283, opened 2022-03-07
[DataPipe] Implement BucketBatcher with max token

ejguan:
Add `max_token_bucketizer`

There are two options to do streaming bucketizer:
- Using `buffer` as a priority queue to keep yielding the shortest token dynamically. Whenever a token is fetched from buffer, a new token is pushed into the buffer.
- Sort `buffer` and yield batches from `buffer` without adding new tokens from DataPipe until the batches are yielded from `buffer`.

The time complexity of these two methods should be the same as `O(N log(buffer_size)`. I prefer the first approach as we would reduce the potential skewness of data within local buffer.

Added `in_batch_shuffle` as the syntax sugar and for future deterministic data preprocessing.

### DIFF
--- a/torchdata/datapipes/iter/transform/bucketbatcher.py
+++ b/torchdata/datapipes/iter/transform/bucketbatcher.py
@@ -1,6 +1,9 @@
 # Copyright (c) Facebook, Inc. and its affiliates.
+import heapq
 import random
-from typing import Callable, Optional, TypeVar
+
+from functools import partial
+from typing import Callable, Iterator, List, Optional, TypeVar
 
 from torch.utils.data import DataChunk
 
@@ -10,10 +13,35 @@ from torchdata.datapipes.iter import IterDataPipe
 T_co = TypeVar("T_co", covariant=True)
 
 
-def _in_batch_shuffle_fn(data: DataChunk):
-    d = list(data)
-    random.shuffle(d)
-    return DataChunk(d)
+@functional_datapipe("in_batch_shuffle")
+class InBatchShufflerIterDataPipe(IterDataPipe[DataChunk[T_co]]):
+    r"""
+    Shuffles each mini-batch from the prior DataPipe (functional name: ``in_batch_shuffle``).
+
+    Args:
+        datapipe: Iterable DataPipe with batched data
+
+    Example:
+        >>> from torchdata.datapipes.iter import IterableWrapper
+        >>> source_dp = IterableWrapper(range(10))
+        >>> batch_dp = source_dp.batch(batch_size=3, drop_last=True)
+        >>> list(batch_dp)
+        [[0, 1, 2], [3, 4, 5], [6, 7, 8]]
+        >>> in_batch_shuffle_dp = batch_dp.in_batch_shuffle()
+        >>> list(in_batch_shuffle_dp)
+        [[2, 0, 1], [3, 5, 4], [7, 8, 6]]
+    """
+
+    def __init__(self, datapipe: IterDataPipe[DataChunk[T_co]]):
+        self.datapipe = datapipe
+
+    def __iter__(self) -> Iterator[DataChunk[T_co]]:
+        for batch in self.datapipe:
+            random.shuffle(batch)
+            yield batch
+
+    def __len__(self) -> int:
+        return len(self.datapipe)
 
 
 @functional_datapipe("bucketbatch")
@@ -58,7 +86,6 @@ class BucketBatcherIterDataPipe(IterDataPipe[DataChunk[T_co]]):
     bucket_num: int
     sort_key: Optional[Callable]
     in_batch_shuffle: bool
-    length: Optional[int]
 
     def __new__(
         cls,
@@ -80,7 +107,7 @@ class BucketBatcherIterDataPipe(IterDataPipe[DataChunk[T_co]]):
         # Shuffle by pool_size
         if bucket_num > 1 or sort_key is None:
             if in_batch_shuffle:
-                datapipe = datapipe.batch(batch_size=pool_size, drop_last=False).map(fn=_in_batch_shuffle_fn).unbatch()
+                datapipe = datapipe.batch(batch_size=pool_size, drop_last=False).in_batch_shuffle().unbatch()
             else:
                 datapipe = datapipe.shuffle(buffer_size=pool_size)
         # Sort by bucket_size if sort_key is given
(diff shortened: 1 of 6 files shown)


torchdata/datapipes/iter/transform/bucketbatcher.py
```diff
@@ -92,7 +104,116 @@ def __new__(
...
+            heapq.heappush(buffer, d)
+            if len(buffer) == self.buffer_size:
+                length, token = heapq.heappop(buffer)
+                if batch_size + length > self.max_token_count:
+                    if not batch:
+                        warnings.warn(
+                            "Empty batch is yielded due to length of a single token is larger than"
+                            "``max_token_count``. Please consider add ``max_len`` to filter such token or"
+                            "increase ``max_token_count``."
+                        )
```
nateanl, line 198: This check can be done before the iteration. If ``max_token_count`` is not ``None`` and `max_len`` is not given, we should check the max of the lengths should be shorter than ``max_token_count``.
You can make it by ``datapipe_max_count = datapipe.filter(partial(_token_max_filter_fn, max_len=max_token_count))`` and verify if the length of datapipe is the same as before.

torchdata/datapipes/iter/transform/bucketbatcher.py
```diff
@@ -92,7 +104,116 @@ def __new__(
...
+                batch_size += length
+        while buffer:
+            length, token = heapq.heappop(buffer)
+            if batch_size + length > self.max_token_count:
+                if not batch:
+                    warnings.warn(
+                        "Empty batch is yielded due to length of a single token is larger than"
+                        "``max_token_count``. Please consider add ``max_len`` to filter such token or"
+                        "increase ``max_token_count``."
+                    )
```
nateanl, line 212: Same here.

torchdata/datapipes/iter/transform/bucketbatcher.py
ejguan: Okay, I see. Totally make sense to me.

torchdata/datapipes/iter/transform/bucketbatcher.py
```diff
@@ -10,18 +13,27 @@
...
 
 
 @functional_datapipe("bucketbatch")
 class BucketBatcherIterDataPipe(IterDataPipe[DataChunk[T_co]]):
     r"""
-    Creates mini-batches of data from sorted bucket (functional name: ``bucketbatch``). An outer
+    Creates mini-batches of data from sorted bucket. An outer
     dimension will be added as ``batch_size`` if ``drop_last`` is set to ``True``,
-    or ``length % batch_size`` for the last batch if ``drop_last`` is set to ``False``.
+    or ``length % batch_size`` for the last batch if ``drop_last`` is set to ``False``. (functional name: ``bucketbatch``).
```
NivekT, line 36: We generally need to keep "(functional name: ``bucketbatch``)" before the first `.` for the documentation to render correctly. Does the functional name show up in the table with this change?

torchdata/datapipes/iter/transform/bucketbatcher.py
```diff
@@ -92,7 +103,104 @@ def __new__(
...
+        return False
+    return True
+
+
+@functional_datapipe("max_token_bucketize")
+class MaxTokenBucketizerIterDataPipe(IterDataPipe[DataChunk[T_co]]):
+    r"""
+    Iterable DataPipe to create mini-batches of data from bucket, and the
+    total length of each batch will be limited by ``max_token_count``.
+    (functional name: ``max_token_bucketize``).
```
NivekT, line 139: Ditto about functional name in documentation.

torchdata/datapipes/iter/transform/bucketbatcher.py
```diff
@@ -92,7 +103,104 @@ def __new__(
...
+
+    Args:
+        datapipe: Iterable DataPipe being batched
+        max_token_count: Maximum length of total length of data in each batch
+        len_fn: Function to be applied to each element to get lengths. ``len(data)`` is used by default.
+        min_len: Optional minimum length to be included into each batch
+        max_len: Optional maximum length to be included into each batch.
+        buffer_size: This restricts how many tokens are taken from prior DataPipe to bucketize
+    """
```
NivekT, line 148: If you have time, can you add usage example (`Example:`) here

NivekT (commented):
We also need to add this DataPipe to `test_serialization.py` and the documentation `rst` file

torchdata/datapipes/iter/transform/bucketbatcher.py
ejguan: Let me do a quick experiment.

torchdata/datapipes/iter/transform/bucketbatcher.py
```diff
@@ -10,18 +13,27 @@
...
 
 
-def _in_batch_shuffle_fn(data: DataChunk):
-    d = list(data)
-    random.shuffle(d)
-    return DataChunk(d)
+@functional_datapipe("in_batch_shuffle")
+class InBatchShufflerIterDataPipe(IterDataPipe[DataChunk[T_co]]):
+    r"""
+    Shuffles each mini-batches from prior DataPipe (functional name: ``in_batch_shuffle``).
```
NivekT, line 19: Need `args` and `example` here

torchdata/datapipes/iter/transform/bucketbatcher.py
```diff
@@ -10,18 +13,27 @@
...
+    """
+
+    def __init__(self, datapipe: IterDataPipe[DataChunk[T_co]]):
+        self.datapipe = datapipe
+
+    def __iter__(self) -> Iterator[DataChunk[T_co]]:
+        for batch in self.datapipe:
+            random.shuffle(batch)
+            yield batch
```
NivekT, line 28: Can we add a `__len__` function too? Return from `self.datapipe` when available.

torchdata/datapipes/iter/transform/bucketbatcher.py
```diff
@@ -92,7 +103,104 @@ def __new__(
...
+    return True
+
+
+@functional_datapipe("max_token_bucketize")
+class MaxTokenBucketizerIterDataPipe(IterDataPipe[DataChunk[T_co]]):
+    r"""
+    Iterable DataPipe to create mini-batches of data from bucket, and the
+    total length of each batch will be limited by ``max_token_count``.
+    (functional name: ``max_token_bucketize``).
+
```
NivekT, line 140: @nateanl or @ejguan can we add a one sentence description here about what the practical use case of this DataPipe will be?

We also did the same for the other DataPipe in this file because it seems complicated at a first glance and the use case isn't immediately obvious.

torchdata/datapipes/iter/transform/bucketbatcher.py
```diff
@@ -92,7 +103,104 @@ def __new__(
...
+def _token_len_fn(token, len_fn):
+    return len_fn(token), token
+
+
+def _token_min_filter_fn(data, *, min_len):
+    length, token = data
+    if length < min_len:
+        return False
+    return True
```
NivekT, line 124: Super nit: `return length >= min_len`

torchdata/datapipes/iter/transform/bucketbatcher.py
```diff
@@ -92,7 +103,104 @@ def __new__(
...
+
+
+def _token_max_filter_fn(data, *, max_len):
+    length, token = data
+    if length > max_len:
+        return False
+    return True
```
NivekT, line 131: Super nit: `return length <= max_len`

torchdata/datapipes/iter/transform/bucketbatcher.py
```diff
@@ -92,7 +103,104 @@ def __new__(
...
+
+        if min_len < 0 or min_len > max_len:
+            raise ValueError("``min_len`` should be larger than 0 and equal to or smaller than ``max_len``.")
+        if max_len > max_token_count:
+            raise ValueError("``max_token_count`` must be equal to or greater than ``max_len``.")
+        datapipe = datapipe.map(partial(_token_len_fn, len_fn=len_fn))
+        if min_len is not None:
+            datapipe = datapipe.filter(partial(_token_min_filter_fn, min_len=min_len))
+        if max_len is not None:
+            datapipe = datapipe.filter(partial(_token_max_filter_fn, max_len=max_len))
```
NivekT, line 176: Super nit: Will it be more performant if we compare two steps into one filter?
```python
if min_len is not None or max_len is not None:
    datapipe.filter(partial(_token_len_filter_fn, min_len=min_len, max_len=max_len))
```

NivekT: I guess it depends on how `_token_len_filter` is implemented but it probably isn't faster.

torchdata/datapipes/iter/transform/bucketbatcher.py
```diff
@@ -92,7 +103,104 @@ def __new__(
...
+        self.buffer_size = buffer_size
+
+    def __iter__(self) -> Iterator[DataChunk[T_co]]:
+        buffer: List = []
+        batch: List = []
+        batch_size: int = 0
+        for d in self.datapipe:
+            heapq.heappush(buffer, d)
+            if len(buffer) == self.buffer_size:
+                length, token = heapq.heappop(buffer)
```
NivekT, line 190: Any particular reason why we are always returning the token with the minimum length? It would be good to mention why in the docstring

torchdata/datapipes/iter/transform/bucketbatcher.py
ejguan: Updated.

torchdata/datapipes/iter/transform/bucketbatcher.py
```diff
@@ -92,7 +119,109 @@ def __new__(
...
+    length, token = data
+    return length >= min_len and length <= max_len
+
+
+@functional_datapipe("max_token_bucketize")
+class MaxTokenBucketizerIterDataPipe(IterDataPipe[DataChunk[T_co]]):
+    r"""
+    Creates mini-batches of data from a min-heap with limited size, and the total length of samples
+    returned by ``len_fn`` within each batch will be limited by ``max_token_count``.
+    (functional name: ``max_token_bucketize``). If ``min_len`` or ``max_len`` is set, the samples with
```
NivekT, line 146: Nit: Does the docgen look right? It is still after the `.`

As long as the doc looks right, this is fine

torchdata/datapipes/iter/transform/bucketbatcher.py
```diff
@@ -92,7 +119,109 @@ def __new__(
...
+    length that is out of ``[min_len, max_len]`` will be filtered out.
+
+    The purpose of this DataPipe is to batch samples with similar length according to ``len_fn``.
+    Min-heap is used here to make sure the samples are sorted incrementally based on the length. And,
+    the total length of samples in each batch is guaranteed to be smaller than ``max_token_count`.
```
NivekT, line 151: Missing a ` in the end

NivekT (approved):
LGTM! Thanks!

torchdata/datapipes/iter/transform/bucketbatcher.py
ejguan: Good catch.

torchdata/datapipes/iter/transform/bucketbatcher.py
```diff
@@ -92,7 +119,109 @@ def __new__(
...
+        min_len: Optional minimum length to be included into each batch
+        max_len: Optional maximum length to be included into each batch.
+        buffer_size: This restricts how many tokens are taken from prior DataPipe to bucketize
+
+    Example:
+        >>> from torchdata.datapipes.iter import IterableWrapper
+        >>> source_dp = IterableWrapper(['1', '22', '1', '4444', '333', '1', '22', '22', '333'])
+        >>> batch_dp = source_dp.max_token_bucketize(max_token_count=5)
+        >>> list(batch_dp)
+        [['1', '1', '1', '22'], ['22', '22'], ['333'], ['333'], ['4444']]
```
nateanl, line 168: I feel like this example is kind of confusing, since the len_fn is missing and `len(str)` is used as it. Can you change it to use `1` for all samples (like `1`, `111`, `1111`), or add indication how to define the length of the string?

nateanl (approved):

torchdata/datapipes/iter/transform/bucketbatcher.py
ejguan: Sure, I can do that. I was using different number to represent the length of each string.

PR #285, opened 2022-03-09
Adding LimiterIterDataPipe

NivekT:
* __->__ #285

This DataPipe is the upstreamed version of `TakerDataPipe` in `torchvision`. I read through that implementation and made a best guess about what the desirable behavior should be.

https://github.com/pytorch/vision/blob/d32bc4ba9ac32ca413825a1a062455f1c747ce5f/torchvision/prototype/datasets/utils/_internal.py#L202-L232

This implementation is not final at the moment as I still have some questions about what this DataPipe should do.

### DIFF
--- a/torchdata/datapipes/iter/util/header.py
+++ b/torchdata/datapipes/iter/util/header.py
@@ -2,6 +2,9 @@
 from typing import Iterator, TypeVar
 from warnings import warn
 
+import torch.distributed as dist
+
+from torch.utils.data import get_worker_info
 from torchdata.datapipes import functional_datapipe
 from torchdata.datapipes.iter import IterDataPipe
 
@@ -53,3 +56,65 @@ class HeaderIterDataPipe(IterDataPipe[T_co]):
                 "The actual value may be smaller if the actual length of source_datapipe is smaller than the limit."
             )
             return self.limit
+
+
+@functional_datapipe("limit")
+class LimiterIterDataPipe(IterDataPipe[T_co]):
+    """
+    Limits the number of samples that the source DataPipe can yield based the input as well as the number of
+    processes and workers if distribute- or multi-processing is enabled (functional name: ``limit``).
+
+    The number of samples that will be yielded equals ``num_take // (num_processes * worker_info.num_workers)``.
+
+    Args:
+        source_datapipe: IterDataPipe with samples of data that will be yielded
+        num_take: the number of samples that you wish to yield across all processes and workers, should be less than
+            or equal to the ``len(source_datapipe)``
+
+    Example:
+        >>> source_dp = IterableWrapper(range(20))
+        >>> limiter_dp = source_dp.limit(5)
+        >>> list(limiter_dp)
+        [0, 1, 2, 3, 4]
+    """
+
+    def __init__(self, source_datapipe: IterDataPipe[T_co], num_take: int) -> None:
+        super().__init__()
+        self.source_datapipe: IterDataPipe[T_co] = source_datapipe
+        self.num_take: int = num_take
+        self.num_processes: int = 1
+        if dist.is_available() and dist.is_initialized():
+            self.num_processes = dist.get_world_size()
+
+    def _get_num_workers(self):
+        num_workers = self.num_processes
+        worker_info = get_worker_info()
+        if worker_info is not None:
+            # Note that we are assuming each distributed process has the same number of workers
+            num_workers *= worker_info.num_workers
+        return num_workers
+
+    def __iter__(self) -> Iterator:
+        # TODO: pmeier - this is weird as it drops more elements than it should
+        worker_num_take = self.num_take // self._get_num_workers()
+
+        for i, data in enumerate(self.source_datapipe):
+            # TODO: If each instance of the source_datapipe (of different worker/process) is not independently
+            #  shuffled, then all Limiter will return the same set of samples
+            #  Should we return samples based on process_id and worker_id?
+            #  Alternatively, should we optionally allow people to shuffle inside `__init__`?
+            if i < worker_num_take:
+                yield data
+            else:
+                break
+
+    # TODO: What is the length that user wants here
+    #       1) The total dataset length, or 2) the length that the specific worker/process will yield
+    def __len__(self) -> int:
+        max_num_take = self.num_take
+        try:
+            if len(self.source_datapipe) < self.num_take:
+                max_num_take = len(self.source_datapipe)
+        except TypeError:
+            pass
+        return max_num_take // self._get_num_workers()
--- a/docs/source/torchdata.datapipes.iter.rst
+++ b/docs/source/torchdata.datapipes.iter.rst
@@ -180,6 +180,7 @@ These DataPipes helps you select specific samples within a DataPipe.
 
     Filter
     Header
+    Limiter
 
 Text DataPipes
 -----------------------------
--- a/test/test_datapipe.py
+++ b/test/test_datapipe.py
@@ -329,6 +329,24 @@ class TestDataPipe(expecttest.TestCase):
             pass
         self.assertEqual(20, len(header_dp))
 
+    def test_limiter_iterdatapipe(self) -> None:
+        # Functional Test: ensure the limit is enforced
+        source_dp = IterableWrapper(range(20))
+        limiter_dp = source_dp.limit(5)
+        self.assertEqual(list(range(5)), list(limiter_dp))
+
+        # Functional Test: still works when the limit is greater than the length of the source
+        limiter_dp = source_dp.limit(30)
+        self.assertEqual(list(range(20)), list(limiter_dp))
+
+        # __len__ Test: returns the limit when it is less than the length of source
+        limiter_dp = source_dp.limit(5)
+        self.assertEqual(5, len(limiter_dp))
+
+        # __len__ Test: returns the length of source when it is less than the limit
+        limiter_dp = source_dp.limit(30)
+        self.assertEqual(20, len(limiter_dp))
+
     def test_enumerator_iterdatapipe(self) -> None:
         letters = "abcde"
         source_dp = IterableWrapper(letters)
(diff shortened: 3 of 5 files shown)


torchdata/datapipes/iter/util/header.py
```diff
@@ -53,3 +56,65 @@ def __len__(self) -> int:
...
+        self.num_take: int = num_take
+        self.num_processes: int = 1
+        if dist.is_available() and dist.is_initialized():
+            self.num_processes = dist.get_world_size()
+
+    def _get_num_workers(self):
+        num_workers = self.num_processes
+        worker_info = get_worker_info()
+        if worker_info is not None:
+            # Note that we are assuming each distributed process has the same number of workers
```
NivekT, line 93: Any thought on this assumption? Is it a reasonably safe assumption?

torchdata/datapipes/iter/util/header.py
```diff
@@ -53,3 +56,65 @@ def __len__(self) -> int:
...
+            num_workers *= worker_info.num_workers
+        return num_workers
+
+    def __iter__(self) -> Iterator:
+        # TODO: pmeier - this is weird as it drops more elements than it should
```
NivekT, line 98: @pmeier Can you elaborate on this?

torchdata/datapipes/iter/util/header.py
```diff
@@ -53,3 +56,65 @@ def __len__(self) -> int:
...
+
+        for i, data in enumerate(self.source_datapipe):
+            # TODO: If each instance of the source_datapipe (of different worker/process) is not independently
+            #  shuffled, then all Limiter will return the same set of samples
+            #  Should we return samples based on process_id and worker_id?
+            #  Alternatively, should we optionally allow people to shuffle inside `__init__`?
+            if i < worker_num_take:
+                yield data
+            else:
+                break
```
NivekT, line 109: This behavior is worth mentioning, I think the options are:
1. This behavior is expected and should be kept as it is, and user should shuffle ahead of time if they want something different.
2. Provide an option to shuffle the `source_datapipe` within `__init__`

torchdata/datapipes/iter/util/header.py
```diff
@@ -53,3 +56,65 @@ def __len__(self) -> int:
...
+
+    # TODO: What is the length that user wants here
+    #       1) The total dataset length, or 2) the length that the specific worker/process will yield
+    def __len__(self) -> int:
```
NivekT, line 113: The implementation currently aims to return 2.

NivekT:
cc: @NicolasHug @fmassa since you may know more about what the ideal behavior for this DataPipe should be.

torchdata/datapipes/iter/util/header.py
ejguan: We should do the 1 option. This is the limit rather than sampling. Users need to take care of shuffling.

torchdata/datapipes/iter/util/header.py
ejguan: 2 please

torchdata/datapipes/iter/util/header.py
```diff
@@ -53,3 +56,65 @@ def __len__(self) -> int:
...
+        [0, 1, 2, 3, 4]
+    """
+
+    def __init__(self, source_datapipe: IterDataPipe[T_co], num_take: int) -> None:
+        super().__init__()
+        self.source_datapipe: IterDataPipe[T_co] = source_datapipe
+        self.num_take: int = num_take
+        self.num_processes: int = 1
+        if dist.is_available() and dist.is_initialized():
+            self.num_processes = dist.get_world_size()
```
pmeier, line 87: I'm a little worried about this bit. Is it a safe assumption that a user wants to use distributed if it is initialized?

torchdata/datapipes/iter/util/header.py
```diff
@@ -53,3 +56,65 @@ def __len__(self) -> int:
...
+            or equal to the ``len(source_datapipe)``
+
+    Example:
+        >>> source_dp = IterableWrapper(range(20))
+        >>> limiter_dp = source_dp.limit(5)
+        >>> list(limiter_dp)
+        [0, 1, 2, 3, 4]
+    """
+
+    def __init__(self, source_datapipe: IterDataPipe[T_co], num_take: int) -> None:
```
pmeier, line 81: The name `num_take` was somewhat tied to the name `TakerDataPipe`. Is this still expressive enough?

torchdata/datapipes/iter/util/header.py
pmeier: I can't since I've not written nor ever used it. cc @fmassa for context.

torchdata/datapipes/iter/util/header.py
NicolasHug: If we're going for `LimiterIterDataPipe`, `limit` might be adequate

NicolasHug (commented):
Looks like this class does 2 different things:

- limit a datapipe to a certain length
- handles DDP **and** dataloader multiprocessing.

Should we dissociate these 2 things?

torchdata/datapipes/iter/util/header.py
pmeier: Might be a little awkward though in the functional interface

```py
dp.limit(limit=...)`
```

NivekT:
> Looks like this class does 2 different things:
> 
> * limit a datapipe to a certain length
> * handles DDP **and** dataloader multiprocessing.
> 
> Should we dissociate these 2 things?

For the first one, `HeaderIterDataPipe` in the same file should cover that functionality. Let me know if the implemetation is insufficient.

https://github.com/pytorch/data/blob/823d216535782a2c31a19bcfaeb450597a53f8c4/torchdata/datapipes/iter/util/header.py#L11-L26

This new `LimiterIterDataPipe`, in its current form, only takes it one step further. It doesn't just use the input argument `num_take` (or `limit`), it tries to account for how many processes/workers there are, and set the true limit to be `num_take // num_workers`. Is that what your library need?

Can you elaborate more on "handles DDP **and** dataloader multiprocessing"?

pmeier:
> For the first one, `HeaderIterDataPipe` in the same file should cover that functionality. Let me know if the implemetation is insufficient.

I think for my use case this is sufficient. I'll open an issue with a related feature request. Otherwise, I'll let the others the rest, since I don't have enough background for it.

torchdata/datapipes/iter/util/header.py
ejguan: I think it's still working without such assumption because each distributed process should get corresponding `worker_info` from on the specific distributed process.

ejguan:
I have another question for DPP distributed use case.
Do we want to have a limit over the total amount of data for **ALL** distributed process or we want to have a limit over the amount of data of a **SINGLE** distributed process?
IIUC, current implementation reflects the first option. For the latter option, we then won't need to worry about if `dist` is initialized or not. We would only take care the total amount of data across workers.

fmassa (commented):
Hi,

Just to chime in in the reason why I made `TakerIterDataPipe` support DDP + multiprocess as well.

The original reason is the following: when doing DDP, we need to make sure that the number of batches seen on every process is the same (otherwise we can have deadlocks due to the distributed communication in DDP). This is handled originally in the `DistributedSampler` https://github.com/pytorch/pytorch/blob/9b203f667ac096db9f5f5679ac3e3d7931c34d36/torch/utils/data/distributed.py#L104-L114

With `DataPipes`, we can do a similar approach as well, but one complication is that we don't know in principle what is the number of elements in the datapipe, so we can't follow exactly the same approach as `DistributedSampler`.

Thus, when implementing `TakerIterDataPipe` (which specifies a size for the datapipe), I decided that I would also add the behavior of `DistributedSampler` inside it so that things work by default.

Now, it is totally possible to factor out this into a separate DataPipe that performs this operation, but then we need to guarantee that the datapipe has a valid `__len__`.

The above answers the reason for adding DDP-specific information. But the reason for the multiprocessing-specific information is slightly different.

For multiprocessing, the reason why I use it is because with vanilla DataLoader, if you limit the datapipe size by `N` and use `K` workers in the DataLoader, then the effective dataset size would end up being `N * K`, which is most definitely not expected.

So combining both together seemed like the simplest thing so that using DataPipes in real scenarios with DDP + multiple workers to work as expected. But there are some edge cases as of now, as we indeed drop more elements than needed in principle (which was done for simplicity). In the `DistributedSampler` case, instead of dropping elements they pad the dataset with repeated elements, which I think would be a potentially preferred solution.

Happy to brainstorm other possible solutions

torchdata/datapipes/iter/util/header.py
fmassa: This is a similar heuristic as what we do in `DistributedSampler` https://github.com/pytorch/pytorch/blob/89d6f3e6094f36adade286e85b5c3b7d05c4003b/torch/utils/data/distributed.py#L63-L70

As I mentioned in another comment, this can be factored out in a different element, but there are also other things to consider

torchdata/datapipes/iter/util/header.py
fmassa: The reason that it drops more elements than it should is because we just need to make sure that every DDP process has the same number of elements, and not every individual worker thread in the dataloader.

This is what `DistributedSampler` does, as the DataLoader properly takes care of `worker_threads` with different number of elements.

torchdata/datapipes/iter/util/header.py
fmassa: Note that, due to https://github.com/pytorch/data/pull/285/files#r824693825 , the length of this datapipe will not be exactly equal to the number of returned elements in the DataLoader. But maybe this is fine?

torchdata/datapipes/iter/util/header.py
VitalyFedyunin: Please add TODO to pull this information from the ReadingService

VitalyFedyunin:
Lets discuss it offline, as I would love to see integration with sharding here, instead of pulling info from the global context.

NivekT:

PR #286, opened 2022-03-10
[Release 0.3.0] Enable official release

ejguan:
Example workflow without uploading: https://github.com/pytorch/data/runs/5498280111?check_suite_focus=true

Conda is still missing because of the missing token.


NivekT (approved):
LGTM! Thanks!!!

PR #287, opened 2022-03-10
Change PyTorch path to official PYPI

ejguan:
Per title


.github/workflows/_build_test_upload.yml
```diff
@@ -82,12 +82,14 @@ jobs:
           else
             PYTORCH_VERSION=torch
           fi
-          # Use test channel for official release
```
ejguan, line 85: I was using test channel, which was incorrect.

NivekT (approved):

PR #290, opened 2022-03-10
[BE Hackathon] Automatically generate datapipe.pyi in setup.py

NivekT:
* #304
* __->__ #290

This should only land after https://github.com/pytorch/pytorch/pull/73991

~~CI will fail until that PR is landed into Core's nightly.~~ This is working now.

Automatically run the generation of `datapipe.pyi` in `setup.py` and remove the interface (.pyi) file from version control

Differential Revision: [D34939044](https://our.internmc.facebook.com/intern/diff/D34939044)


NivekT:
Same question as the other PR:
Should we keep the .pyi file in Git VCS or not?

NivekT:
@nivekt has imported this pull request.  If you are a Facebook employee, you can view this diff [on Phabricator](https://www.internalfb.com/diff/D34939044).

.github/workflows/domain_ci.yml
```diff
@@ -125,6 +125,9 @@ jobs:
       - name: Check out torchdata repository
         uses: actions/checkout@v2
 
+      - name: Install mssing dependency
```
NivekT, line 128: Other domain libraries generally require and install `requests` (which is required by `torchdata`) but not `torchaudio`. Therefore we are installing it here prior to installing `torchdata`.

https://github.com/pytorch/audio/blob/main/setup.py

NivekT:
@nivekt has imported this pull request.  If you are a Facebook employee, you can view this diff [on Phabricator](https://www.internalfb.com/diff/D34939044).

.github/workflows/domain_ci.yml
NivekT: Note that I fixed the typo

.github/workflows/domain_ci.yml
```diff
@@ -125,6 +125,9 @@ jobs:
       - name: Check out torchdata repository
         uses: actions/checkout@v2
 
+      - name: Install missing dependency
+        run: pip install requests
+
       - name: Install torchdata
         run: python setup.py install
```
ejguan, line 132: Thanks for fixing it. Can you move this part into `Install torchdata` with `pip install -r requirements.txt`?

ejguan (approved):
LGTM with one comment about GHA workflow.

pmeier:
This breaks [`torchvision` CI](https://app.circleci.com/pipelines/github/pytorch/vision/15927/workflows/f4145728-c5a8-4513-b36c-80044325204d/jobs/1288641). Since there is no nightly release that I'm aware of, `torchvision` installs `torchdata` with `pip install git+https://github.com/pytorch/data`.

This PR is problematic, because you now have a build time dependency on `requirements.txt` as well as `torch` that is not picked up by `pip`.  Thus, we need to now install with

```shell
pip install -r https://raw.githubusercontent.com/pytorch/data/main/requirements.txt
pip install --no-build-isolation git+https://github.com/pytorch/data
```

ejguan:
> Since there is no nightly release that I'm aware of, torchvision installs torchdata with pip install git+https://github.com/pytorch/data.

I am adding a nightly release by the end of this week. Does it help if nightly release is available?

pmeier:
> Does it help if nightly release is available?

Yes. Installing from a git repo means that we are installing from source. Thus, we need to have all build requirements available without `pip` knowing what they are. If you have a nightly release, we can simply install the binary and `pip` automatically picks up all runtime requirements.

ejguan:
> > Does it help if nightly release is available?
> 
> Yes. Installing from a git repo means that we are installing from source. Thus, we need to have all build requirements available without `pip` knowing what they are. If you have a nightly release, we can simply install the binary and `pip` automatically picks up all runtime requirements.

Sure. Then, I will try to land nightly workflow asap

PR #300, opened 2022-03-16
Enable release workflow for both conda and wheel and enable nightly release

ejguan:
Changes:
- Remove redundant logic to compare commits between release wheels
- Add `version.txt` as the single source of truth for the version of TorchData.
  - Reason: Conda needs the version without running `python setup.py ...`. Without this PR, the version of TorchData is determined by running `python setup.py ...`.
- Fix a bug in `setup.py` that pytorch version is incorrect
- Add `packaging/build_conda.sh` and `packaging/build_wheel.sh` and related utility for TorchData to build packages and wheels. (taking reference from TorchVision and TorchText)
  - Files in `packaging/torchdata/...` are required by conda-build
  - Reason: Conda won't to run `python setup.py ...`
  - Remove --release/--nightly option for `setup.py` as we would use the script above to do so.
- Add conda package build/upload workflow
- Enable nightly build (Example: https://github.com/pytorch/data/runs/5647039889?check_suite_focus=true)
  - Validated the version number for nightly release (wheel/conda)
  - Validated the dependency for nightly release (wheel/conda)


.github/workflows/_build_test_upload.yml
```diff
@@ -52,153 +52,91 @@ jobs:
...
           - ubuntu-latest
-          # - windows-latest
+          - windows-latest
         python-version:
-          # - 3.7
-          # - 3.8
+          - 3.7
+          - 3.8
           - 3.9
+          - "3.10"
```
ejguan, line 70: Turn on matrix for the coming S3 PR.

.github/workflows/_build_test_upload.yml
```diff
@@ -52,153 +52,91 @@ jobs:
...
-            echo "::set-output name=value::true"
-          else
-            echo 0 > torchdata_${{ matrix.os }}_${{ matrix.python-version }}_whl.txt
-            echo "::set-output name=value::false"
-          fi
-        id: trigger_build
-      # The following steps for pre-release will be skipped if the last commit
-      # from branch is the same as the last commit in release wheel
-      - name: Build and Install TorchData Wheel
-        if: steps.trigger_build.outputs.value == 'true'
```
ejguan, line 132: Removing this part to check if we need to create wheel if there is no change in TorchData because I figured out PyTorch core may have a new nightly release even though TorchData doesn't have any change. In order to provide the up-to-date nightly release for both PyTorch and TorchData, this part of workflow should be removed.

.github/workflows/_build_test_upload.yml
```diff
@@ -52,153 +52,91 @@ jobs:
...
-          if [[ ${{ needs.get_release_type.outputs.type }} == 'nightly' ]]; then
-            python setup.py bdist_wheel --nightly
-          else
-            python setup.py bdist_wheel --release
-          fi
-          pip3 install dist/torchdata*.whl
+        run: packaging/build_wheel.sh
+      - name: Install TorchData Wheel
+        shell: bash
+        run: pip3 install dist/torchdata*.whl
```
ejguan, line 88: Using script to build TorchData wheels.

.github/workflows/nightly_release.yml
```diff
@@ -1,24 +1,17 @@
...
-  # 2. In the left sidebar, click the workflow you want to run
-  # 3. Above the list of workflow runs, select Run workflow
-  # 4. Use the Branch dropdown to select the release/* branch
-  # 5. Click Run workflow
-  workflow_dispatch:
-  # TODO: Automatically trigger nightly release
-  # schedule:
-  #   - cron: 30 23 * * *
+  schedule:
+    - cron: 30 23 * * *
```
ejguan, line 5: Turn on nightly release. 

For Doc generation, I think there are two options for us:
- Create another folder called `nightly` just for nightly release
- Disable the doc generation workflow that we are using in main branch, and enable the doc gen workflow whenever a nightly release is uploaded.

cc: @NivekT

packaging/README.md
```diff
@@ -0,0 +1,31 @@
+# Build TorchData Release
```
ejguan, line 1: The rest of packaging scripts are mimicked from TorchText/TorchVision

setup.py
```diff
@@ -15,25 +12,21 @@
...
-        os_build_version = os.getenv("BUILD_VERSION")
-        if os_build_version:
-            version = os_build_version
-        elif sha != "Unknown":
-            version += "+" + sha[:7]
+    os_build_version = os.getenv("BUILD_VERSION")
+    if os_build_version:
+        version = os_build_version
+    elif sha != "Unknown":
+        version += "+" + sha[:7]
```
ejguan, line 29: Let packaging script to determine release/nightly because of conda.

setup.py
```diff
@@ -45,26 +38,10 @@ def _export_version(version, sha):
...
-        help="Official/RC Release",
-    )
-    return parser
-
-
 # Use new version of torch on main branch
 pytorch_package_dep = "torch>1.11.0"
 if os.getenv("PYTORCH_VERSION"):
-    pytorch_package_dep = pytorch_package_dep.split(">=")[0]
+    pytorch_package_dep = pytorch_package_dep.split(">")[0]
```
ejguan, line 44: bug

version.txt
```diff
@@ -0,0 +1 @@
+0.4.0a0
```
ejguan, line 1: This is going to be our single source of truth for version

ejguan (commented):
Adding a few comments to help reviewing this PR.

ejguan:
It seems the CI of doc gen is still failing. I plan to disable the doc gen after nightly release is successfully uploaded.

NivekT:
> It seems the CI of doc gen is still failing. I plan to disable the doc gen after nightly release is successfully uploaded.

Looks like the initial commit fix worked but [this update](https://github.com/pytorch/data/pull/314/commits/0d36e31405838946960aae047010a8f2b27714e4) broke it. I am investigating now.

ejguan:
> > It seems the CI of doc gen is still failing. I plan to disable the doc gen after nightly release is successfully uploaded.
> 
> Looks like the initial commit fix worked but [this update](https://github.com/pytorch/data/pull/314/commits/0d36e31405838946960aae047010a8f2b27714e4) broke it. I am investigating now.

Thanks! Let me know your opinion on https://github.com/pytorch/data/pull/300#discussion_r832408371 after fixing it.

.github/workflows/nightly_release.yml
NivekT: I think the second one (only generate doc for nightly) is fine. We likely don't need to re-generate the doc for every single commit to main.

ejguan:
This PR is ready for review.

.github/workflows/nightly_release.yml
ejguan: Sounds great. Will wait until the nightly works.

.github/workflows/_build_test_upload.yml
```diff
@@ -34,12 +34,10 @@ jobs:
     steps:
       - name: Get Release Type
         run: |
-          if [[ ${{ inputs.branch }} == release/* ]]; then
-            if [[ ${{ inputs.pre_dev_release }} == true ]]; then
-              RELEASE_TYPE=test
-            else
-              RELEASE_TYPE=official
-            fi
+          if [[ ${{ inputs.branch }} == v* ]] && [[ ${{ inputs.pre_dev_release }} == false ]]; then
```
NivekT, line 37: Should this be `release/*` instead of `v*`?

NivekT (approved):
Only small comments. Great work! Thanks!!

.github/workflows/_build_test_upload.yml
ejguan: For official release, we should use tags rather than release branch. In order to retrieve tags, the input branch for official release has to be something like `v0.3.0`.

PR #308, opened 2022-03-19
Add tfrecord_loader implementation

### LINKED ISSUE #306: Add tfrecord support
### 🚀 The feature

Please add tfrecord support. This could be implemented as a "TFRecordLoader" similar to "TarArchiveLoader". It would load the tfrecord file and parse the records.

### Motivation, pitch

A lot of TensorFlow users have their datasets stored as tfrecords for efficient data loading. It would be great to support tfrecords to be able to use the same data format for both frameworks. This would make it easier to transfer TensorFlow models to PyTorch ecosystem and vice versa.

### Alternatives

There is the following project: https://github.com/vahidk/tfrecord which works great. It would be great, however, if we allow something similar with the new data API.

### Additional context

_No response_

jkulhanek:
Fixes #306

### Changes
- Added `TFRecordLoaderIterDataPipe` that parses opened tfrecord streams and yields the individual records.  It supports both `tf.train.Example` and `tf.train.SequenceExample` and users can additionally specify shape/dtype for each feature in the dict.
- Added tests for  `TFRecordLoaderIterDataPipe`.


VitalyFedyunin:
Can you please drop `tensorflow` dependency and leave only `protobuf`. It is ok if you would need to commit binaries that are required for testing directly into the repo (please keep them within 100kb).

jkulhanek:
I removed the TensorFlow dependency.

jkulhanek:
Any update?

torchdata/datapipes/iter/util/tfrecordloader.py
```diff
@@ -0,0 +1,241 @@
...
+    ) -> None:
+        super().__init__()
+        _assert_protobuf()
+
+        self.datapipe: Iterable[Tuple[str, BufferedIOBase]] = datapipe
+        self.length: int = length
+        self.spec = spec
+
+    def __iter__(self) -> Iterator[Example]:
+        from . import _tfrecord_example_pb2 as example_pb2
```
VitalyFedyunin, line 224: I believe various TFRecords might have different protibufs, for this you probably need to pass protobuf_schema as parameter of pipe, something like:

```
from . import _tfrecord_example_pb2 as example_pb2
tfrecord_loader_dp = datapipe2.load_from_tfrecord(example_pb2)
```

torchdata/datapipes/iter/util/_tfrecord_example_pb2.py
```diff
@@ -0,0 +1,698 @@
+# Generated by the protocol buffer compiler.  DO NOT EDIT!
```
VitalyFedyunin, line 1: This file should be generated, not part of git repo. Instead please put the original protobuf file + generation script (inside of tests folder).

torchdata/datapipes/iter/util/tfrecordloader.py
jkulhanek: The protobuf schema for Example was kept the same in TF so far. If they changed it, TF would not load older tfrecord files. Therefore, I believe there is no need to pass the protobuf schema explicitly.

torchdata/datapipes/iter/util/_tfrecord_example_pb2.py
jkulhanek: I think we would need to include several protobuf files (example.proto depends on other files). Should the file be generated in setup.py?

NivekT: If there are only a few protobuf files, we can add them to the library without having them generated.

Do you know how many there are?

torchdata/datapipes/iter/util/tfrecordloader.py
NivekT: Since the schema has been consistent I don't think we will need it as a parameter, can you add a comment there stating that is the assumption.

torchdata/datapipes/iter/util/_tfrecord_example_pb2.py
jkulhanek: I think there are two: "example.proto" and "feature.proto" https://github.com/tensorflow/tensorflow/tree/master/tensorflow/core/example

jkulhanek: But are you sure the "tests" folder is a good fit for ".proto" files?

torchdata/datapipes/iter/util/tfrecordloader.py
jkulhanek: I added the comment.

torchdata/datapipes/iter/util/_tfrecord_example_pb2.py
NivekT: No, let's not put them in the `tests` folder, but somewhere within the main folder that actually get distributed. Let me take another holistic look at this PR and get back to you.

NivekT (commented):
Overall, looks good! I left a few comments above. And:

1. Can you rebase to the latest origin/main? You might need to update to header from "Facebook" to "Meta Platforms"
2. Add the DataPipe to `test_serialization`?
3. Run `flake8` and `mypy` on each of the `.py` file

Thank you so much for working with us on this PR!

torchdata/datapipes/iter/util/_tfrecord_example_pb2.py
NivekT: @jkulhanek Can we generate this file once the first time the DataPipe is called (and just use it and not re-generate in the subsequent usages)?

In terms of location, I think `torchdata/datapipes/iter/util/protobuf_template` should be appropriate.

torchdata/datapipes/iter/util/tfrecordloader.py
NivekT: nit (mostly for discussion): I understand that TF consistently has the same `protobuf` schema, but we can potentially make this a generic `ProtobufLoaderDataPipe` by accepting schema as an argument (but let the TF one be the default).

Will this be useful for users? What other `protobuf` schema will they use? Or should we just stick with TFRecord only for now? @VitalyFedyunin

jkulhanek: I would stick with TFRecord only. In the TFRecordLoader we assume the data are Examples in several places. This allows us to load the data in a more "user-friendly" way (e.g. we load FloatList data as tensors and allow the user to reshape them). Therefore, I think it would be better to keep ProtobufLoader separately if there was a need for it.

torchdata/datapipes/iter/util/_tfrecord_example_pb2.py
jkulhanek: This also means we would have to take care of locking (multiple threads/processes could be writing to the same file). This could likely happen in distributed training. Can't we generate it when shipping the package such that the generated file would be a part of it?

NivekT: In that case, putting it in `setup.py` is fine. Does the generation require extra dependencies?

torchdata/datapipes/iter/util/tfrecordloader.py
NivekT: I agree that supporting TFRecord for now but our team worries that we may run into trademark issues if we choose to name it something related to TF. Can you please rename it to `ProtobufRecordLoader` but mention in the description that it, by default, supports loading of TF records?

jkulhanek: This is a problem. "tfrecord" file describes a sequence of binary strings with headers and protobuf-searialized binary payload. The way the protobuf-serialized records are concatenated in the final tfrecord file is not a part of the protobuf specification. A simple protobuf deserialization could not load a tfrecord file (and would not be efficient, since the whole file would be loaded at once and not sequentially).

jkulhanek: The "tensorboard" name was used in the PyTorch repo. Could we do something similar in this case? Is the "tfrecord" specification legally protected?

torchdata/datapipes/iter/util/_tfrecord_example_pb2.py
NivekT: What about we raise an error message with instructions (and a script) to generate the file within the library if the file is missing? I think that avoids add the generation to `setup.py` and the multiprocessing issue.

cc: @VitalyFedyunin

torchdata/datapipes/iter/util/tfrecordloader.py
NivekT: It seems like we may want to keep the `TFRecord` name, @VitalyFedyunin what do you think?

torchdata/datapipes/iter/util/_tfrecord_example_pb2.py
jkulhanek: Can't we just leave the generated file in the package? It seems to me like the easiest option and we would not require users to install protobuf compiler before using the loader.

VitalyFedyunin: Agree to leave it as is as soon as we place source .proto files besides it.

torchdata/datapipes/iter/util/tfrecordloader.py
NivekT: We are checking with our legal team about this. Will update here when I have more info.

NivekT: We are in the clear with naming and etc. No need to change any names here.

torchdata/datapipes/iter/util/_tfrecord_example_pb2.py
NivekT: Agree, we should add the generated files and the source .proto files as well. Somewhere like `torchdata/datapipes/iter/util/protobuf_template`.

jkulhanek:
Thank you for your help with the PR!
I have updated the files according to your comments. After merging the main branch, one of the tests is still failing on class "torchdata.datapipes.iter.util.unzipper.UnZipperIterDataPipe". Otherwise, it looks good. Both mypy and flake8 report no errors in the added code.

NivekT (approved):
LGTM!

jkulhanek:
I merged the main branch, all tests are passing locally now (except for torchtext and torchaudio - these I dit not run).

VitalyFedyunin:
Let's forward fix it, if it possible to do quickly.

PR #311, opened 2022-03-20
Deprecate transforms.RandomSizedCrop in torchvision example

xiaohu2015:
Fixes #{issue number}

### Changes

-
-


ejguan (approved):
Thanks for fixing it.

NivekT (approved):

ejguan:
@NivekT Just want to confirm the mypy Error is caused by the change in PyTorch Core and not related to the PR.

PR #325, opened 2022-03-24
Adding UnZipperMapDataPipe

NivekT:
* #329
* #328
* #327
* #326
* __->__ #325

This `MapDataPipe` seems simple enough to add but we should talk about what is the general guideline for adding `MapDataPipe` before actually doing it.

Differential Revision: [D35979906](https://our.internmc.facebook.com/intern/diff/D35979906)


NivekT:
@VitalyFedyunin @ejguan I updated this whole stack and it is ready to review, but not urgent. Look at this stack of PR if you have time.

ejguan:
> @VitalyFedyunin @ejguan I updated this whole stack and it is ready to review, but not urgent. Look at this stack of PR if you have time.

I will take a look tmrw.

test/test_datapipe.py
```diff
@@ -18,7 +18,6 @@
 import torchdata
```
ejguan, line 18: Could you change the name of this test to `test_iterdatapipe.py`?

torchdata/datapipes/map/util/unzipper.py
```diff
@@ -0,0 +1,73 @@
+# Copyright (c) Facebook, Inc. and its affiliates.
+from typing import Optional, Sequence, TypeVar
+
+from torch.utils.data.datapipes.datapipe import MapDataPipe
+
+
+T = TypeVar("T")
+
+
+# @functional_datapipe("unzip")  # This line must be kept for .pyi signature parser
```
ejguan, line 10: Why the `functional_datapipe` is commented out?

test/test_mapdatapipe.py
```diff
@@ -0,0 +1,45 @@
...
+        dp1, dp2, dp3 = UnZipper(source_dp, sequence_length=3)  # type: ignore[misc]
+        self.assertEqual(list(range(10)), list(dp1))
+        self.assertEqual(list(range(10, 20)), list(dp2))
+        self.assertEqual(list(range(20, 30)), list(dp3))
+
+        # Functional Test: unzips each sequence, with `sequence_length` specified
+        dp1, dp2, dp3 = source_dp.unzip(sequence_length=3)
+        self.assertEqual(list(range(10)), list(dp1))
+        self.assertEqual(list(range(10, 20)), list(dp2))
+        self.assertEqual(list(range(20, 30)), list(dp3))
```
ejguan, line 25: Emmm. What is the difference between this test and the test above?

torchdata/datapipes/map/util/unzipper.py
```diff
@@ -0,0 +1,73 @@
...
+        self.instance_id = instance_id
+
+    def __getitem__(self, index) -> T:
+        return self.main_datapipe[index][self.instance_id]
+
+    def __len__(self) -> int:
+        return len(self.main_datapipe)
+
+
+MapDataPipe.register_datapipe_as_function("unzip", UnZipperMapDataPipe)
```
ejguan, line 73: Emmm. I completely forget why we need to do this for `MapDataPipe` rather than using `functional_datapipe`

ejguan (approved):
Overall, looks good to me. Thank you

torchdata/datapipes/map/util/unzipper.py
ejguan: This is the last time I used this similar approach because I want to register an operation to `IterDataPipe` that returns a `MapDataPipe` here:
https://github.com/pytorch/data/blob/b6ade8f097bc9ac08460cd403034a35daff09cfa/torchdata/datapipes/map/util/utils.py#L115

torchdata/datapipes/map/util/unzipper.py
NivekT: `UnZipperMapDataPipe` actually doesn't implement the `MapDataPipe` protocol since it doesn't have a `__getitem__`. Its child `_UnZipperMapDataPipe` does. Therefore, it can't be registered directly.
The following line in the bottom must be used instead.
`MapDataPipe.register_datapipe_as_function("unzip", UnZipperMapDataPipe)`

torchdata/datapipes/map/util/unzipper.py
NivekT: See the other comment. It is just that `UnZipper` in this case actually doesn't implement `MapDataPipe`'s protocol (doesn't have `__getitem__`).

torchdata/datapipes/map/util/unzipper.py
ejguan: I think what `functional_datapipe` does is the same as `MapDataPipe.register_datapipe_as_function("unzip", UnZipperMapDataPipe)`. `functional_datapipe` is just a syntax sugar, right?
Like `UnZipperIterDataPipe`, we use the `functional_datapipe` as well even though it's just a DataPipe to return multiple `ChildDataPipe`. https://github.com/pytorch/data/blob/b6ade8f097bc9ac08460cd403034a35daff09cfa/torchdata/datapipes/iter/util/unzipper.py#L17-L18

NivekT: I double-checked and you are right. It works even when `UnZipper` doesn't have all the methods of `MapDataPipe`

NivekT:
@nivekt has imported this pull request.  If you are a Facebook employee, you can view this diff [on Phabricator](https://www.internalfb.com/diff/D35979906).

PR #326, opened 2022-03-24
Small refactor to ensure 'to_map_datapipe' is generated as a method in IterDataPipe's pyi interface

NivekT:
Stack from [ghstack](https://github.com/ezyang/ghstack):
* #329
* #328
* #327
* __->__ #326
* #325

Previously, `.to_map_datapipe` didn't show up as a hint due to how the file crawling/parsing work in `gen_pyi.py`. This fixes it.

Differential Revision: [D35979907](https://our.internmc.facebook.com/intern/diff/D35979907)

Issue #200, opened 2022-02-08
Register Functional API to different DataPipe


ejguan:
### 🚀 The feature

```py
@functional_datapipe("func")
class IterToMapMapDataPipe(MapDataPipe):
    def __init__(self, datapipe: IterDataPipe):
        self.datapipe = datapipe
```
Using the `functional_datapipe` above would register this functional API to `MapDataPipe`. As we want to register this functional_datapiep to the `IterDataPipe`, I have to do the following trick.

```py
class IterToMapMapDataPipe(MapDataPipe):
    def __init__(self, datapipe: IterDataPipe):
        self.datapipe = datapipe

IterDataPipe.register_datapipe_as_function("func", IterToMapMapDataPipe)
```

### Motivation, pitch

Option 1:
Use the type annotation of the first input datapipe from `__init__` or `__new__` function to determine which DataPipe class we should register.

```py
class IterToMapMapDataPipe(MapDataPipe):
    def __init__(self, datapipe: IterDataPipe):  # <---------- IterDataPipe here is used to register
        self.datapipe = datapipe
```

Option 2:
Add an optional argument for `functional_datapipe` API to explicitly set the right DataPipe class.
```py
@functional_datapipe("func", register_cls=IterDataPipe)
class IterToMapMapDataPipe(MapDataPipe):
    def __init__(self, datapipe: IterDataPipe):
        self.datapipe = datapipe
```

NivekT:
I personally like Option 2 better. It is more obvious what is being registered to which class.

PR #327, opened 2022-03-24
Adding MapToIterConverter

NivekT:
* #329
* #328
* __->__ #327
* #326
* #325

I think this converter is necessary regardless if we decide to add more `MapDataPipe`.

Differential Revision: [D35979908](https://our.internmc.facebook.com/intern/diff/D35979908)


torchdata/datapipes/map/util/converter.py
```diff
@@ -0,0 +1,23 @@
...
+from torch.utils.data import IterDataPipe, MapDataPipe
+
+
+# @functional_datapipe("to_iter_datapipe")  # This line must be kept for .pyi signature parser
+class MapToIterConverterIterDataPipe(IterDataPipe):
+    def __init__(self, datapipe: MapDataPipe, indices: Optional[List] = None):
+        if not isinstance(datapipe, MapDataPipe):
+            raise TypeError(f"MapToIterConverter can only apply on MapDataPipe, but found {type(datapipe)}")
+        self.datapipe: MapDataPipe = datapipe
+        self.indices = indices if indices else range(len(datapipe))
```
NivekT, line 13: Assume 0-index when `indices` are not given.

torchdata/datapipes/map/util/converter.py
```diff
@@ -0,0 +1,23 @@
+# Copyright (c) Facebook, Inc. and its affiliates.
+from typing import List, Optional
+
+from torch.utils.data import IterDataPipe, MapDataPipe
+
+
+# @functional_datapipe("to_iter_datapipe")  # This line must be kept for .pyi signature parser
+class MapToIterConverterIterDataPipe(IterDataPipe):
+    def __init__(self, datapipe: MapDataPipe, indices: Optional[List] = None):
```
NivekT, line 9: Any thoughts on requiring `indices` to be a `List` rather than say `Iterable`?

Taking in an `Iterable` complicates things when we have to serialize since in theory a generator can be passed. Then, we will have to materialize it, but that is only possible from we haven't started reading the generator already.

VitalyFedyunin, line 9: indexes should be Iterable, not necessary the List

torchdata/datapipes/map/util/converter.py
```diff
@@ -0,0 +1,23 @@
+# Copyright (c) Facebook, Inc. and its affiliates.
+from typing import List, Optional
+
+from torch.utils.data import IterDataPipe, MapDataPipe
+
+
+# @functional_datapipe("to_iter_datapipe")  # This line must be kept for .pyi signature parser
```
VitalyFedyunin, line 7: Wait, why it is not working? Do we need to update functional_datapipe to determine the proper DP type?

torchdata/datapipes/map/util/converter.py
NivekT: Yea, I was thinking about that. From my other comment:

Any thoughts on requiring `indices` to be a `List` rather than say `Iterable`?

Taking in an `Iterable` complicates things when we have to serialize since in theory a generator can be passed. Then, we will have to materialize it, but that is only possible from we haven't started reading the generator already."

torchdata/datapipes/map/util/converter.py
NivekT: * https://github.com/pytorch/data/issues/200

Basically, it is because the input and output type are different (from `MapDataPipe` to `IterDataPipe`). `@functional_datapipe` currently doesn't handle that well.

torchdata/datapipes/map/util/converter.py
NivekT: One option is to materialize it during initialization and save that as an instance variable. Does that seem more appropriate?

VitalyFedyunin: Makes sense. Lets leave a comment on why it is purposely List.

torchdata/datapipes/iter/util/converter.py
```diff
@@ -68,6 +68,12 @@ def __getitem__(self, index):
             self._load_map()
         return self._map[index]  # type: ignore[index]
 
+    def __iter__(self):
+        if self._map is None:
+            self._load_map()
+        else:
+            yield from self._map.values()
+
```
NivekT, line 76: Adding this method makes sense to me since in some cases people may want to iterate over a `MapDataPipe` and since the source is an `IterDataPipe` it provides an easy order that we can follow.

ejguan: But, it invalidates our definition for `MapDataPipe` right?
Even without `__iter__`, users would still be able iterate over this DataPipe using `iter()`, which would implicitly use `[0-n)` as the indices and invoke `__getitem__`.

torchdata/datapipes/map/util/converter.py
```diff
@@ -0,0 +1,42 @@
...
+            raise TypeError(f"MapToIterConverter can only apply on MapDataPipe, but found {type(datapipe)}")
+        self.datapipe: MapDataPipe = datapipe
+        self.indices = indices if indices else range(len(datapipe))
+
+    def __iter__(self):
+        for idx in self.indices:
+            yield self.datapipe[idx]
+
+    def __len__(self):
+        return len(self.datapipe)
```
ejguan, line 39: It's better to use `len(self.indices)` here because the length can be different from `datapipe`.

ejguan (approved):
LGTM. Do we need to add this DataPipe to `torchdata.datapipes.map.rst`?

torchdata/datapipes/iter/util/converter.py
NivekT: It currently doesn't terminate correctly:
```python
from torchdata.datapipes.map import IterToMapConverter

dp = IterableWrapper([(i, i) for i in range(10)])
dp = IterToMapConverter(dp)
for i in iter(dp):
    print(i)
```
In the end, you will get `KeyError: 10`. 


In terms of definition, I think `__iter__` should not be required for `MapDataPipe`, but it should be fine to have `__iter__` if desired?

ejguan: I am sorry I get confused by the convertor. This is the `IterDataPipe` not `MapDataPipe`. Then, it's fine to have both `__iter__` and `__getitem__`.
I will comment on the PR for `IterToMapConverter` for the Error.

NivekT:
@nivekt has imported this pull request.  If you are a Facebook employee, you can view this diff [on Phabricator](https://www.internalfb.com/diff/D35979908).

PR #328, opened 2022-03-24
Adding InMemoryCacheHolderMapDataPipe

NivekT:
* #329
* __->__ #328
* #327
* #326
* #325

I think it will be useful to have a `MapDataPipe` for caching data, especially right before shuffling and etc. The use case should be common enough that we should support it rather than asking users to use the `iter` version.

This also have the added functionality for users to customize the wrapper around their data (e.g. user numpy array instead of list).

Differential Revision: [D35979905](https://our.internmc.facebook.com/intern/diff/D35979905)


torchdata/datapipes/map/util/cacheholder.py
```diff
@@ -0,0 +1,54 @@
...
+
+def list_cache(size: int) -> List:
+    cache = [None] * size
+    try:
+        from numpy import array
+
+        cache = array(cache)
+    except (ImportError, ModuleNotFoundError):
+        pass
+    return cache
```
NivekT, line 19: My intention here is to default to a `numpy.array` when available, and a `List` when it is not.

I thought about the memory leak issues related to DataLoader when `List` is used and I want to avoid that by defaulting the `numpy.array` instead. Any thoughts on how this should be handled here?

torchdata/datapipes/map/util/cacheholder.py
```diff
@@ -0,0 +1,54 @@
...
+        >>> list(cache_dp)
+        [0, 1, 2, 3, 4, 5, 6, 7, 8, 9]
+    """
+
+    def __init__(self, source_dp: MapDataPipe[T_co], cache_fn: Optional[Callable] = None) -> None:
+        self.source_dp: MapDataPipe[T_co] = source_dp
+        if cache_fn is None:
+            cache_fn = list_cache
+        self.cache = cache_fn(len(source_dp))
+        if not (hasattr(self.cache, "__getitem__") and hasattr(self.cache, "__setitem__")):
```
NivekT, line 44: The only requirement is for the initialized `cache` to have these methods. Does this seem reasonable enough?

NivekT:
@ejguan @VitalyFedyunin Since there were some questions about whether this is needed (or if users should use the `IterDataPipe` version), can you have a look at this before I add the docstring/tests?

torchdata/datapipes/map/util/cacheholder.py
```diff
@@ -0,0 +1,54 @@
...
+    Example:
+        >>> from torchdata.datapipes.map import SequenceWrapper
+        >>> source_dp = SequenceWrapper(range(10))
+        >>> cache_dp = source_dp.in_memory_cache()
+        >>> list(cache_dp)
+        [0, 1, 2, 3, 4, 5, 6, 7, 8, 9]
+    """
+
+    def __init__(self, source_dp: MapDataPipe[T_co], cache_fn: Optional[Callable] = None) -> None:
+        self.source_dp: MapDataPipe[T_co] = source_dp
```
ejguan, line 40: If the source_dp is a `MapDataPipe` and it holds the memory of the Dataset, why we need this DataPipe as all data has been stored in `source_dp`.

NivekT: 1. IIUC, since `MapDataPipe` is lazily computed, the operations are applied every time `__getitem__` is called.

Here is a silly example:
```python
def map_dp_quick_test(cached: bool):
    print(f"cached={cached}")
    dp = SequenceWrapper(range(100000))
    dp = dp.map(lambda x: x + 1)
    dp = dp.map(lambda x: x * 10)
    dp = dp.map(lambda x: x + 1)
    dp = dp.map(lambda x: x * 10)
    dp = dp.map(lambda x: x + 1)
    if cached:
        dp = dp.in_memory_cache()
    dps = [list(dp.map(lambda x: x + 1)) for _ in range(100)]

map_dp_quick_test(cached=True)
gc.collect()
map_dp_quick_test(cached=False)
```
Results:
```
[4.35055778s] map_dp_quick_test(cached=True)
[14.84188485s] map_dp_quick_test(cached=False)
```

It may be useful to cache these operations if `cache_dp` ended up being used by multiple DataPipes downstream. It is unlikely to be as useful as the `IterDataPipe` version though.

2. I think it will be helpful to provide this as a pattern to encourage using `numpy.array` (or similar wrapper) to avoid the "copy on write" memory issue that some users have experienced.

3. Lastly, I have seen some questions by users about caching between epoch while having multiple workers. I think using `multiprocessing.Array` here can achieve that goal (I am not sure if that should be the recommended pattern, but it is a possibility). Let me know what you think.

NivekT: > > 1. IIUC, since `MapDataPipe` is lazily computed, the operations are applied every time `__getitem__` is called.
> 
> It makes sense
> 
> For 2 and 3, we do want to support such behavior. But, I don't know this is the right approach. For 2, we are adding a dependency of numpy. For 3, the overhead of `mp.Array` is really high IIRC
> 
> I would love to use list as the default data structure but let user can provide different class for cache if they want to take care of COW problem

Do you think the soft dependency on `numpy` as implemented here is reasonable? I made that choice because I feel like users aren't usually aware of the COW problem until it is a very big one. Using `numpy.array` by default here somewhat mitigates that problem (or at least prompt user to take a look if we describe it well in the docstring).

I don't plan to use `mp.Array` by default here but it will be good to mention it and provide an example.

torchdata/datapipes/map/util/cacheholder.py
```diff
@@ -0,0 +1,46 @@
...
+
+    Example:
+        >>> from torchdata.datapipes.map import SequenceWrapper
+        >>> source_dp = SequenceWrapper(range(10))
+        >>> cache_dp = source_dp.in_memory_cache()
+        >>> list(cache_dp)
+        [0, 1, 2, 3, 4, 5, 6, 7, 8, 9]
+    """
+
+    def __init__(self, source_dp: MapDataPipe[T_co], cache: Optional[Callable] = None) -> None:
```
ejguan, line 32: Why do we want to expose `cache` to users? I believe a `dict` should be sufficient for the cache.

torchdata/datapipes/map/util/cacheholder.py
```diff
@@ -0,0 +1,46 @@
...
+        self.cache = {} if cache is None else cache
+        if not (hasattr(self.cache, "__getitem__") and hasattr(self.cache, "__setitem__")):
+            raise TypeError("`cache` must be an object with  both '__getitem__' and '__setitem__.")
+        self.index_cached: Set = set()
+
+    def __getitem__(self, index) -> T_co:
+        if index not in self.index_cached:
+            self.cache[index] = self.source_dp[index]  # type: ignore[index]
+            self.index_cached.add(index)
+        return self.cache[index]  # type: ignore[index]
```
ejguan, line 43: I don't think we need `self.index_cached`. All the information are stored in `self.cache`. You can do:
```py
if index not in self.cache:
    self.cache[index] = self.source_dp[index]
return self.cache[index]
```

torchdata/datapipes/map/util/cacheholder.py
```diff
@@ -0,0 +1,46 @@
...
+        >>> from torchdata.datapipes.map import SequenceWrapper
+        >>> source_dp = SequenceWrapper(range(10))
+        >>> cache_dp = source_dp.in_memory_cache()
+        >>> list(cache_dp)
+        [0, 1, 2, 3, 4, 5, 6, 7, 8, 9]
+    """
+
+    def __init__(self, source_dp: MapDataPipe[T_co], cache: Optional[Callable] = None) -> None:
+        self.source_dp: MapDataPipe[T_co] = source_dp
+        self.cache = {} if cache is None else cache
```
ejguan, line 34: You can add a type annotation here `Dict[Any, T_co]`

torchdata/datapipes/map/util/cacheholder.py
NivekT: `self.index_cached` is needed if users specify an array-like cache. If we remove that option, then yes we will not need that.

torchdata/datapipes/map/util/cacheholder.py
ejguan: If it's LRU cache, the element will be popped out the cache when it's full. Then, there is a bug that `index_cached` will still keep the index that has already removed from the cache.

I believe you want to use `numpy.array` or `mp.Array` for COW issue during multiprocessing. However, I don't think this is the problem in this case because `self.cache` is empty when we pass this DataPipe instance into multiprocessing. It's different from putting list of indices into `Dataset` causing the COW issue. This is legitimate to have increasing memory usage as each process will hold a separate `cache` object.

torchdata/datapipes/map/util/cacheholder.py
ejguan: With array-like cache, you are assuming the index is 0 based, which is not always true.

torchdata/datapipes/map/util/cacheholder.py
NivekT: If the COW issue isn't relevant here, then I will just use a `dict`. Thanks for the comment!

torchdata/datapipes/map/util/cacheholder.py
NivekT: I am switching to `Dict`-based cache

torchdata/datapipes/map/util/cacheholder.py
```diff
@@ -0,0 +1,48 @@
...
+    """
+
+    def __init__(self, source_dp: MapDataPipe[T_co]) -> None:
+        self.source_dp: MapDataPipe[T_co] = source_dp
+        self.cache: Dict[Any, T_co] = {}
+
+    def __getitem__(self, index) -> T_co:
+        if index not in self.cache:
+            self.cache[index] = self.source_dp[index]  # type: ignore[index]
+        return self.cache[index]  # type: ignore[index]
```
ejguan, line 45: I have a crazy idea that might be over-engineering here. Should we remove `self.source_dp` after all values are cached? We can save memory for users.
😃

ejguan (approved):
LGTM thank you.

torchdata/datapipes/map/util/cacheholder.py
NivekT: Haha interesting idea. I will hold off on that for now since I am unsure how that will interact with other features such as graph traversal and snapshotting.

For example, if the cache is filled and we delete `self.source_dp`, then we run `traverse`, then the graph will be incomplete? I am going to add a note here to state that it may be possible though.

NivekT:
@nivekt has imported this pull request.  If you are a Facebook employee, you can view this diff [on Phabricator](https://www.internalfb.com/diff/D35979905).

PR #333, opened 2022-03-25
Removing old nightly build when uploading

ejguan:
Add extra logic to list all previous nightly release from conda and remove the corresponding conda packages if there is new package built based on the same Python version and platform


.github/workflows/nightly_release.yml
```diff
@@ -3,7 +3,7 @@ name: Push Nightly Release
 on:
   workflow_dispatch:
   schedule:
-    - cron: 00 12 * * *
+    - cron: 00 15 * * *
```
ejguan, line 6: Move it to UTC 3:00 PM because PyTorch nightly are normally compiled and uploaded successfully before this time.

NivekT (approved):

PR #344, opened 2022-04-01
Fix codegen files not included in wheel

### LINKED ISSUE #342: [Codegen] pyi files are not included in the wheel
### 🐛 Describe the bug

We generate pyi file after setup function. https://github.com/pytorch/data/blob/71a16a3b63966af0f60af74d3044b3c963cfc5fb/setup.py#L89

This is not useful if users doing `python setup.py install`. And, for the pre-compiled library using `python setup.py bdist_wheel`, the pyi file would never be packaged into the wheel/conda.

### Versions

main branch

ejguan:
Fixes #342

See issue

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -2,11 +2,12 @@
 # Copyright (c) Facebook, Inc. and its affiliates.
 import os
 import subprocess
+import sys
 
 from pathlib import Path
 
 from setuptools import find_packages, setup
-from torchdata.datapipes.gen_pyi import gen_pyi
+from tools.gen_pyi import gen_pyi
 
 
 ROOT_DIR = Path(__file__).parent.resolve()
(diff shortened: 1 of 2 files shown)


setup.py
```diff
@@ -58,6 +59,11 @@ def _export_version(version, sha):
 
     print("-- Building version " + VERSION)
 
+    if sys.argv[1] != "clean":
+        gen_pyi()
+        # TODO: Fix #343
+        os.chdir(str(ROOT_DIR))
```
ejguan, line 65: Run codegen before `setup`

setup.py
```diff
@@ -82,8 +88,12 @@ def _export_version(version, sha):
             "Programming Language :: Python :: Implementation :: CPython",
             "Topic :: Scientific/Engineering :: Artificial Intelligence",
         ],
+        package_data={
+            "torchdata": [
+                "datapipes/iter/*.pyi",
+            ],
+        },
```
ejguan, line 95: Include pyi files into package

ejguan:
Closing this PR because I have integrated this path into the PR from S3

PR #345, opened 2022-04-03
Added CI tests for torcharrow

### LINKED ISSUE #301: Add TorchArrow Nightly CI Test
### 🚀 The feature

TorchArrow nightly build is now [available for Linux](https://download.pytorch.org/whl/nightly/cpu/torch_nightly.html) (other versions will be next).

We should add TorchArrow nightly CI tests for these [TorchArrow dataframe related unit tests](https://github.com/pytorch/data/blob/main/test/test_dataframe.py).

### Motivation, pitch

This will ensure that our usages remain compatible with TA's APIs.

### Additional context
This is a good first issue for people who want to understand how our CI works. Other [domain CI tests](https://github.com/pytorch/data/blob/main/.github/workflows/domain_ci.yml) (for Vision, Text) can serve as examples on how to set this up.

lezwon:
Fixes #301

### Changes

-Added torcharrow config to domain_ci.yml
-

### DIFF
--- a/.github/workflows/domain_ci.yml
+++ b/.github/workflows/domain_ci.yml
@@ -145,3 +145,47 @@ jobs:
         run: pytest --no-header -v test/test_audio_examples.py
         env:
           PYTORCH_TEST_WITH_SLOW: 1
+
+  torcharrow:
+    if: ${{ github.repository_owner == 'pytorch' }}
+    runs-on: ${{ matrix.os }}
+    strategy:
+      fail-fast: false
+      matrix:
+        os:
+          # - macos-latest
+          - ubuntu-latest
+          # - windows-latest
+        python-version:
+          - 3.7
+          - 3.8
+          - 3.9
+    steps:
+      - name: Setup Python ${{ matrix.python-version }}
+        uses: actions/setup-python@v2
+        with:
+          python-version: ${{ matrix.python-version }}
+
+      - name: Install torch and torcharrow from nightlies
+        run: pip install --pre torch torcharrow -f https://download.pytorch.org/whl/nightly/cpu/torch_nightly.html
+
+      - name: Check out torchdata repository
+        uses: actions/checkout@v2
+
+      - name: Install torchdata
+        run: |
+          pip install -r requirements.txt
+          python setup.py install
+
+      - name: Install test requirements
+        run: pip install dill expecttest numpy
+
+      - name: Run torcharrow example datasets tests
+        if: ${{ ! contains(github.event.pull_request.labels.*.name, 'ciflow/slow') }}
+        run: pytest --no-header -v test/test_dataframe.py
+
+      - name: Run torcharrow example datasets (including slow tests)
+        if: ${{ contains(github.event.pull_request.labels.*.name, 'ciflow/slow') }}
+        run: pytest --no-header -v test/test_dataframe.py
+        env:
+          PYTORCH_TEST_WITH_SLOW: 1
--- a/.github/workflows/domain_ci.yml
+++ b/.github/workflows/domain_ci.yml
@@ -178,7 +178,7 @@ jobs:
           python setup.py install
 
       - name: Install test requirements
-        run: pip install dill expecttest numpy
+        run: pip install dill expecttest numpy pytest
 
       - name: Run torcharrow example datasets tests
         if: ${{ ! contains(github.event.pull_request.labels.*.name, 'ciflow/slow') }}


NivekT (approved):
LGTM! Thanks for adding this!

.github/workflows/domain_ci.yml
```diff
@@ -145,3 +145,47 @@ jobs:
...
+          - 3.8
+          - 3.9
+    steps:
+      - name: Setup Python ${{ matrix.python-version }}
+        uses: actions/setup-python@v2
+        with:
+          python-version: ${{ matrix.python-version }}
+
+      - name: Install torch and torcharrow from nightlies
+        run: pip install --pre torch torcharrow -f https://download.pytorch.org/whl/nightly/cpu/torch_nightly.html
```
wenleix, line 170: FWIW, when GitHub Actions provides `E5-2673 (v3 or v4) ` as CPU, today TorchArrow binary will encountered "Illegal Instruction" issue due to Velox requires some instructions not supported on that CPU :( 

https://github.com/facebookresearch/torcharrow/pull/92#issuecomment-982850395

lezwon: Hey @wenleix, I asked this question on the GIthub community forum. I don't think there's a way to do this yet. 
https://github.community/t/exclude-cpu-type-in-github-actions/243459

ejguan: We can add a step in this workflow to figure out the cpu model. If the model is among the exclusion list, we can skip all the following steps.

I only know we can use `lscpu` on linux. Not sure how to achieve it on windows machine though.

NivekT: @lezwon Can you make a change based on ejguan's suggestion above? Thanks!

lezwon: @NivekT sure thing.

lezwon: @NivekT @ejguan I'm not very sure how to go about adding a condition for the same in GitHub action. I have made some changes in the latest commit. Do let me know if its valid. :)

ejguan: cc: @wenleix It seems there are more CPU instruction sets not supported by Velox based on https://github.com/pytorch/data/runs/5989057961?check_suite_focus=true

.github/workflows/domain_ci.yml
```diff
@@ -145,3 +145,60 @@ jobs:
...
+          - 3.8
+          - 3.9
+    steps:
+      - name: Check if the CPU is supported
+        run: |
+          if [ "$RUNNER_OS" == "Linux" ]; then
+            cpu=$(lscpu | grep E5-2673)
+          elif [ "$RUNNER_OS" == "macOS" ]; then
+            cpu=$(sysctl -a | grep E5-2673)
+          fi
```
ejguan, line 170: Could we add a few `echo` for future debugging?
Add
```sh
else
    echo "$RUNNER_OS is not supported"
    exit 1
fi
```

.github/workflows/domain_ci.yml
```diff
@@ -145,3 +145,60 @@ jobs:
...
+
+          if [ $cpu ]; then
+            echo "E5-2673 is not supported"
+            exit 1
+          fi
```
ejguan, line 175:
```sh
else
   echo "The CPU on $RUNNER_OS  is $cpu"
fi
```

ejguan (commented):
Overall, looks good.

.github/workflows/domain_ci.yml
wenleix: @ejguan I guess the "check if CPU is supported" doesn't really kick in?

```
Run if [ "$RUNNER_OS" == "Linux" ]; then
/home/runner/work/_temp/7b748253-1d41-4cd6-8083-de5[13](https://github.com/pytorch/data/runs/5989057961?check_suite_focus=true#step:2:13)e8058fd.sh: line 7: [: too many arguments
```

wenleix: Also maybe print cpuinfo so we can debug?

lezwon:
@ejguan sorry for the delay.Been a bit occupied for the past weeks. I have made the changes as you have requested.

.github/workflows/domain_ci.yml
```diff
@@ -145,3 +145,65 @@ jobs:
...
+          elif [ "$RUNNER_OS" == "macOS" ]; then
+            cpu=$(sysctl -a | grep E5-2673)
+          else
+              echo "$RUNNER_OS is not supported"
+              exit 1
+          fi
+
+          if [ $cpu ]; then
+            echo "E5-2673 is not supported"
+            exit 1
```
ejguan, line 177: Not sure why the job doesn't exit
image

Could you first use double bracket since we are using `bash` here? And, maybe try to do `if [[ -z $cpu ]]; then` to see if this condition is passed and the job exits normally.

.github/workflows/domain_ci.yml
```diff
@@ -145,3 +145,65 @@ jobs:
...
+          # - macos-latest
+          - ubuntu-latest
+          # - windows-latest
+        python-version:
+          - 3.7
+          - 3.8
+          - 3.9
+    steps:
+      - name: Check if the CPU is supported
+        run: |
```
ejguan, line 165: Let's add `set -x` here to see which line raises the `exit 1`

.github/workflows/domain_ci.yml
NivekT: Looks like it will not be necessary to check for certain CPUs anymore after this PR has landed:
https://github.com/facebookresearch/torcharrow/pull/315#issuecomment-1119151010

We can remove those related logic in this PR. Thanks @wenleix !

PR #359, opened 2022-04-12
Implement map_batches to align with TorchArrow API

ejguan:
Per title


torchdata/datapipes/iter/transform/callable.py
```diff
@@ -0,0 +1,129 @@
...
+
+        if isinstance(self.input_col, (list, tuple)):
+            args = [[data[idx] for idx in self.input_col] for data in batch]
+        else:
+            args = [data[self.input_col] for data in batch]
+        return self.fn(args)
+
+    def __iter__(self) -> Iterator[T_co]:
+        batch: List = []
+        for d in self.datapipe:
```
NivekT, line 81: nit: can this be done by `.batch(batch_size).map(...)`?

test/test_datapipe.py
```diff
@@ -649,6 +649,46 @@ def _custom_len_fn(token):
         with self.assertRaises(TypeError):
             len(batch_dp)
 
+    def test_map_batches_iterdatapipe(self):
```
NivekT, line 652: This looks good, can you add this DataPipe to `test_serialization` as well?

NivekT (approved):
Overall, looks good! Thanks for adding this!

torchdata/datapipes/iter/transform/callable.py
ejguan: hhhh, totally make sense to me. 😃

ejguan: After testing, `.batch(batch_size).map()` is not equal to `.map_batches()` when `input_col` is specified because `map` would treat input as a single data structure rather than a batch. Then, the `input_col` would be applied at batch level rather than data level.
Let's say our input is `[(0,1), (2,3), (3,4), (5,6), (7,8), (9,10)]` and batch size as 3. With `input_col` as 1, the inputs sent into `fn` are different between the two implementations.
- `.batch().map()`: Inputs are `(2, 3)` and `(7, 8)`
- `.map_batches()`: inputs are `(1, 3, 4)` and `(6, 8, 10)`

NivekT: I see. This is because we have `data[idx]` rather than `batch[idx]` within `_apply_fn`.

PR #363, opened 2022-04-18
Add input_col to flatmap 

ejguan:
This change is required to align DataPipe and DataFrame API


wenleix (approved):
LGTM

NivekT (approved):
LGTM! I think `map` has `output_col`, is the plan to generally only add `input_col` to align with other APIs?

ejguan:
> LGTM! I think `map` has `output_col`, is the plan to generally only add `input_col` to align with other APIs?

`flatmap` is not going to support `output_col` unless we receive more feature request because adding a `output_col` needs broadcasting the columns from the original input and it's not safe to do the broadcasting when `DataPipe` takes arbitrary python objects.

PR #367, opened 2022-04-21
add support for WebDataset-style archives

tmbdev:
WebDataset is a simple convention for representing datasets in sequentially read archives (like tar archives). It is supported natively by the [webdataset/webdataset ](https://github.com/webdataset/webdataset/) library. Torchdata contains most of the functionality of the webdataset/webdataset library already except for the grouping of samples. This PR adds a filter that groups files inside tar files into samples according to WebDataset conventions.

### Changes

- adds WebDatasetIterDataPipe
- adds test cases
- adds test case file generator to _create_fake_data.py


torchdata/datapipes/iter/util/webdataset.py
```diff
@@ -0,0 +1,66 @@
...
+    prefix, suffix = re.search(r"^(.*?)(\.[^/]*)$", p).groups()
+    return prefix, suffix
+
+
+@functional_datapipe("webdataset")
+class WebDatasetIterDataPipe(IterDataPipe[Dict]):
+    r"""
+    Iterable DataPipe that accepts stream of (path, data) tuples and aggregates them into
+    a single dictionary based on the basenames of the paths (WebDataset file convention).
+
```
VitalyFedyunin, line 21: Please mention that it is aggregates only sequential data, meaning if same key is going to be separated by different one, it will yield 3 data pieces.

torchdata/datapipes/iter/util/webdataset.py
```diff
@@ -0,0 +1,66 @@
...
+            if prefix != current:
+                if current != "":
+                    yield sample
+                sample = {}
+                current = prefix
+            sample[suffix] = data
+        if sample != {}:
+            yield sample
+
+    def __len__(self) -> int:
```
msaroufim, line 65: sorry a n00b question, what does length mean for a streaming dataset?

torchdata/datapipes/iter/util/webdataset.py
```diff
@@ -0,0 +1,66 @@
...
+    Returns:
+        a DataPipe yielding a stream of dictionaries
+    Examples:
+        >>> from torchdata.datapipes.iter import FileLister, FileOpener
+        >>>
+        >>> def decode(item):
+        >>>     key, value = item
+        >>>     if key.endswith(".txt"):
+        >>>         return key, value.read().decode("utf-8")
+        >>>     if key.endswith(".bin"):
```
msaroufim, line 35: It seems like the PR isn't only limited to text data though right? Would it make sense to have a more generic `decode` function available outside of the test to deal with images

msaroufim (changes requested):

torchdata/datapipes/iter/util/webdataset.py
tmbdev: In most cases, it seems to be a user-settable value indicating the actual size of the dataset. In any case, this implementation just follows the torchdata conventions here.

torchdata/datapipes/iter/util/webdataset.py
tmbdev: I will submit another PR for a class that helps with decoding soon. However, such a class is just there for convenience, since decoding with `map` is easy enough; you don't need it to use WebDataset.

Since it isn't essential and since a decoding class involves discussions about how to handle external dependencies and extensibility, it probably will require more discussions, that's why I think it's best to keep it to a separate PR.

test/_fakedata/_create_fake_data.py
```diff
@@ -35,6 +36,22 @@ def create_files(folder, suffix, data, encoding=False):
         archive.add(folder)
 
 
+def add_data_to_tar(archive, name, value):
```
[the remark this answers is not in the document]
tmbdev: I'm not sure what you mean. The tar file for the test case, `wds.tar` is not empty. I added it to `test/_fakedata` and the repository because there are already other tar files used for test cases there as well. Of course, it could also be generated on the fly.

test/test_datapipe.py
```diff
@@ -871,6 +872,46 @@ def test_itertomap_mapdatapipe(self):
             self.assertEqual(len(wa), 1)
             self.assertRegex(str(wa[0].message), r"Found duplicate key")
 
+    def test_webdataset(self) -> None:
```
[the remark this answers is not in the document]
tmbdev: For the example, I wanted to show image loading and (possibly) remote loading, so I'm planning a slightly more elaborate example. But before writing that, I would like to make sure that the API is accepted.

torchdata/datapipes/iter/util/webdataset.py
tmbdev: I have updated the comment and added the commit to this PR.

test/test_datapipe.py
```diff
@@ -871,6 +872,46 @@ def test_itertomap_mapdatapipe(self):
...
+        def decode(item):
+            key, value = item
+            if key.endswith(".txt"):
+                return key, value.read().decode("utf-8")
+            if key.endswith(".bin"):
+                return key, value.read().decode("utf-8")
+
+        datapipe1 = FileLister("test/_fakedata", "wds*.tar")
+        datapipe2 = FileOpener(datapipe1, mode="b")
+        dataset = datapipe2.load_from_tar().map(decode).webdataset()
```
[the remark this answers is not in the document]
tmbdev: Please see the comment above.

torchdata/datapipes/iter/util/webdataset.py
```diff
@@ -0,0 +1,69 @@
...
+        sample = {}
+        current = ""
+        for path, data in self.source_datapipe:
+            assert isinstance(path, str), path
+            prefix, suffix = pathsplit(path)
+            if suffix == "":
+                continue
+            if prefix != current:
+                if current != "":
+                    yield sample
```
NivekT, line 61: nit: Not familiar with the usage of `WebDataset`, will the users be interested in knowing what the `prefix` is for the sample that is being yielded? 

If so, maybe we should `yield prefix, sample`. If not, ignore this comment. I suppose users may also be able to get that info from the `data` within `sample.

torchdata/datapipes/iter/util/webdataset.py
```diff
@@ -0,0 +1,69 @@
...
+        self.source_datapipe: IterDataPipe[List[Union[Dict, List]]] = source_datapipe
+
+    def __iter__(self) -> Iterator[Dict]:
+        sample = {}
+        current = ""
+        for path, data in self.source_datapipe:
+            assert isinstance(path, str), path
+            prefix, suffix = pathsplit(path)
+            if suffix == "":
+                continue
```
NivekT, line 58: nit: How often will people run into this? And if they do, is it because `pathsplit` made a mistake or the users named the files incorrectly? I am wondering if we should raise an exception or warning instead.

torchdata/datapipes/iter/util/webdataset.py
```diff
@@ -0,0 +1,69 @@
...
+    return prefix, suffix
+
+
+@functional_datapipe("webdataset")
+class WebDatasetIterDataPipe(IterDataPipe[Dict]):
+    r"""
+    Iterable DataPipe that accepts stream of (path, data) tuples (usually,
+    representing the pathnames and files of a tar archive) and aggregates
+    consecutive items with the same basename into a single dictionary, using the
+    extensions as keys (WebDataset file convention). Any text after the first
```
NivekT, line 22:
```suggestion
    extensions as keys based on the WebDataset file convention (functional name: ``webdataset``). Any text after the first
```

In our [documentation](https://pytorch.org/data/beta/torchdata.datapipes.iter.html), we want to show the functional name of each DataPipe in the summary table. In order for it to render properly, we typically add `(function name: ...)` at the end of the first sentence in the docstring, hence this suggestion.

NivekT (commented):
Thanks for adding this! This looks good and I just have some small, in-line questions and suggestions.

Also, can you add this DataPipe to:
1. [`test_serialization.py`](https://github.com/pytorch/data/blob/6da391e3d342248a6005c74c61499a3209bda010/test/test_serialization.py#L260) above the specific line with sample inputs
2. Our [documentation RST file](https://github.com/pytorch/data/blob/main/docs/source/torchdata.datapipes.iter.rst) under Archive DataPipes

torchdata/datapipes/iter/util/webdataset.py
tmbdev: The prefix is already returned as part of the `sample` in the `__key__` field; there is no need to return it separately.

torchdata/datapipes/iter/util/webdataset.py
tmbdev: Files without extensions can be used for metadata; it's fine for readers to ignore them. 

Since users need an extension to refer to a file anyway, there really is no risk of having file names without extensions accidentally; if they generated such files, they would immediately realize that there is no way to refer to the file inside the sample.

torchdata/datapipes/iter/util/webdataset.py
tmbdev: OK, I'll have a look.

torchdata/datapipes/iter/util/webdataset.py
NivekT: Makes sense to me. Do you mind adding a short inline comment there for future reference?

torchdata/datapipes/iter/util/webdataset.py
tmbdev: The current prefix is now returned in `sample["__key__"]`, consistent with the WebDataset library.

torchdata/datapipes/iter/util/webdataset.py
tmbdev: Done.

torchdata/datapipes/iter/util/webdataset.py
tmbdev: Done.

tmbdev:
I tried to address all the comments. The test cases are all working and it passes pre-commit. Please let me know if you need any additional updates.

(Plans: I will separately submit a file decoding filter, and once that is done, a longer example using remote access. After that, I'm planning on submitting an AIStore client (AIStore is an S3-compatible object store with additional functionality for deep learning, such as server-side shuffling and off-loading of data transformation/augmentation to the storage system)).

msaroufim:
@tmbdev the AIStore integration sounds super compelling, very excited

NivekT (approved):
Hi @tmbdev, thanks for working with us on this PR. Overall, it looks good! We just need to two more small fixes before I can merge this:
1. [Linting](https://github.com/pytorch/data/runs/6177509075?check_suite_focus=true) - just the two related to `webdataset.py`, ignore the third
2. [Windows CI failures](https://github.com/pytorch/data/runs/6177511087?check_suite_focus=true) - it may be due to how the paths are specified.

tmbdev:
> Hi @tmbdev, thanks for working with us on this PR. Overall, it looks good! We just need to two more small fixes before I can merge this:
> 
>     1. [Linting](https://github.com/pytorch/data/runs/6177509075?check_suite_focus=true) - just the two related to `webdataset.py`, ignore the third

Fixed.

>     2. [Windows CI failures](https://github.com/pytorch/data/runs/6177511087?check_suite_focus=true) - it may be due to how the paths are specified.

OK, I think this is fixed now. There is a lengthy comment explaining how this works on Windows.

tmbdev:
Looks like all the WebDataset-related tests now pass on all platforms.

NivekT:
> Looks like all the WebDataset-related tests now pass on all platforms.

Actually one more change is needed since our internal test is failing. Ideally we should generate the test data files during setup of the test rather than storing them in `_fakedata` whenever possible. Let me quickly put something together.

tmbdev:
I have moved the test file creation to the test case

NivekT:
Hi @tmbdev, thanks for your patience and sorry for the back and forth.

I have made a [patch for some small changes](https://github.com/pytorch/data/pull/378/files) (of two commits) that are necessary. I have tested those changes to make sure they pass out internal tests.
1. `test_local_io.py` has temporary directory setup for testing where files can be written. Moving the news test to that file will make things simpler and consistent.
2. With the test files being generated during the test, all changes within `test/_fakedata` can be reverted.
3. Small addition of copyright header to `webdataset.py`.

You should be able to cherry-pick that [patch](https://github.com/pytorch/data/pull/378/files) (two commits). If you run into issues patching/reverting the tar files within `_fakedata`, this command might help:
`git restore --source=b6ade8f097bc9ac08460cd403034a35daff09cfa --staged --worktree -- test/_fakedata`

tmbdev:
Thanks. I have incorporated the changes.

test/test_local_io.py
```diff
@@ -721,6 +723,63 @@ def test_rar_archive_loader(self):
...
+            [
+                {".jpg": b"1", ".cls": b"2", "__key__": "/path/to/file1"},
+                {".jpg": b"3", ".cls": b"4", "__key__": "/path/to/file2"},
+            ],
+            list(web_dataset),
+        )
+
+    def test_webdataset2(self) -> None:
+        # Setup
+        self._create_wds_tar(os.path.join(self.temp_dir.name, "wds.tar"))
```
VirrageS, line 766: Maybe we should define `record_count = 10` and then pass it here?
```suggestion
        record_count = 10
        self._create_wds_tar(os.path.join(self.temp_dir.name, "wds.tar"), record_count)
```

It's not obvious that at the end of the test you are expecting `10` items.

tmbdev: OK, added. I also am using it in the assertion later.

test/test_datapipe.py
```diff
@@ -6,13 +6,16 @@
 
 import io
 import itertools
+import os
```
NivekT, line 9: @tmbdev sorry can you revert the changes to this file too?

NivekT: thank you so much!!!

tmbdev: Done. (I'm surprised the linter doesn't check for unused imports.)

PR #386, opened 2022-05-05
Updating test code to follow single iterator constraint

NivekT:
Stack from [ghstack](https://github.com/ezyang/ghstack):
* #388
* __->__ #386

This should be landed after https://github.com/pytorch/pytorch/pull/75995

Note that the CI is expected to fail until that PR goes into nightly

Differential Revision: [D36182556](https://our.internmc.facebook.com/intern/diff/D36182556)

PR #388, opened 2022-05-06
Moving DataPipe buffers from __iter__ to instance (self)

NivekT:
* __->__ #388
* #386

This PR impacts `ParagraphAggregator` and `IterKeyZipper` as they previously have their buffer with the iterator rather than the instance.

This should be landed along with https://github.com/pytorch/pytorch/pull/76999 and https://github.com/pytorch/pytorch/pull/77775 

Note that the CI is expected to fail until that PR goes into nightly

Differential Revision: [D36519522](https://our.internmc.facebook.com/intern/diff/D36519522)


torchdata/datapipes/iter/util/combining.py
```diff
@@ -92,26 +92,60 @@ def __iter__(self) -> Iterator:
...
-            res = self.merge_fn(data, buffer.pop(key)) if self.merge_fn else (data, buffer.pop(key))
+                    self.buffer.popitem(last=False)
+                self.buffer[ref_key] = ref_data
+            res = self.merge_fn(data, self.buffer.pop(key)) if self.merge_fn else (data, self.buffer.pop(key))
             if self.keep_key:
                 yield key, res
             else:
                 yield res
+        if self.buffer:
+            self.buffer.clear()
```
NivekT, line 112: Interesting enough the domain CI test for Vision started to fail for a previous commit of this PR.

Perhaps we need clear the buffer when the iterator is exhausted? I am adding this line to see if the issue is resolved. Let me know if you have any other thoughts.

NivekT: Just tried and clearing the buffer at the end of `__iter__` doesn't help.

@pmeier Do you think this is related to `pytest` or something that is introduced in this PR?

NivekT:
@nivekt has imported this pull request.  If you are a Facebook employee, you can view this diff [on Phabricator](https://www.internalfb.com/diff/D36519522).

torchdata/datapipes/iter/util/paragraphaggregator.py
```diff
@@ -46,22 +46,37 @@ def __init__(self, source_datapipe: IterDataPipe[Tuple[str, T_co]], joiner: Call
...
+        self.buffer = []
+
+    def __getstate__(self):
+        if IterDataPipe.getstate_hook is not None:
+            return IterDataPipe.getstate_hook(self)
+        state = (self.source_datapipe, self.joiner)
+        return state
+
+    def __setstate__(self, state):
+        (self.source_datapipe, self.joiner) = state
```
ejguan, line 79: To follow other DataPipe's implementation, should we add `self.buffer = []` here?

ejguan (approved):
LGTM
We can still discuss if we want to serialize/deserialize buffer in `getstate` and `setstate` later in your snapshotting design.

torchdata/datapipes/iter/util/combining.py
NivekT: For reference, here is the thread where we discussed a similar issue:
https://github.com/pytorch/vision/pull/5801

NivekT: I am going to ignore the warning and land this based on the previous investigation.

PR #395, opened 2022-05-10
Added `file_lock` in `IoPathSaverIterDataPipe`

### LINKED ISSUE #144: Multiprocessing with any DataPipe writing to local file
### 🐛 Describe the bug

We need to take extra care all DataPipe that would write to file system when DataLoader2 triggered multiprocessing. If the file name on the local file system is same across multiple processes, it would be a racing condition.
This is found when TorchText team is using `on_disk_cache` to cache file.
DataLoader needs to know such DataPipe must be sharded with multiprocessing or enforce it into single process.

As a workaround, users have to download the file to local file system to prevent writing within DataPipe.

### Versions

main branch

msaroufim:
Untested, made this PR to make communication easier

Partially fixes #144 

The main change is to lock a file when multiple processes are writing to the same file

The fix was upstreamed from an internal fix @hudeven made

```python
with iopath.file_lock(filepath):
    if not os.path.exists(filepath):
        with self.pathmgr.open(filepath, self.mode) as f:
            f.write(data)
```

And the test was extended from a gist @ejguan sent me

It does seem to be working as intended right now, @parmeet please let me know if this works for you

### Logs with fix
```
Ran 17 tests in 3.982s

OK (skipped=1)
```

## Logs without fix

```
  File "/home/ubuntu/data/test/test_local_io.py", line 638, in test_io_path_saver_iterdatapipe
    self.assertEqual(expected_paths, res_file_paths)
AssertionError: Lists differ: ['/tmp/tmpbgt32c3h/1.txt', '/tmp/tmpbgt32c3h/2.txt', '/tmp/tmpbgt32c3h/3.txt'] != ['/tmp/tmpbgt32c3h/3.txt']

First differing element 0:
'/tmp/tmpbgt32c3h/1.txt'
'/tmp/tmpbgt32c3h/3.txt'

First list contains 2 additional elements.
First extra element 1:
'/tmp/tmpbgt32c3h/2.txt'

- ['/tmp/tmpbgt32c3h/1.txt', '/tmp/tmpbgt32c3h/2.txt', '/tmp/tmpbgt32c3h/3.txt']
+ ['/tmp/tmpbgt32c3h/3.txt']


```

### DIFF
--- a/test/test_graph.py
+++ b/test/test_graph.py
@@ -51,14 +51,26 @@ class TempReadingService(ReadingServiceInterface):
             dp.started = False
 
 
+def _x_and_x_plus_5(x):
+    return [x, x + 5]
+
+
+def _x_mod_2(x):
+    return x % 2
+
+
+def _x_mult_2(x):
+    return x * 2
+
+
 class TestGraph(expecttest.TestCase):
     def _get_datapipes(self) -> Tuple[IterDataPipe, IterDataPipe, IterDataPipe]:
         src_dp = IterableWrapper(range(20))
-        m1 = src_dp.map(lambda x: [x, x + 5])
+        m1 = src_dp.map(_x_and_x_plus_5)
         ub = m1.unbatch()
-        c1, c2 = ub.demux(2, lambda x: x % 2)
+        c1, c2 = ub.demux(2, _x_mod_2)
         dm = c1.main_datapipe
-        m2 = c1.map(lambda x: x * 2)
+        m2 = c1.map(_x_mult_2)
         dp = m2.zip(c2)
 
         return traverse(dp, only_datapipe=True), (src_dp, m1, ub, dm, c1, c2, m2, dp)


torchdata/datapipes/iter/load/iopath.py
```diff
@@ -183,8 +184,11 @@ def __init__(
     def __iter__(self) -> Iterator[str]:
         for meta, data in self.source_datapipe:
             filepath = meta if self.filepath_fn is None else self.filepath_fn(meta)
-            with self.pathmgr.open(filepath, self.mode) as f:
-                f.write(data)
+
+            if not os.path.exists(filepath):
+                with self.pathmgr.file_lock(filepath):
```
NivekT, line 189: Two questions:
1. Not familiar with this `if not os.path.exists(filepath)`. Does this means it only tries to lock if the path doesn't exist?
2. What is the behavior here if the file is already open elsewhere (perhaps there is another `file_lock`)? Does it perform a non-blocking wait?

msaroufim: This was in reference to https://www.internalfb.com/diff/D35459528?dst_version_fbid=2351551101810096&transaction_fbid=471804224682410

Ask was to skip writing to file if it already exists so for now I brought that check down to the write

Moved to PR to draft still scoping out the problem and solution, just wanted a way to communicate with you and @ejguan

ejguan: > Not familiar with this if not os.path.exists(filepath). Does this means it only tries to lock if the path doesn't exist?

I believe `filelock` will create a lock file as a mutex across processes. It's not using `os.path.exists(filepath)` to lock.

The implementation is a little bit weird though. I think we should do:
```py
            with self.pathmgr.file_lock(filepath):
                if not os.path.exists(filepath):
                    with self.pathmgr.open(filepath, self.mode) as f:
                        f.write(data)
```

parmeet:
> do you have some existing test/script I can leverage? Just wanna make sure your usecase is solved

Unfortunately, no. we observed this issue when running DDP while training doc classification workflow using TorchRecipes internally. cc: @hudeven

torchdata/datapipes/iter/load/iopath.py
```diff
@@ -183,8 +183,12 @@ def __init__(
     def __iter__(self) -> Iterator[str]:
         for meta, data in self.source_datapipe:
             filepath = meta if self.filepath_fn is None else self.filepath_fn(meta)
-            with self.pathmgr.open(filepath, self.mode) as f:
-                f.write(data)
+
+            with iopath.file_lock(filepath):
```
parmeet, line 187: This introduces dependency on iopath, whereas iopath itself depends on portalocker for file locks (https://github.com/facebookresearch/iopath/blob/4481d29d6ae873a2413de5b1aba41c0b0a723f75/iopath/common/file_io.py#L98).may be we should instead use portalocker directly?

msaroufim: That's an easy change, I saw that `iopath` was already used in the file which is why I picked it. Curious why you think it'd be better to use portalocker? Just about reducing a layer of indirection?

parmeet: Yes, reducing a layer of indirection is one reason. The other being is I think we no longer want to have file_locks as optional dependency since we are adding it to core datapipe. iopath as of now is optional and is needed only when users work with iopath related datapipes.

VitalyFedyunin:
It does not fixes #144 as on_disk_cache using https://github.com/pytorch/data/blob/2623265ed3cfcbb039d8e0c28c183d2cdca1a4b6/torchdata/datapipes/iter/util/saver.py#L18 to save files. 

Also you probably need to cross lock writing with https://github.com/pytorch/data/blob/c1d89fe9a1b06e610f32f823359771557b1ca12a/torchdata/datapipes/iter/util/cacheholder.py#L185

@ejguan please correct me if I'm wrong.

VitalyFedyunin:
Side comment: I'm ok to have `portalocker` as hard dependency.

ejguan:
> It does not fixes #144 as on_disk_cache using
> 
> https://github.com/pytorch/data/blob/2623265ed3cfcbb039d8e0c28c183d2cdca1a4b6/torchdata/datapipes/iter/util/saver.py#L18
> 
> to save files.
> Also you probably need to cross lock writing with
> 
> https://github.com/pytorch/data/blob/c1d89fe9a1b06e610f32f823359771557b1ca12a/torchdata/datapipes/iter/util/cacheholder.py#L185
> 
> @ejguan please correct me if I'm wrong.

Completely correct. We are using this iopath file to test `portalocker`. We can add another PR to fix `Saver` and `CacheHolder`

test/test_local_io.py
```diff
@@ -655,6 +655,42 @@ def filepath_fn(name: str) -> str:
...
+    @skipIfNoIoPath
+    def test_io_path_saver_file_lock(self):
+        def filepath_fn(name: str) -> str:
+            return os.path.join(self.temp_dir.name, os.path.basename(name))
+
+        # Same filename with different name
+        name_to_data = {"1.txt": b"DATA1", "1.txt": b"DATA2", "2.txt": b"DATA3", "2.txt": b"DATA4"}  # noqa: F601
+
+        # Add sharding_filter to shard data into 2
+        source_dp = IterableWrapper(name_to_data.items()).sharding_filter()
```
ejguan, line 667: Based on the failure on workflow, could you please fix:
```suggestion
        source_dp = IterableWrapper(list(name_to_data.items())).sharding_filter()
```

ejguan (approved):
LGTM with one comment on test.

test/test_local_io.py
```diff
@@ -655,6 +655,42 @@ def filepath_fn(name: str) -> str:
         # __len__ Test: returns the length of source DataPipe
         self.assertEqual(3, len(saver_dp))
 
+    @skipIfNoIoPath
+    def test_io_path_saver_file_lock(self):
+        def filepath_fn(name: str) -> str:
+            return os.path.join(self.temp_dir.name, os.path.basename(name))
```
ejguan, line 661: Could you please also move this function out of local scope? Python's multiprocessing doesn't support well with local functions.

test/test_local_io.py
```diff
@@ -655,6 +655,42 @@ def filepath_fn(name: str) -> str:
...
+        # Use appending as the mode
+        saver_dp = source_dp.save_by_iopath(filepath_fn=filepath_fn, mode="ab")
+
+        import torch.utils.data.graph_settings
+
+        def init_fn(worker_id):
+            info = torch.utils.data.get_worker_info()
+            num_workers = info.num_workers
+            datapipe = info.dataset
+            torch.utils.data.graph_settings.apply_sharding(datapipe, num_workers, worker_id)
```
ejguan, line 678: Ditto

test/test_local_io.py
```diff
@@ -655,6 +655,42 @@ def filepath_fn(name: str) -> str:
...
+
+        from torch.utils.data import DataLoader
+
+        num_workers = 2
+        line_lengths = []
+        dl = DataLoader(saver_dp, num_workers=num_workers, worker_init_fn=init_fn)
```
ejguan, line 684: Could you please add multiprocessing context here?
```suggestion
        dl = DataLoader(saver_dp, num_workers=num_workers, worker_init_fn=init_fn, multiprocessing_context="spawn")
```

ejguan (commented):
Could you please fix the followings before landing? Thank you!

NivekT (approved):

test/test_local_io.py
```diff
@@ -626,35 +629,60 @@ def test_io_path_file_loader_iterdatapipe(self):
...
+    @skipIfNoIoPath
+    def test_io_path_saver_file_lock(self):
+        # Same filename with different name
+        name_to_data = {"1.txt": b"DATA1", "1.txt": b"DATA2", "2.txt": b"DATA3", "2.txt": b"DATA4"}  # noqa: F601
+
+        # Add sharding_filter to shard data into 2
+        source_dp = IterableWrapper(list(name_to_data.items())).sharding_filter()
+
+        # Use appending as the mode
+        saver_dp = source_dp.save_by_iopath(filepath_fn=self._filepath_fn, mode="ab")
```
ejguan, line 667: `self` might not global enough. Could you make this function out of this class?

test/test_local_io.py
```diff
@@ -626,35 +629,60 @@ def test_io_path_file_loader_iterdatapipe(self):
...
+
+        import torch.utils.data.graph_settings
+
+
+        from torch.utils.data import DataLoader
+
+        num_workers = 2
+        line_lengths = []
+        dl = DataLoader(saver_dp, num_workers=num_workers, worker_init_fn=self._init_fn, multiprocessing_context="spawn")
```
ejguan, line 676: Same as `_init_fn`

ejguan (commented):
Could you please fix lint as well in this file? Really appreciate.

msaroufim:
Note to self - there's now a genuine failure since I refactored `self.fn` to be a global function in the test class, as far as this saver goes, I need to make sure this gets fixed

```python
    def __iter__(self) -> Iterator[str]:
        for filepath, data in self.source_datapipe:
            if self.fn is not None:
>               filepath = self.fn(filepath)
E               TypeError: filepath_fn() missing 1 required positional argument: 'name'
E               This exception is thrown by __iter__ of SaverIterDataPipe()

```

test/test_local_io.py
```diff
@@ -280,29 +292,28 @@ def is_nonempty_json(path_and_stream):
             len(json_dp)
 
     def test_saver_iterdatapipe(self):
-        def filepath_fn(name: str) -> str:
-            return os.path.join(self.temp_dir.name, os.path.basename(name))
-
         # Functional Test: Saving some data
         name_to_data = {"1.txt": b"DATA1", "2.txt": b"DATA2", "3.txt": b"DATA3"}
         source_dp = IterableWrapper(sorted(name_to_data.items()))
         saver_dp = source_dp.save_to_disk(filepath_fn=filepath_fn, mode="wb")
```
ejguan, line 298: Original `filepath_fn` only takes a single parameter, but now it requires two. So, we have to use `partial` from python `functools` to add `self.temp_dir.name` to this new function.
```suggestion
        saver_dp = source_dp.save_to_disk(filepath_fn=partial(filepath_fn, self.temp_dir.name), mode="wb")`
```

test/test_local_io.py
```diff
@@ -280,29 +292,28 @@ def is_nonempty_json(path_and_stream):
...
+        expected_paths = [filepath_fn(self.tmp_dir.name, name) for name in name_to_data.keys()]
         self.assertEqual(expected_paths, res_file_paths)
         for name in name_to_data.keys():
-            p = filepath_fn(name)
+            p = filepath_fn(self.tmp_dir.name, name)
             with open(p) as f:
                 self.assertEqual(name_to_data[name], f.read().encode())
 
         # Reset Test:
         saver_dp = Saver(source_dp, filepath_fn=filepath_fn, mode="wb")
```
ejguan, line 308: Ditto

test/test_local_io.py
```diff
@@ -582,9 +593,6 @@ def test_decompressor_iterdatapipe(self):
             len(tar_decompress_dp)
 
     def _write_text_files(self):
-        def filepath_fn(name: str) -> str:
-            return os.path.join(self.temp_dir.name, os.path.basename(name))
-
         name_to_data = {"1.text": b"DATA", "2.text": b"DATA", "3.text": b"DATA"}
         source_dp = IterableWrapper(sorted(name_to_data.items()))
         saver_dp = source_dp.save_to_disk(filepath_fn=filepath_fn, mode="wb")
```
ejguan, line 598: Ditto.

test/test_local_io.py
```diff
@@ -626,35 +634,59 @@ def test_io_path_file_loader_iterdatapipe(self):
 
     @skipIfNoIoPath
     def test_io_path_saver_iterdatapipe(self):
-        def filepath_fn(name: str) -> str:
-            return os.path.join(self.temp_dir.name, os.path.basename(name))
-
         # Functional Test: Saving some data
         name_to_data = {"1.txt": b"DATA1", "2.txt": b"DATA2", "3.txt": b"DATA3"}
         source_dp = IterableWrapper(sorted(name_to_data.items()))
         saver_dp = source_dp.save_by_iopath(filepath_fn=filepath_fn, mode="wb")
```
ejguan, line 640: Ditto

ejguan:
I think you might need to rebase onto `main` branch again. There are lots of changes that are not supposed to be in this PR.

PR #401, opened 2022-05-13
file decoder

tmbdev:
This PR adds a filter that simplifies file decoding for archive files. It allows glob patterns to be matched against a stream of (filename, stream) pairs and invokes the corresponding decoder on the stream.

### DIFF
--- a/torchdata/datapipes/iter/util/filedecoder.py
+++ b/torchdata/datapipes/iter/util/filedecoder.py
@@ -1,0 +1,98 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
+# All rights reserved.
+#
+# This source code is licensed under the BSD-style license found in the
+# LICENSE file in the root directory of this source tree.
+
+import pickle
+import re
+from fnmatch import fnmatch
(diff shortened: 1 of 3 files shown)


torchdata/datapipes/iter/util/filedecoder.py
```diff
@@ -0,0 +1,98 @@
...
+def _decode_text(stream):
+    binary = stream.read()
+    return binary.decode("utf-8")
+
+
+def _decode_pickle(stream):
+    return pickle.load(stream)
+
+
+default_decoders = [
```
msaroufim, line 29: @wenleix I like this feels like a natural place to add more exotic decoders like NVJpeg

tmbdev: Image decoders depend on third party libraries. Furthermore, the output people desire from image decoders (PIL, tensor, ndarray, ...) depends on what data augmentation pipeline they use. That's why image decoders aren't present by default. Users can select the image decoder they want to use with just a single line.

Of course, if you prefer, we could pick a default for image decoding.

There are a few more default decoders one might want to add (e.g., JSON).

PR #404, opened 2022-05-13
incremental shuffle

tmbdev:
This PR adds a filter that performs inline shuffling. Unlike the default shuffle, the incremental shuffle permits fast startup by having both a minimal buffer size and a maximum buffer size. As soon as more than the minimum buffer size samples have been read, samples are passed on to the next stage. 

This filter is commonly used with WebDataset to shuffle training samples inline.


msaroufim:
@NicolasHug @abhinavarora

PR #406, opened 2022-05-13
extract keys

tmbdev:
This PR adds an ExtractKeys filter that turns samples represented as dictionaries into tuples. Tuples are constructed by selecting values from the dictionaries by matching the key against a given set of patterns.

### DIFF
--- a/test/test_iterdatapipe.py
+++ b/test/test_iterdatapipe.py
@@ -32,6 +32,7 @@ from torchdata.datapipes.iter import (
     MapKeyZipper,
     MaxTokenBucketizer,
     ParagraphAggregator,
+    ExtractKeys,
     Rows2Columnar,
     SampleMultiplexer,
     UnZipper,
@@ -901,6 +902,19 @@ class TestIterDataPipe(expecttest.TestCase):
         output_dp = input_dp1.mux_longest(input_dp_no_len)
         with self.assertRaises(TypeError):
             len(output_dp)
+
+    def test_extractor(self):
+
+        # Functional Test: verify that extracting by patterns yields correct output
+        stage1 = IterableWrapper([
+            {"1.txt": "1", "1.bin": "1b"},
+            {"2.txt": "2", "2.bin": "2b"},
+        ])
+        stage2 = ExtractKeys(stage1, "*.txt", "*.bin")
+        output = list(iter(stage2))
+        assert len(output) == 2
+        assert output[0][0] == "1"
+        assert output[0][1] == "1b"
 
     def test_zip_longest_iterdatapipe(self):
 
--- a/torchdata/datapipes/iter/util/extractkeys.py
+++ b/torchdata/datapipes/iter/util/extractkeys.py
@@ -1,0 +1,59 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
+# All rights reserved.
+#
+# This source code is licensed under the BSD-style license found in the
+# LICENSE file in the root directory of this source tree.
+
+from fnmatch import fnmatch
+from typing import Dict, Iterator, Tuple
+
+from torchdata.datapipes import functional_datapipe
+from torchdata.datapipes.iter import IterDataPipe
+
+
+@functional_datapipe("extract_keys")
+class ExtractKeysIterDataPipe(IterDataPipe[Dict]):
+    r"""
+    Given a stream of dictionaries, return a stream of tuples by selecting keys using glob patterns.
+
+    Args:
+        source_datapipe: a DataPipe yielding a stream of dictionaries.
+        duplicate_is_error: it is an error if the same key is selected twice (True)
+        ignore_missing: skip any dictionaries where one or more patterns don't match (False)
+        *args: list of glob patterns or list of glob patterns
+
+    Returns:
+        a DataPipe yielding a stream of tuples
+
+    Examples:
+        >>> dp = FileLister(...).load_from_tar().webdataset().decode(...).extract_keys(["*.jpg", "*.png"], "*.gt.txt")
+    """
+
+    def __init__(
+        self, source_datapipe: IterDataPipe[Dict], *args, duplicate_is_error=True, ignore_missing=False
+    ) -> None:
+        super().__init__()
+        self.source_datapipe: IterDataPipe[Dict] = source_datapipe
+        self.duplicate_is_error = duplicate_is_error
+        self.patterns = args
+        self.ignore_missing = ignore_missing
+
+    def __iter__(self) -> Iterator[Tuple]:
+        for sample in self.source_datapipe:
+            result = []
+            for pattern in self.patterns:
+                pattern = [pattern] if not isinstance(pattern, (list, tuple)) else pattern
+                matches = [x for x in sample.keys() if any(fnmatch(x, p) for p in pattern)]
+                if len(matches) == 0:
+                    if self.ignore_missing:
+                        continue
+                    else:
+                        raise ValueError(f"Cannot find {pattern} in sample keys {sample.keys()}.")
+                if len(matches) > 1 and self.duplicate_is_error:
+                    raise ValueError(f"Multiple sample keys {sample.keys()} match {pattern}.")
+                value = sample[matches[0]]
+                result.append(value)
+            yield tuple(result)
+
+    def __len__(self) -> int:
+        return len(self.source_datapipe)
--- a/torchdata/datapipes/iter/__init__.py
+++ b/torchdata/datapipes/iter/__init__.py
@@ -109,7 +109,12 @@ from torchdata.datapipes.iter.util.tfrecordloader import (
     TFRecordLoaderIterDataPipe as TFRecordLoader,
 )
 from torchdata.datapipes.iter.util.unzipper import UnZipperIterDataPipe as UnZipper
-from torchdata.datapipes.iter.util.webdataset import WebDatasetIterDataPipe as WebDataset
+from torchdata.datapipes.iter.util.webdataset import (
+    WebDatasetIterDataPipe as WebDataset,
+)
+from torchdata.datapipes.iter.util.extractkeys import (
+    ExtractKeysIterDataPipe as ExtractKeys,
+)
 from torchdata.datapipes.iter.util.xzfileloader import (
     XzFileLoaderIterDataPipe as XzFileLoader,
     XzFileReaderIterDataPipe as XzFileReader,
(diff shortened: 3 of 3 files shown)


torchdata/datapipes/iter/util/extractkeys.py
```diff
@@ -0,0 +1,59 @@
...
+            result = []
+            for pattern in self.patterns:
+                pattern = [pattern] if not isinstance(pattern, (list, tuple)) else pattern
+                matches = [x for x in sample.keys() if any(fnmatch(x, p) for p in pattern)]
+                if len(matches) == 0:
+                    if self.ignore_missing:
+                        continue
+                    else:
+                        raise ValueError(f"Cannot find {pattern} in sample keys {sample.keys()}.")
+                if len(matches) > 1 and self.duplicate_is_error:
```
VitalyFedyunin, line 52: It does not checks if there are multiple matches across patterns arrays: `a.jpg` will match
`[[a*],[*.jpg]]` but will not trigger `Multiple sample keys`.

It would be more efficient to flattern patterns before entering this loop.

VitalyFedyunin (commented):
This DataPipe looks useful for other cases as well if me modify it a bit.

Could you please separate it by two pipes:
- 1st one to filter dict keys based on pattern:

```python
        stage1 = IterableWrapper([
            {"1.txt": "1", "1.bin": "1b", "3.jpg":"foo"},
            {"2.txt": "2", "2.bin": "2b"},
        ])
        stage2 = ExtractKeys(stage1, "*.txt", "*.bin")
        output = list(iter(stage2))
        self.assertEqual({"1.txt": "1", "1.bin": "1b"}, output[0])
```

- Second is simple map, to drop keys:
```python
dp = dp.map(lambda x: x.values())
```

test/test_iterdatapipe.py
```diff
@@ -902,6 +903,19 @@ def test_mux_longest_iterdatapipe(self):
...
+    def test_extractor(self):
+
+        # Functional Test: verify that extracting by patterns yields correct output
+        stage1 = IterableWrapper([
+            {"1.txt": "1", "1.bin": "1b"},
+            {"2.txt": "2", "2.bin": "2b"},
+        ])
+        stage2 = ExtractKeys(stage1, "*.txt", "*.bin")
+        output = list(iter(stage2))
+        assert len(output) == 2
```
VitalyFedyunin, line 915: Please replace `assert` with `self.assertEqual` as they do way better error formatting.

test/test_iterdatapipe.py
```diff
@@ -902,6 +903,19 @@ def test_mux_longest_iterdatapipe(self):
...
+        assert output[0][0] == "1"
+        assert output[0][1] == "1b"
+
```
VitalyFedyunin, line 918: Missing boolean flags tests.

torchdata/datapipes/iter/util/extractkeys.py
tmbdev: I'm detecting both now.

test/test_iterdatapipe.py
tmbdev: done

test/test_iterdatapipe.py
tmbdev: done (I think)

NivekT (commented):
Hi @tmbdev,

Thanks for your commits on these PRs. Let us know if these are ready for review (but no rush at all!). @VitalyFedyunin and I will be happy to have a look.

Again, thanks for contributing to our library!

torchdata/datapipes/iter/util/extractkeys.py
```diff
@@ -0,0 +1,73 @@
...
+
+from fnmatch import fnmatch
+from typing import Dict, Iterator, Tuple, Union
+
+from torchdata.datapipes import functional_datapipe
+from torchdata.datapipes.iter import IterDataPipe
+
+
+@functional_datapipe("extract_keys")
+class ExtractKeysIterDataPipe(IterDataPipe[Dict]):
```
NivekT, line 15: Can we rename this to `KeyExtractor` to follow our naming convention? Thanks.

We can still keep `"extract_keys"` as the functional name.

test/test_iterdatapipe.py
```diff
@@ -951,6 +952,30 @@ def test_mux_longest_iterdatapipe(self):
         with self.assertRaises(TypeError):
             len(output_dp)
 
+    def test_extractor(self):
```
NivekT, line 955:
```suggestion
    def test_key_extractor(self):
```
nit: We used to have a different `extractor`

torchdata/datapipes/iter/util/extractkeys.py
```diff
@@ -0,0 +1,73 @@
...
+    r"""
+    Given a stream of dictionaries, return a stream of tuples by selecting keys using glob patterns.
+
+    Args:
+        source_datapipe: a DataPipe yielding a stream of dictionaries.
+        duplicate_is_error: it is an error if the same key is selected twice (True)
+        ignore_missing: skip any dictionaries where one or more patterns don't match (False)
```
NivekT, line 22:
```suggestion
```
Duplicate lines of descriptions

torchdata/datapipes/iter/util/extractkeys.py
```diff
@@ -0,0 +1,73 @@
...
+        *args: list of glob patterns or list of glob patterns
+        duplicate_is_error: it is an error if the same key is selected twice (True)
+        ignore_missing: allow patterns not to match (i.e., incomplete outputs)
+        as_tuple: return a tuple instead of a dictionary
```
NivekT, line 26:
```suggestion
        as_tuple: return a tuple instead of a dictionary (True or False here)
```

torchdata/datapipes/iter/util/extractkeys.py
```diff
@@ -0,0 +1,73 @@
...
+
+    Returns:
+        a DataPipe yielding a stream of tuples
+
+    Examples:
+        >>> dp = FileLister(...).load_from_tar().webdataset().decode(...).extract_keys(["*.jpg", "*.png"], "*.gt.txt")
+    """
+
+    def __init__(
+        self, source_datapipe: IterDataPipe[Dict], *args, duplicate_is_error=True, ignore_missing=False, as_tuple=False
```
NivekT, line 36: Do we want to default `as_tuple=False`? Based on the docstring I would've guessed you wanted `True` instead.
```suggestion
        self, source_datapipe: IterDataPipe[Dict], *args, duplicate_is_error: bool = True, ignore_missing: bool = False, as_tuple: bool = False
```

torchdata/datapipes/iter/util/extractkeys.py
```diff
@@ -0,0 +1,73 @@
...
+                    result.append(value)
+                else:
+                    result.append((matches[0], value))
+            if self.as_tuple:
+                yield tuple(result)
+            else:
+                yield {k: v for k, v in result}
+
+    def __len__(self) -> int:
+        return len(self.source_datapipe)
```
NivekT, line 73:
```suggestion
```

I think we should remove these lines because the length will be different from the source and we do not know the length in advance.

torchdata/datapipes/iter/util/extractkeys.py
```diff
@@ -0,0 +1,73 @@
...
+class ExtractKeysIterDataPipe(IterDataPipe[Dict]):
+    r"""
+    Given a stream of dictionaries, return a stream of tuples by selecting keys using glob patterns.
+
+    Args:
+        source_datapipe: a DataPipe yielding a stream of dictionaries.
+        duplicate_is_error: it is an error if the same key is selected twice (True)
+        ignore_missing: skip any dictionaries where one or more patterns don't match (False)
+        *args: list of glob patterns or list of glob patterns
+        duplicate_is_error: it is an error if the same key is selected twice (True)
```
NivekT, line 24:
```suggestion
        duplicate_is_error: it is an error if the same key is selected twice (True), otherwise returns the first matched value
```

torchdata/datapipes/iter/util/extractkeys.py
```diff
@@ -0,0 +1,73 @@
...
+                matches = [x for x in sample.keys() if any(fnmatch(x, p) for p in pattern)]
+                if len(matches) == 0:
+                    if self.ignore_missing:
+                        continue
+                    else:
+                        raise ValueError(f"extract_keys: cannot find {pattern} in sample keys {sample.keys()}.")
+                if len(matches) > 1 and self.duplicate_is_error:
+                    raise ValueError(f"extract_keys: multiple sample keys {sample.keys()} match {pattern}.")
+                if matches[0] in used and self.duplicate_is_error:
+                    raise ValueError(f"extract_keys: key {matches[0]} is selected twice.")
```
NivekT, line 60:
```suggestion
                    raise ValueError(f"extract_keys: key {matches[0]} is selected twice by multiple patterns.")
```
nit

torchdata/datapipes/iter/util/extractkeys.py
```diff
@@ -0,0 +1,73 @@
...
+from typing import Dict, Iterator, Tuple, Union
+
+from torchdata.datapipes import functional_datapipe
+from torchdata.datapipes.iter import IterDataPipe
+
+
+@functional_datapipe("extract_keys")
+class ExtractKeysIterDataPipe(IterDataPipe[Dict]):
+    r"""
+    Given a stream of dictionaries, return a stream of tuples by selecting keys using glob patterns.
```
NivekT, line 17:
```suggestion
    Given a stream of dictionaries, return a stream of dicts (or tuples) by selecting keys using glob patterns.
```

torchdata/datapipes/iter/util/extractkeys.py
```diff
@@ -0,0 +1,73 @@
...
+        duplicate_is_error: it is an error if the same key is selected twice (True)
+        ignore_missing: allow patterns not to match (i.e., incomplete outputs)
+        as_tuple: return a tuple instead of a dictionary
+
+    Returns:
+        a DataPipe yielding a stream of tuples
+
+    Examples:
+        >>> dp = FileLister(...).load_from_tar().webdataset().decode(...).extract_keys(["*.jpg", "*.png"], "*.gt.txt")
+    """
```
NivekT, line 33: In addition to the one example with webdataset, please add an example with sample outputs here. Copying from the test cases is totally fine to me.

NivekT (commented):
A few comments. Feel free to not accept every requested change.

Again, thank you so much for your contribution!

torchdata/datapipes/iter/util/extractkeys.py
NivekT: nit: `allow_duplicate` might be a better name than `duplicate_is_error`

PR #414, opened 2022-05-17
Make requirements.txt single source of truth

### LINKED ISSUE #410: Make requirements.txt as the single source of truth for TorchData dependencies
### 🚀 The feature

Currently, we have two locations containing TorchData requirements:
- https://github.com/pytorch/data/blob/main/requirements.txt
- https://github.com/pytorch/data/blob/3c7769682b5263acd889c7e7f0187323f7d9e9da/setup.py#L108-L112
We can let `setup.py` to decode `requirements.txt` to generate requirements automatically.

### Motivation, pitch

It will be safer and easy to maintain.

### Alternatives

_No response_

### Additional context

_No response_

tal66:
Hi, this fixes #410.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -100,13 +100,14 @@ def _export_version(version, sha):
 
 def _get_requirements():
     req_list = []
-    with Path('requirements.txt').open('r') as f:
+    with Path("requirements.txt").open("r") as f:
         for line in f:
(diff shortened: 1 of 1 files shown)


tal66:
*I signed, but the CLA check didn't automatically update here.

PR #415, opened 2022-05-17
Use manylinux2014 & enable macos for AWSSDK

ejguan:
TorchText binary uses `manylinux2014`. And, I tested both `torch` and `torchtext` is using the ABI `manylinux2014`. See failure in https://github.com/pytorch/text/pull/1731

- Change target platform to manylinux2014 for linux (See workflow: https://github.com/pytorch/data/runs/6479078172?check_suite_focus=true#step:9:87)
  - `manylinux_2_17_x86_64` is an alias of `manylinux2014`
  - See [PEP600](https://peps.python.org/pep-0600/)
- Re-enable MacOS binary with AWSSDK after https://github.com/pytorch/text/pull/1728 is landed.

See workflow: https://github.com/pytorch/data/actions/runs/2341532489

cc: @bearzx @wenleix  I think TorchArrow might need to align with such requirements `manylinux2014` since TorchText will be a downstream lib for you.


.github/workflows/_build_test_upload.yml
```diff
@@ -52,6 +52,7 @@ jobs:
   wheel_build_test:
     needs: get_release_type
     runs-on: ${{ matrix.os }}
+    container: ${{ startsWith( matrix.os, 'ubuntu' ) && 'quay.io/pypa/manylinux2014_x86_64' || null }}
```
ejguan, line 55: I am using official docker from pypa

third_party/CMakeLists.txt
```diff
@@ -70,7 +70,7 @@ else()
     aws_sdk
     SOURCE_DIR ${aws_cpp_sdk_SOURCE_DIR}
     INSTALL_DIR ${aws_cpp_sdk_INSTALL}
-    CMAKE_ARGS "-DBUILD_SHARED_LIBS=OFF" "-DBUILD_ONLY=transfer;s3" "-DENABLE_TESTING=OFF" "-DCMAKE_BUILD_TYPE=Release" "-DCMAKE_INSTALL_PREFIX=${aws_cpp_sdk_INSTALL}"
+    CMAKE_ARGS "-DBUILD_SHARED_LIBS=OFF" "-DBUILD_ONLY=transfer;s3" "-DENABLE_TESTING=OFF" "-DCMAKE_BUILD_TYPE=Release" "-DCMAKE_INSTALL_PREFIX=${aws_cpp_sdk_INSTALL}" "-DCMAKE_INSTALL_LIBDIR=lib"
```
ejguan, line 73: `CMAKE_INSTALL_LIBDIR` is needed because the lib directory for AWSSDK can vary between `lib` and `lib64`

tools/setup_helpers/extension.py
```diff
@@ -100,11 +99,6 @@ def build_extension(self, ext):
 
         build_args = ["--config", cfg]
 
-        if _BUILD_S3 and _AWSSDK_DIR:
-            cmake_args += [
-                f"-DAWSSDK_DIR={_AWSSDK_DIR}",
-            ]
```
ejguan, line 106: Removing `AWSSDK_DIR` as it's redundant now because it's compiled in third-party library. And, users can use system-wise `AWSSDK` if needed with environment variable `USE_SYSTEM_AWS_SDK_CPP`

VitalyFedyunin (approved):

NivekT (approved):

PR #421, opened 2022-05-18
Change to manylinux1 based on last release

ejguan:
Per title


.github/workflows/ci.yml
```diff
@@ -65,7 +62,7 @@ jobs:
         run: |
           python setup.py install
         env:
-          BUILD_S3: ${{ matrix.with-s3 }}
+          BUILD_S3: 1
```
NivekT, line 65: Is there any situation where we do not enable `BUILD_S3`? 

It looks like only when the setup is Windows with conda, why is that?

NivekT (approved):
LGTM! Thanks for making all these changes!

Can you elaborate on "Make our testing CI running with AWSSDK enabled by default to save more time/money to run CI tests"? How does that work?

ejguan:
> Make our testing CI running with AWSSDK enabled by default to save more time/money to run CI tests"?

By simply removing one test case of `BUILD_S3=0`.

.github/workflows/ci.yml
ejguan: For binary releases, it's just because there are too many Error that Idk how to solve with AWS enabled for conda on windows.

For CI testing, you can treat it as wheels, we could enable testing with awssdk.

Generally, the reason I think it makes sense to always enable AWSSDK in our testing CI is the >95% tests are the same with or without AWSSDK enabled.

PR #422, opened 2022-05-19
data/benchmarks

### LINKED ISSUE #416: `data/benchmarks/`
### 🚀 The feature

We're proposing a folder to hold all benchmark scripts which would be easily reproducible by anyone from the core PyTorch Data team, PyTorch domain teams and the broader community. original author @vitaly-fedyunin 

### Motivation, pitch

As `pytorch/data` gains more widespread adoption there's going to be more questions about its performance. So it's important to have reusable reproducible scripts we can use. The dev team also needs to be able to monitor for regressions between releases and use benchmarks to inform additional performance optimizations

The script should be runnable with clear instructions and dependencies in a `README.md`, it should be possible to run the same script in CI with no changes. The script should output metrics in a human-readable markdown file

The main metric we're going to take a look is time to convergence in training against the traditional Dataset baseline using both DataLoader v1 and the experimental DataLoaderv2. The second most important metric is model accuracy to make sure we don't degrade training performance too much (see shuffling issues)

The final outcome should be support the cross product of all of the below configurations.

**Datasets**

- [ ] COCO
- [ ] ImageNet (if license allows)
- [ ] [The Pile](https://pile.eleuther.ai/)
- [ ] [mc4](https://huggingface.co/datasets/mc4 )

**Models**
- [ ] Resnet50
- [ ] Resnet128
- [ ] BERT-B

**Storage configuration**
* SSD 
* HDD
* NFS
* Cloud (S3)
* Web (HTTP) 

**Other Metrics**
* Things to track:
* Time per batch
* Time per epoch
* Precision Over Time
* CPU load
* GPU load (starvation)
* Memory Usage

### Alternatives

_No response_

### Additional context

For each of the datasets continuously track the next implementations:
* Baseline implementation (before conversion to DataPipes).
* Migrated DataSet (to DataPipes), version as-is from Vision GitHub. 
* Option with Tar (or other simple archiving).
* Option with Data Preproc (rearrange/repack/reformat) - using DataPipes & serialization, consider the speed of repacking too).
* For Vision: Use NVidia Webdataset DataLoader and NVidia DALI DataLoader for comparison.
* For Text: Use the HuggingFace dataset for comparison.

Ideally we would put one of these large datasets in an S3 bucket but they will throttle it so instead it’s best to setup an EC2 instance with a simple http server that makes the dataset available on an attached SSD disk which will allow us to do single node 8 GPU experiments. For multi-nodes we need to come up with a story for distributed storage.

The main metrics we need to look at 
* GPU utilization - higher is better we want preprocessing and training to happen concurrently
* Data throughput per datapipe to measure any bottlenecks - more useful for data pipe authors. Can add some simple telemetry by default to any datapipe
* End user is going to be looking at time to epoch and any accuracy loss

msaroufim:
creating your pull request.

Fixes #416 

### Changes

- Added CLI to run various benchmarks on various datasets and measure key metrics


benchmarks/README.md
```diff
@@ -0,0 +1,34 @@
+# Install dependencies
+
+```
+pip3 install --pre torch torchvision torchaudio --extra-index-url https://download.pytorch.org/whl/nightly/cu113
```
NicolasHug, line 4:
```suggestion
pip3 install --pre torch torchvision --extra-index-url https://download.pytorch.org/whl/nightly/cu113
```

benchmarks/run_benchmark.py
```diff
@@ -0,0 +1,128 @@
...
+    from torch.utils.data.dataloader_experimental import DataLoader2 as DataLoader
+else:
+    raise(f"dataloaderv{dataloaderv} is not a valid option")
+
+# Util function for multiprocessing
+def init_fn(worker_id):
+    info = torch.utils.data.get_worker_info()
+    num_workers = info.num_workers
+    datapipe = info.dataset
+    torch.utils.data.graph_settings.apply_sharding(datapipe, num_workers, worker_id)
```
NicolasHug, line 50: We should make sure we can remove this eventually as users shouldn't have to shard manually with DataLoaderV2

msaroufim:
Ok we can now get a trace and have an end to end example working, data loading is not the bottleneck here yet so will keep experimenting

image

msaroufim:
Screen Shot 2022-05-25 at 5 28 51 PM

msaroufim:
Next thing I'd like to try out is including these datasets from torchtext.datasets import amazonfullreview

msaroufim:
profile

Screen Shot 2022-05-26 at 5 14 02 PM

NicolasHug (commented):
Gave it a quick look before our meeting!

benchmarks/run_benchmark.py
```diff
@@ -0,0 +1,118 @@
...
+
+
+
+total_start = time.time()
+per_epoch_durations = []
+batch_durations = []
+
+
+
+with torch.profiler.profile(
```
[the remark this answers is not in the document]
msaroufim: Yeah it does work, did require putting `p.step()` in the right place

msaroufim:
Discussion with Vitaly June 21
* Will focus on running on mc4 with dataloader v1 on various hardware configurations (SSD, HDD) and use a few starter cloudformation templates to make this easier

benchmarks/report.py
```diff
@@ -0,0 +1,6 @@
+from statistics import mean
+
+def create_report(per_epoch_durations, batch_durations, total_duration):
+    print(f"Total duration is {total_duration}")
+    print(f"Per epoch duration {mean(per_epoch_durations)}")
+    print(f"Per batch duration {mean(batch_durations)}")
```
NivekT, line 6: nit: Showing percentile stats (e.g. 1%, 25%, 50%, 75%, 99%) may be more helpful.

benchmarks/utils.py
```diff
@@ -0,0 +1,8 @@
+# Util function for multiprocessing
+import torch
+
+def init_fn(worker_id):
```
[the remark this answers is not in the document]
NivekT: I think with the recent changes, manual sharding  should no longer be needed.

benchmarks/datasets.py
```diff
@@ -0,0 +1,28 @@
...
+        l.append(int(char))
+    return l
+
+def prepare_gtsrb_datapipe(batch_size, device, dp):
+    # Filter out bounding box and path to image
+    dp = dp.map(lambda sample : {"image" : sample["image"], "label" : sample["label"]})
+
+    # Apply image preprocessing
+    dp = dp.map(lambda sample : transform(sample.decode().to(torch.device(device))), input_col="image")
+    dp = dp.map(lambda sample : torch.tensor(str_to_list(sample.to_categories())).to(torch.device(device)), input_col="label")
```
NivekT, line 24: I wonder if the usage of `lambda` will lead to issues with multiprocessing if `dill` isn't installed. Or is `dill` always installed here?

benchmarks/run_benchmark.py
```diff
@@ -0,0 +1,118 @@
...
+    if ispipe:
+        dp = loadpipe(dataset, split="train")
+        logging.debug(f"data format before preprocessing is {next(iter(dp))}")
+
+        dp = prepare_gtsrb_datapipe(batch_size, device, dp)
+        logging.debug(f"data format after preprocessing is \n {next(iter(dp))}\n")
+
+    else:
+        # No further preprocessing needed this returns a tuple of Images and labels as ints
+        # Do I need to do batching and collation manually?
```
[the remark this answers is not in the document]
NivekT: Nicolas is right that you can do that for `DataLoader`.

I would think it is better to use `.batch` since it will be necessary for `DataLoaderV2`. Then you can pass the same DataPipe to both versions of DL without more changes later.

NivekT (commented):
Overall, LGTM

nit: Need copyright headers for `.py` files

msaroufim:
> Overall, LGTM with a few comments. Let me know what you additional features you plan to add.
> 
> nit: Need copyright headers for `.py` files

Thanks @NivekT will address all your feedback - As far as new features to add for this PR not much I think there's a bunch of cleanup I need to do
* Clean up report into its own dataclass which you can then export to whatever format you want: html, md, csv etc..
* Address all your feedback
* Some more cleanup

And I think the next PR should be focused around integrating the aws cli into CI where we can benchmark a distributed systems setup per @NicolasHug's request

And after that we can see which of the partner integrations should be added to this setup as weell

benchmarks/run_benchmark.py
```diff
@@ -0,0 +1,118 @@
...
+    logging.debug(nvidiasmi)
+
+lscpu = subprocess.check_output("lscpu", shell=True, text=True)
+logging.debug(lscpu)
+
+
+if dataloaderv == 1:
+    from torch.utils.data import DataLoader
+elif dataloaderv == 2:
+    from torch.utils.data.dataloader_experimental import DataLoader2 as DataLoader
```
VitalyFedyunin, line 39: Please avoid using this one, perhaps you need to create your own wrapper, which will use DLv2 from torchdata repo and automatically create MultiProcessingReadingService

VitalyFedyunin (commented):
Need to drop usage of dataloader_experimental.

benchmarks/run_benchmark.py
VitalyFedyunin: This is up to @NivekT , if he wants to have this update as follow-up or within this PR.

benchmarks/datasets.py
msaroufim: I can just add dill to the list of dependencies in my own requirements.txt

benchmarks/run_benchmark.py
msaroufim: @NivekT do you have more context, I'm not sure I follow what change is being requested

benchmarks/report.py
msaroufim: Revamping this to include a MetricCache from which any metrics can then be exported more easily

benchmarks/report.py
```diff
@@ -0,0 +1,50 @@
...
+from dataclasses import dataclass, fields
+from statistics import mean
+from typing import Dict, list, tuple
+
+import numpy as np
+
+duration = int
+
+
+@dataclass
```
msaroufim, line 12: @NivekT what do you think about this kind of reporting instead? Probably still a few bugs around but just wanna make sure I'm not overengineering things

benchmarks/run_benchmark.py
NicolasHug: I think `dataloader_experimental.DataLoader2` is bound to be removed eventually. The actual `DataLoader2` is in `torchdata.dataloader2`.

I have an example in my benchmarking PR, if you're interested https://github.com/pytorch/vision/pull/6196/files#diff-32b42103e815b96c670a0b5f0db055fe63f10fc8776ccbb6aa9b61a6940abba0R207-R211

NicolasHug:
@msaroufim @VitalyFedyunin @NivekT following up on my earlier comments in https://github.com/pytorch/data/issues/416#issuecomment-1164404834 I also have a separate PR that already provides support for **the cross-product** of:

- Distributed Learning (DDP) **vs** 1-GPU training
- Datapipes (with DataLoader or `torchdata.dataloader2`) **vs** Iterable datasets (non-DP) **vs** MapStyle Datasets
- Full training procedure **or** Data-loading only (with or without transforms) **or** Model training only
- Timing of data-loading **vs** model training
- any classification model from torchvision 

(It also has FFCV support, but that's less relevant for us here).

Since it's directly adapted from torchvision recipes, it's also a bit closer to the kind of training that users would be doing in the wild.

Do you think it would make sense to join our benchmarking efforts here? I'm happy to provide support if you'd like to collaborate.

CC @nairbv

NivekT:
@NicolasHug  I am in the processing of going through both setups, running them on our AWS cluster, and identifying the differences. I agree that combining the efforts is the right approach. Let me dig a bit deeper first and I can schedule a meeting for all of us to chat.

msaroufim:
@NicolasHug I think the right way to divide this up would be
* I work on the infra setup, the benchmark artifact and the benchmark export
* I leverage your model training scripts since you're the domain expert

I would like to also eventually do something like pull any of the HF datasets and just benchmark there but I don't believe the datasets there give me sufficient information to create a toy model with the right shapes automatically

But yeah would love to talk

benchmarks/report.py
```diff
@@ -0,0 +1,52 @@
...
+class MetricCache:
+    epoch_durations: list[duration]
+    batch_durations: list[duration]
+    total_duration: int = 0
+
+
+class MetricExporter(ABC):
```
NicolasHug, line 19: Following up on our discussion from Monday, I took a look at this and I think it should be reasonably simple to plug it into the other `MetricLogger` from the torchvision recipe (it's "metric calculator" more than a "logger", and we can remove the "log" part in favour of this one)

PR #425, opened 2022-05-19
Add portalocker to Conda & Disable tests for release

ejguan:
I have to disable tests due to #386 can't be landed until the corresponding PR/Diff in PyTorch Core is landed.
Without disabling the tests, I can't release any nightly build. Since PyTorch nightly has been updated, we should be fine.


NivekT (approved):
LGTM! Thanks!

Issue #364, opened 2022-04-19
Linter for DataPipe/DataLoader2 


NivekT:
### 🚀 The feature

This issue proposes the addition of a linter for DataPipes and DataLoader2. The linter can analyze the graph of DataPipes and input arguments to DataLoaderV, and inform the users if any errors may occur ahead of time. The incomplete list of issues that the linter may try to analyze and raise is below. Please feel free to edit the list directly to add more or comment below.

Essential:
- [ ] Multiple references to the same iterator/DataPipe
  - This can cause issue when serialized, suggest users to `fork`
- [ ] Duplicate usage of shuffle/batch/collate
- [ ] Multiprocess/distributed behavior related to sharding/shuffling

Nice-to-have:
- [ ] A
- [ ] B

### Motivation, pitch
Having a linter will encourage best practices of DataPipe usages and reduces the number of unexpected bugs/behaviors in the data loading process during runtime. 

### Alternatives
Only raise exceptions during runtime.

### Additional context
This linter is expected to work with DataPipes and DataLoaderV2. We should consider if it should work with the original DataLoader as well (and how).

cc: @VitalyFedyunin @ejguan

pmeier:
IIRC, we should enforce shuffling before sharding. cc @NicolasHug

VitalyFedyunin:
If datasource is already shuffled to some extent, shard before shuffle might be valid operation. So warning IMO suffice.

VitalyFedyunin:
We should add one more linter to check DataPipe object size and warn if it is too big (ex premature initialization of large structures).

ejguan:
#429 This PR introduces the linter to check if there is a `shuffle` before each `sharding`. This should mainly be used by domain libraries.

For user-facing linter, we might provide a variant from this linter function. Raise warning when no `shuffle` before `sharding` and raise Error/warning if there is a `shuffle` behind `sharding.

VitalyFedyunin:
[ ] Warn if `filter` appears between `on_disk_cache` and `end_caching` sections.

PR #429, opened 2022-05-19
Add linter to check if shuffle before sharding

ejguan:
This PR introduces the linter function to validate there is a shuffle operation before sharding. (Required by TorchText in https://github.com/pytorch/text/pull/1729)

- When `sharding_filter` is not presented in the graph, this function always returns `True`
- For single-path graph, `shuffle` needs to be placed before `sharding_filter`.
- For multi-path graph, any `sharding_filter` requires a `shuffle` before along the path

This linter function won't check if there are multiple `sharding_filter` in the graph or `sharding_filter` is at the right place

### DIFF
--- a/test/test_linter.py
+++ b/test/test_linter.py
@@ -7,8 +7,9 @@
 
 import unittest
 
-from torchdata.datapipes.iter import IterableWrapper, ShardingFilter, Shuffler
 from torchdata.dataloader2.linter import _check_shuffle_before_sharding
+
+from torchdata.datapipes.iter import IterableWrapper, ShardingFilter, Shuffler
 
 
 def dummy_fn(x):
--- a/test/test_linter.py
+++ b/test/test_linter.py
@@ -32,6 +32,9 @@ class LinterTest(unittest.TestCase):
         dp = source_dp.map(dummy_fn).sharding_filter()
         self.assertFalse(_check_shuffle_before_sharding(dp))
 
+        dp = source_dp.map(dummy_fn).sharding_filter().shuffle()
+        self.assertFalse(_check_shuffle_before_sharding(dp))
+
         # Multi pathes
         def _multi_path_dp_1(shuffle):
             s_dp = source_dp.shuffle() if shuffle else source_dp
--- a/test/test_linter.py
+++ b/test/test_linter.py
@@ -1,0 +1,77 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
+# All rights reserved.
+#
+# This source code is licensed under the BSD-style license found in the
+# LICENSE file in the root directory of this source tree.
+
+
+import unittest
+
+from torchdata.datapipes.iter import IterableWrapper, ShardingFilter, Shuffler
+from torchdata.dataloader2.linter import _check_shuffle_before_sharding
+
+
+def dummy_fn(x):
+    return x
+
+
+class LinterTest(unittest.TestCase):
+    def test_sharding_shuffle(self):
+        source_dp = IterableWrapper(list(range(20)))
+
+        # Single path
+        dp = source_dp.map(dummy_fn).shuffle()
+        self.assertTrue(_check_shuffle_before_sharding(dp))
+        dp = source_dp.map(dummy_fn)
+        self.assertTrue(_check_shuffle_before_sharding(dp))
+
+        dp = source_dp.map(dummy_fn).shuffle().sharding_filter()
+        self.assertTrue(_check_shuffle_before_sharding(dp))
+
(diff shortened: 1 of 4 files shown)


torchdata/dataloader2/linter.py
```diff
@@ -0,0 +1,57 @@
...
+# This source code is licensed under the BSD-style license found in the
+# LICENSE file in the root directory of this source tree.
+
+
+from torchdata.datapipes.iter import IterDataPipe, ShardingFilter, Shuffler
+
+from ._graph_utils import DataPipeGraph, traverse
+
+
+def _check_shuffle_before_sharding(datapipe: IterDataPipe) -> bool:
```
ejguan, line 13: Any suggestion on the function name is welcomed. :P

NivekT (commented):
Are we planning to expose some API to the public? Some options that come to my mind:

1. Users can directly import and call `linter` as a function
    ```python
    dp = dp.map(...).filter(...)
    linter(dp)
    ```
2. No public API, but DataLoader V2 will use call `linter` internally
    - It may be useful for users to be able to enable/disable certain rules
3. Both

test/test_linter.py
```diff
@@ -0,0 +1,78 @@
...
+
+def dummy_fn(x):
+    return x
+
+
+class LinterTest(unittest.TestCase):
+    def test_sharding_shuffle(self):
+        source_dp = IterableWrapper(list(range(20)))
+
+        # Single path
```
parmeet, line 23: Should we also test for False assertion when sharding is applied before?

```python
        dp = source_dp.map(dummy_fn).sharding_filter().shuffle()
        self.assertFalse(_check_shuffle_before_sharding(dp))
```

parmeet (approved):
Overall LGTM! Thanks for providing this functionality. I think vision might be interested to use this as well cc: @NicolasHug

ejguan:
> Are we planning to expose some API to the public? Some options that come to my mind:
> 
> 1. Users can directly import and call `linter` as a function
>    ```python
>    dp = dp.map(...).filter(...)
>    linter(dp)
>    ```
> 2. No public API, but DataLoader V2 will use call `linter` internally
>    
>    * It may be useful for users to be able to enable/disable certain rules
> 3. Both

I prefer this function being private for now. I think there are multiple linter functions needed. I would think aggregating all these linter functions together to provide a common API like `linter(dp, sharding_before_shuffle=True, ...)` with options to turn on each specific functionalities for users.

NivekT (approved):
LGTM!

VitalyFedyunin:
Agree on keeping it private. Eventually, we want to have it on auto in DLv2, but let's navigate the landscape and figure out if we want it on/off by default and if it is applicable for all Reading services or should be reading service level linters.

pmeier:
> This PR introduces the linter function to validate there is a shuffle operation before sharding

Is this the right objective? Shouldn't the objective be "validate that the shuffle operation comes before the operation *iff a shuffle operation is part of the graph*"?

Otherwise something like this will trip the linter:

```py
from torchdata.datapipes.iter import ShardingFilter, IterableWrapper
from torchdata.dataloader2.linter import _check_shuffle_before_sharding

dp = IterableWrapper([])
dp = ShardingFilter(dp)

assert _check_shuffle_before_sharding(dp)
```

ejguan:
> > This PR introduces the linter function to validate there is a shuffle operation before sharding
> 
> Is this the right objective? Shouldn't the objective be "validate that the shuffle operation comes before the sharding operation _iff a shuffle operation is part of the graph_"?
> 
> Otherwise something like this will trip the linter:
> 
> ```python
> from torchdata.datapipes.iter import ShardingFilter, IterableWrapper
> from torchdata.dataloader2.linter import _check_shuffle_before_sharding
> 
> dp = IterableWrapper([])
> dp = ShardingFilter(dp)
> 
> assert _check_shuffle_before_sharding(dp)
> ```

@pmeier 
I thought we always want to have a `shuffle` for domain library. Then, we can provide the right way to run shuffle and sharding. For users who don't want to enable `shuffle`, they still can specify shuffle to False.

pmeier:
> I thought we always want to have a `shuffle` for domain library.

That is true, but this is not a linter for domain libraries, but a general linter applied by the data loader, right? If not, you can disregard my comments.

ejguan:
> > I thought we always want to have a `shuffle` for domain library.
> 
> That is true, but this is not a linter for domain libraries, but a general linter applied by the data loader, right? If not, you can disregard my comments.

@pmeier 
I would like to keep it as it is. I added this PR simply because TorchText team asked me to add this functionality to verify their implementation.

For user-facing linter, I would rely on this #364 to tack the features. Will add your comment to the issue.

PR #430, opened 2022-05-19
Changing imports of DataPipe from Core to TorchData

NivekT:
Stack from [ghstack](https://github.com/ezyang/ghstack):
* #431
* __->__ #430

Differential Revision: [D36530195](https://our.internmc.facebook.com/intern/diff/D36530195)

PR #431, opened 2022-05-19
Adding SerializationWrapper for DataLoader2

NivekT:
* __->__ #431
* #430

Since in some cases the reading service may initialize using the original DataLoader, the wrapper may be applied twice? Although it may not cause errors, we may want to change the logic in DataLoaderV2 to only apply if the DataPipe isn't already a wrapper. Thoughts?

### DIFF
--- a/torchdata/dataloader2/dataloader2.py
+++ b/torchdata/dataloader2/dataloader2.py
@@ -9,6 +9,8 @@ import pickle
 from dataclasses import dataclass
 from typing import Any, Callable, Dict, Generic, Iterator, Optional, TypeVar
 
+from torch.utils.data.datapipes.datapipe import _IterDataPipeSerializationWrapper
+
 from torchdata.datapipes.iter import IterDataPipe
 
 from .error import PauseIteration
@@ -58,6 +60,9 @@ class DataLoader2(Generic[T_co]):
         self.reading_service = reading_service
         self.reading_service_state: Optional[bytes] = None
         self._terminated: bool = False
+
+        # _DataPipeSerializationWrapper container makes it easier to serialize without redefining pickler
+        self.datapipe = _IterDataPipeSerializationWrapper(self.datapipe)  # type: ignore[assignment]
 
         if self.datapipe_adapter_fn is not None:
             self.datapipe = self.datapipe_adapter_fn(self.datapipe)


NivekT:
@nivekt has imported this pull request.  If you are a Facebook employee, you can view this diff [on Phabricator](https://www.internalfb.com/diff/D36530194).

ejguan (commented):
I don't think it should go into `DataLoader2`, this wrapper needs to be in `MultiprocessingReadingService`.

NivekT:
> I don't think it should go into `DataLoader2`, this wrapper needs to be in `MultiprocessingReadingService`.

In that case it might not need it? I think `MultiprocessingReadingService` calls `DataLoader` which does this anyway. I will look at the code to confirm

VitalyFedyunin:
The problem with this wrapper is its appearance in stack traces and graph traversals. 

I'm still not sure if it is good (users will see where exactly `dill` is applied) or bad and we should make something like _serialization_helper(dp: DataPipe) function instead.

VitalyFedyunin:
> I don't think it should go into `DataLoader2`, this wrapper needs to be in `MultiprocessingReadingService`.

Same feeling, for that reason we better go `_serialization_helper(dp: DataPipe) -> (bytes, is_dill_used)` way. And use it inside of `_IterDataPipeSerializationWrapper` as well as inside of the serializability validator and when we pass graph to MPReadingService workers.

NivekT:
1. `MultiprocessingReadingService` uses `DataLoader`, in which the serialization wrapper is already being applied to the `datapipe` that is being passed in. Therefore, do you agree that no change is needed for `MultiprocessingReadingService` to support `dill`?
2. Do we intend for each reading service to apply the wrapper? It may be better to have something that works across reading services, whether it is in the form of a wrapper or a function.
    - Currently, there is [`serialize_datapipe`](https://github.com/pytorch/data/blob/main/torchdata/dataloader2/dataloader2.py#L22-L33), which is the function that is used for serialization within the method [`state_dict`](https://github.com/pytorch/data/blob/main/torchdata/dataloader2/dataloader2.py#L125-L143). We can definitely modify that to use `dill` when needed. However, I don't think that function is used by the reading_service for serialization.
3. Can you elaborate on using `_serialization_helper(dp: DataPipe) -> (bytes, is_dill_used)` inside `_IterDataPipeSerializationWrapper`? How does that alleviate your concern related to stack traces and graph traversal (since a wrapper is still needed)?

NivekT:
@ejguan @VitalyFedyunin Any thoughts on my comment above?

NivekT:
> I think we should have this `dill` serialization magic inside of DataLoaderV2 (or at least available to it). As we require all DataPipes graph to be serializable (when it is passed as argument or processes by adapter). However I'm concerned about wrapping everything into another DataPipe as our graphs become extremely large (to visualize and profile).
> 
> For that reason I would prefer solution like introducing `_serialization_helper(dp: DataPipe) -> (bytes, is_dill_used)` function.
> 
> In this case `is_serializable` check would look a like:
> 
> ```python
> def is_serializable(dp):
>   try:
>     _serialization_helper(dp)
>      return True
>   except:
>     return False
> ```
> 
> Passing DataPipe graph between processes would look a like:
> 
> ```python
> def _unserialization_helper(serialized_dp: Tuple):
>   data, is_dill = serialized_dp
>   if is_dill:
>     return dill.loads(data)
>   else:
>     return pickle.loads(data)
> 
> def startup(serialized_dp, arg1, arg2):
>   dp = _unserialization_helper(serialized_dp)
>   # do staff
> 
> mp.process(startup, [ _serialization_helper(dp: DataPipe), arg1, arg2 ])
> ```
> 
> Yet, we can still leave option of wrapping into `_IterDataPipeSerializationWrapper` as I have seen many cases when users want to do their own multiprocessing and need easy solution to pass Pipes.

I agree with the design, but doesn't this mean the reading service will be responsible for serialization logic?

I am going to follow up on this offline along with the plan for DataLoader V2.

NivekT:
Closing this because #786 has a serialization wrapper.

PR #437, opened 2022-05-23
Remove duplicate graph_utils

### LINKED ISSUE #435: `traverse` is different from PyTorch core and broken
Amongst other things, 0501f56323c09814c6c60a4bee4ffb6ca3d601d4 added the `torchdata/dataloader2/_graph_utils.py` module that implements `traverse`:

https://github.com/pytorch/data/blob/12cfaf8899b1337981cd4edf9deef127f925f1bd/torchdata/dataloader2/_graph_utils.py#L57

Unfortunately, this differs from the one in [PyTorch core](https://github.com/pytorch/pytorch/blob/88fca3be5924dd089235c72e651f3709e18f76b8/torch/utils/data/graph.py#L73). It is at least missing the patches that resolved #237. Trying to lint a datapipe that has circular references through the attributes (not an circular graph) fails again: https://github.com/pytorch/vision/runs/6551564325

Instead of re-implementing, could we maybe simply `from torch.utils.data.graph import traverse` to have a single source of truth?

ejguan:
Fixes #435

### DIFF
--- a/torchdata/dataloader2/_graph_utils.py
+++ b/torchdata/dataloader2/_graph_utils.py
@@ -5,61 +5,12 @@
 # LICENSE file in the root directory of this source tree.
 
 
-import io
-import pickle
-from typing import Dict, List
+from typing import Dict, Union
 
-from torchdata.datapipes.iter import IterDataPipe
+from torch.utils.data.graph import traverse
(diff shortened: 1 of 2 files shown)


NivekT (approved):
LGTM! Thanks!

PR #442, opened 2022-05-23
GDriveReader: handle Virus Scan Warning using confirm=t

### LINKED ISSUE #204: Getting `headers don't contain content-disposition` error when downloading data from Google Drive
### 🐛 Describe the bug

I seem to intermittently get the following error when executing the following piece of code to download torchtext datasets from Google drive.
```
>>> from torchtext.datasets.amazonreviewfull import AmazonReviewFull
>>> dataset = AmazonReviewFull(split="train")
>>> next(iter(dataset))
```
Stack trace
```
  File "/home/nayef211/.local/lib/python3.8/site-packages/torchdata/datapipes/iter/load/online.py", line 98, in __iter__
    yield _get_response_from_google_drive(url, timeout=self.timeout)
  File "/home/nayef211/.local/lib/python3.8/site-packages/torchdata/datapipes/iter/load/online.py", line 74, in _get_response_from_google_drive
    raise RuntimeError("Internal error: headers don't contain content-disposition.")
RuntimeError: Internal error: headers don't contain content-disposition.
```
According to @parmeet:
> This error existed in torchtext even before the migration, So I don't think any changes on torchdata would have triggered this. In my experience, I have seen this when the quota is exceeded. This is a transient error, though I never dig deeper into if there is a way to increase the quota limit or other alternatives to prevent this from happening.

The relevant discussion for this error can be found on this PR thread https://github.com/pytorch/text/pull/1594#issuecomment-1032824032.

cc @ejguan 

### Versions
```
PyTorch version: 1.11.0.dev20220111
Is debug build: False

OS: CentOS Stream 8 (x86_64)
GCC version: (GCC) 8.5.0 20210514 (Red Hat 8.5.0-3)
Clang version: Could not collect
CMake version: version 3.19.6
Libc version: glibc-2.28

Python version: 3.8.12 (default, Oct 12 2021, 13:49:34)  [GCC 7.5.0] (64-bit runtime)
Python platform: Linux-5.6.13-0_fbk18_hardened_6007_g4c10224f1437-x86_64-with-glibc2.17
Is CUDA available: False
CUDA runtime version: No CUDA
cuDNN version: No CUDA
HIP runtime version: N/A
MIOpen runtime version: N/A

[pip3] numpy==1.21.4
[pip3] pytorch-sphinx-theme==0.0.24
[pip3] torch==1.11.0.dev20220111
[pip3] torchdata==0.3.0a0+ec32ee4
[pip3] torchtext==0.12.0a0+b04ae99
[conda] blas                      1.0                         mkl
[conda] cpuonly                   2.0                           0    pytorch
[conda] mkl                       2021.4.0           h06a4308_640
[conda] numpy                     1.21.4                   pypi_0    pypi
[conda] pytorch                   1.11.0.dev20220111     py3.8_cpu_0    pytorch-nightly
[conda] pytorch-mutex             1.0                         cpu    pytorch
[conda] pytorch-sphinx-theme      0.0.24                   pypi_0    pypi
[conda] torchtext                 0.12.0a0+e691934          pypi_0    pypi
```

lolzballs:
Fixes #204

The Google Drive links used to download certain datasets takes us to a page like the following (the screenshot is from https://drive.google.com/uc?export=download&id=0Bz8a_Dbh9QhbUkVqNEszd0pHaFE):
image

It seems like the previous method for dealing with this via the `download_warning` cookie is no longer working. Instead it seems Google Drive now always uses "t" as the confirm_token. My patch sets the confirm_token if it detects the "Download anyway" string in the page.

### DIFF
--- a/torchdata/datapipes/iter/load/online.py
+++ b/torchdata/datapipes/iter/load/online.py
@@ -96,6 +96,11 @@ class HTTPReaderIterDataPipe(IterDataPipe[Tuple[str, StreamWrapper]]):
         return len(self.source_datapipe)
 
 
+def _extract_gdrive_api_response(content: str) -> Optional[str]:
+    match = re.search("<title>Google Drive - (?P<api_response>.+?)</title>", content)
+    return match["api_response"] if match is not None else None
+
+
 def _get_response_from_google_drive(url: str, *, timeout: Optional[float]) -> Tuple[str, StreamWrapper]:
     confirm_token = None
 
@@ -104,11 +109,16 @@ def _get_response_from_google_drive(url: str, *, timeout: Optional[float]) -> Tu
             response = session.get(url, stream=True)
         else:
             response = session.get(url, timeout=timeout, stream=True)
+
         for k, v in response.cookies.items():
             if k.startswith("download_warning"):
                 confirm_token = v
-        if confirm_token is None:
-            if "Quota exceeded" in str(response.content):
+                break
+        else:
+            api_response = _extract_gdrive_api_response(response.text)
+            if api_response == "Virus scan warning":
+                confirm_token = "t"
+            elif api_response == "Quota exceeded":
                 raise RuntimeError(f"Google drive link {url} is currently unavailable, because the quota was exceeded.")
 
         if confirm_token:


torchdata/datapipes/iter/load/online.py
```diff
@@ -110,6 +110,8 @@ def _get_response_from_google_drive(url: str, *, timeout: Optional[float]) -> Tu
         if confirm_token is None:
             if "Quota exceeded" in str(response.content):
                 raise RuntimeError(f"Google drive link {url} is currently unavailable, because the quota was exceeded.")
+            if "Download anyway" in str(response.content):
```
ejguan, line 113: Could you please take a reference from TorchVision to make this function more robust?
See:
- https://github.com/pytorch/vision/blob/ac56f52e6d6788c31726ee2d3a55afa442c6c3ec/torchvision/datasets/utils.py#L201-L214
- https://github.com/pytorch/vision/blob/ac56f52e6d6788c31726ee2d3a55afa442c6c3ec/torchvision/datasets/utils.py#L244-L261

ejguan (commented):
Thank you so much for contributing to TorchData. I left one comment below.

torchdata/datapipes/iter/load/online.py
lolzballs: I've revised the patch to more closely match the torchvision method. I've made sure it works for the "Virus scan warning" case but I'm not sure how to test the quota exceeded case. In any case, I believe the flow in my updated change should match what's in the torchvision files you sent.

torchdata/datapipes/iter/load/online.py
```diff
@@ -107,9 +122,11 @@ def _get_response_from_google_drive(url: str, *, timeout: Optional[float]) -> Tu
...
             if k.startswith("download_warning"):
                 confirm_token = v
-        if confirm_token is None:
-            if "Quota exceeded" in str(response.content):
-                raise RuntimeError(f"Google drive link {url} is currently unavailable, because the quota was exceeded.")
+                break
+        else:
+            api_response, content = _extract_gdrive_api_response(response)
+            if api_response == "Virus scan warning":
+                confirm_token = "t"
```
ejguan, line 129: Let's move `Quota exceeded` here.
```py
            if api_response == "Quota exceeded":
                raise RuntimeError(...)
```

torchdata/datapipes/iter/load/online.py
```diff
@@ -119,14 +136,18 @@ def _get_response_from_google_drive(url: str, *, timeout: Optional[float]) -> Tu
         else:
             response = session.get(url, timeout=timeout, stream=True)
 
+        api_response, content = _extract_gdrive_api_response(response)
+        if api_response == "Quota exceeded":
+            raise RuntimeError(f"Google drive link {url} is currently unavailable, because the quota was exceeded.")
+
```
ejguan, line 142: Let's remove this part.

torchdata/datapipes/iter/load/online.py
```diff
@@ -119,14 +136,18 @@ def _get_response_from_google_drive(url: str, *, timeout: Optional[float]) -> Tu
...
         if "content-disposition" not in response.headers:
             raise RuntimeError("Internal error: headers don't contain content-disposition.")
 
         filename = re.findall('filename="(.+)"', response.headers["content-disposition"])
         if filename is None:
             raise RuntimeError("Filename could not be autodetected")
 
-    return filename[0], StreamWrapper(response.raw)
+    return filename[0], StreamWrapper(content)
```
ejguan, line 150: Please keep it as `response.raw` to make the response in a streaming manner.

lolzballs: I've replaced it back with `response.raw`. And I also made `_extract_gdrive_api_response` operate on the initial `response.text`, since it seems we expect the initial response to be an HTML page instead of the actual data anyways.

torchdata/datapipes/iter/load/online.py
```diff
@@ -4,6 +4,7 @@
 # This source code is licensed under the BSD-style license found in the
 # LICENSE file in the root directory of this source tree.
 
+import itertools
```
ejguan, line 7: nit: `itertools` is not needed anymore.

ejguan (approved):
Overall LGTM with one nit comment. Thank you

torchdata/datapipes/iter/load/online.py
lolzballs: Thanks! Removed it.

torchdata/datapipes/iter/load/online.py
```diff
@@ -95,6 +95,9 @@ def __iter__(self) -> Iterator[Tuple[str, StreamWrapper]]:
     def __len__(self) -> int:
         return len(self.source_datapipe)
 
+def _extract_gdrive_api_response(content: str) -> Optional[str]:
+    match = re.search("<title>Google Drive - (?P<api_response>.+?)</title>", content)
+    return match["api_response"] if match is not None else None
```
ejguan, line 100: Could you please fix the lint Error? Basically, we need two empty lines between functions.

ejguan (commented):
I have seen a few more failures on CI. Domain tests and test with AWSSDK are not related to your PR. It seems an Error pop up in https://github.com/pytorch/data/runs/6579875479?check_suite_focus=true#step:12:146
I will take a look at this Error shortly.

ejguan (commented):
Could you please rebase onto main branch again?

lolzballs:
Rebased.

About those failing GDrive unit tests, I tried on main and it seems it's not working there either (since it doesn't use `confirm=t`). But after trying many times locally it feels like Google Drive is locking me out and returning a 403 which happens in a browser also (unless I'm logged into Google).

ejguan:
> About those failing GDrive unit tests, I tried on main and it seems it's not working there either (since it doesn't use `confirm=t`). But after trying many times locally it feels like Google Drive is locking me out and returning a 403 which happens in a browser also (unless I'm logged into Google).

Lol. If that's the case, we should get the Error of `Quota exceeded`. I have tried to run the tests on main branch and I do get the quota Error, though.

lolzballs:
> If that's the case, we should get the Error of Quota exceeded. I have tried to run the tests on main branch and I do get the quota Error, though.

It seems it's a little different though. In Firefox I can see the Virus scan warning page, but when I click the download anyway button it takes me to a blank page, which in the dev tools I can see is giving a 403.

ejguan:
> > If that's the case, we should get the Error of Quota exceeded. I have tried to run the tests on main branch and I do get the quota Error, though.
> 
> It seems it's a little different though. In Firefox I can see the Virus scan warning page, but when I click the download anyway button it takes me to a blank page, which in the dev tools I can see is giving a 403.

Interesting. I can still download from chrome. Weird. Can this be the problem of some firewall settings?

NivekT (approved):

ejguan:
Based on the result on CI, I think this PR looks fine. I will start to land it.

PR #467, opened 2022-05-25
Add release notes script

ejguan:
This PR is adapted from PyTorch Core to help creating a list of commits with corresponding PR links.
https://github.com/pytorch/pytorch/tree/master/scripts/release_notes

It should help us to aggregate and categorize commits in the future.

I will create a release note runbook for us to follow. The main steps to run this script:
```
python commitlist.py --create_new tags/v0.3.0 <commit_hash_on_new_release>  # Create a csv file to list all commits
python commitlist.py --update_to <commit_hash_on_new_release>  # When we have new cherry-picked PRs landed onto release branch
python commitlist.py --export_markdown  # Based on the categorization, create separate MD files per category
```

### DIFF
--- a/scripts/release_notes/commitlist.py
+++ b/scripts/release_notes/commitlist.py
@@ -10,8 +10,12 @@
 import argparse
 import csv
 import os
+import pprint
+import re
 
-from common import get_features, run
+from collections import defaultdict
+
+from common import CommitDataCache, get_features, run, topics
 
 
 class Commit:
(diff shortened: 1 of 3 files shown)


scripts/release_notes/commitlist.py
```diff
@@ -0,0 +1,278 @@
...
+                category = "dataloader2"
+                break
+            if CommitList.keywordInFile(
+                file,
+                [
+                    "torchdata/datapipes",
+                ],
+            ):
+                category = "datapipe"
+                break
```
ejguan, line 129: Here are the main changes to adapt the structure of TorchData.
Please let me know if you have any recommendation on the categorization.

NivekT (approved):

scripts/release_notes/commitlist.py
NivekT: I think we can also consider mandatory labelling of every PR explicitly (like PyTorch Core does). It will be more accurate that, though take may be 30 sec extra per PR

ejguan: Yeah. That's the reason I add a few labels like PyTorch Core. And, this script does respect such topic label.
For mandatory labelling, I don't have time to do it right now. Treat it as a follow-up task.

PR #471, opened 2022-05-26
Fix TF Record test

ejguan:
Fixes #470

### Changes

- Add `test/requirements.txt` to specify test dependencies
- Specify `protobuf` version smaller than 4
- Remove redundant `numpy` requirements
- Fix lint error on main branch

PR #472, opened 2022-05-26
Add conda token for test channel

ejguan:
We have one missing token for conda RC channel. See: https://github.com/pytorch/data/actions/runs/2390820859

After devinfra team provides this token, we need to update the test_release workflow correspondingly.


ejguan (commented):
Lint and CI errors are fixed in #471

NivekT (approved):

PR #479, opened 2022-05-31
[DataPipe] Adding a 's' to the functional names of open/list DataPipes

NivekT:
* __->__ #479

Differential Revision: [D36785643](https://our.internmc.facebook.com/intern/diff/D36785643)

### DIFF
--- a/torchdata/datapipes/iter/load/fsspec.py
+++ b/torchdata/datapipes/iter/load/fsspec.py
@@ -101,11 +101,11 @@ class FSSpecFileListerIterDataPipe(IterDataPipe[str]):
                             yield abs_path
 
 
-@functional_datapipe("open_file_by_fsspec")
+@functional_datapipe("open_files_by_fsspec")
 class FSSpecFileOpenerIterDataPipe(IterDataPipe[Tuple[str, StreamWrapper]]):
     r"""
     Opens files from input datapipe which contains `fsspec` paths and yields a tuple of
-    pathname and opened file stream (functional name: ``open_file_by_fsspec``).
+    pathname and opened file stream (functional name: ``open_files_by_fsspec``).
 
     Args:
         source_datapipe: Iterable DataPipe that provides the pathnames or URLs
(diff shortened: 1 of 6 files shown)


torchdata/datapipes/iter/load/fsspec.py
```diff
@@ -101,11 +101,11 @@ def __iter__(self) -> Iterator[str]:
                             yield abs_path
 
 
-@functional_datapipe("open_file_by_fsspec")
+@functional_datapipe("open_files_by_fsspec")
```
ejguan, line 104: I think this one is BC breaking from 0.3.0. Should we add `open_file_by_fsspec` to this class as well?

Call `IterDataPipe. register_datapipe_as_function("open_file_by_fsspec", FSSpecFileOpenerIterDataPipe)` at the end of this file.

NivekT: Yea, we can do that.

NivekT:
> Thank you for adding this PR. Could you please also fix tests?

I don't think there are any existing use cases in tests but I will add some.

torchdata/datapipes/iter/load/fsspec.py
```diff
@@ -133,6 +133,10 @@ def __len__(self) -> int:
         return len(self.source_datapipe)
 
 
+# Register for functional API for backward compatibility
+IterDataPipe.register_datapipe_as_function("open_file_by_fsspec", FSSpecFileOpenerIterDataPipe)
```
ejguan, line 137: As a follow-up task, add deprecation warning for the functional API.

NivekT: It seems like we will need to change the registration function and `IterDataPipe` class in PyTorch core to do that.  One possibility is [described here](https://github.com/pytorch/data/issues/480#issuecomment-1142500688).

Do you think we should land this PR as it is or make the changes to core first?

ejguan: Let's land this PR first and change the core afterwards.

ejguan (approved):

torchdata/datapipes/iter/load/fsspec.py
ejguan: And, could you please update the deprecation table as well?

PR #481, opened 2022-05-31
Add Token for Conda RC Release

ejguan:
Follow up PR for #472
The Test token is required to be added to `env` for the uploading step in workflow


NivekT (approved):
LGTM! Thanks!

PR #484, opened 2022-05-31
Adding DataLoader2 Adapter API and shuffle adapter

VitalyFedyunin:
Stack from [ghstack](https://github.com/ezyang/ghstack) (oldest at bottom):
* #485
* __->__ #484

Differential Revision: [D36866291](https://our.internmc.facebook.com/intern/diff/D36866291)

PR #485, opened 2022-05-31
[WIP] pin_memory DataPipe and adapter

VitalyFedyunin:
* __->__ #485
* #484

Differential Revision: [D36866290](https://our.internmc.facebook.com/intern/diff/D36866290)

### DIFF
--- a/test/test_adapter.py
+++ b/test/test_adapter.py
@@ -8,6 +8,8 @@ import warnings
 
 from unittest import TestCase
 
+import torch
+
 from torchdata.dataloader2 import DataLoader2, MultiProcessingReadingService, ReadingServiceInterface
 from torchdata.dataloader2.adapter import shuffle
 from torchdata.datapipes.iter import IterableWrapper, IterDataPipe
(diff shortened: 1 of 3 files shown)


VitalyFedyunin:
@VitalyFedyunin has imported this pull request.  If you are a Facebook employee, you can view this diff [on Phabricator](https://www.internalfb.com/diff/D36866290).

VitalyFedyunin:
@VitalyFedyunin has imported this pull request.  If you are a Facebook employee, you can view this diff [on Phabricator](https://www.internalfb.com/diff/D36866290).

VitalyFedyunin:
Looks like pin_memory should be a parameter of ReadingService, otherwise 'pin' gets lost when tensors are moved from child processes to the main training loop.

CC @ejguan

ejguan:
> Looks like pin_memory should be a parameter of ReadingService, otherwise 'pin' gets lost when tensors are moved from child processes to the main training loop.
> 
> CC @ejguan

Do we want to have special DataPipes that can only be iterated on main process (cannot moved to child process)? If so, we might still keep it as an Adapter and use a special DataPipe. FWIW, we might need such feature for DataPipe operations when Tensor is pinned or sent to cuda.

VitalyFedyunin:
Makes total sense to me, I will update MPReadingService to respect such DataPipes as part of this PR.

PR #490, opened 2022-06-01
Add support for hf `datasets` reader

msaroufim:
In working on #422 I realized we needed an easier way to run benchmarks on larger datasets that may not be available on domain libraries. This PR is a prerequisite to any sort of scaling benchmarks and is generally a useful reader for the community.

https://huggingface.co/docs/datasets/how_to has about 10,000 datasets we can leverage out of the box - in particular mc4 is one we need to prove out large scale benchmarks for text

See test and docstring for usage instructions

### Changes

- Added a new `HuggingFaceHubReaderIterDataPipe` so we can load a large number of datasets for performance benchmarks
- Added a test which is skipped if `datasets` library does not exist 
- pytest passes
- Got rid of `StreamWrapper`
- Is there any documentation update I should make?


torchdata/datapipes/iter/load/huggingface.py
```diff
@@ -0,0 +1,60 @@
...
+class HuggingFaceHubReaderDataPipe(IterDataPipe[Tuple[str, StreamWrapper]]):
+    r"""
+    Takes in dataset names and returns an Iterable HuggingFace dataset
+    Args:
+        source_datapipe: a DataPipe that contains dataset names which will be accepted by the HuggingFace datasets library
+        revision: the specific dataset version
+        data_files: Optional dict to set custom train/test/validation split
+    Example:
+        >>> from torchdata.datapipes.iter import IterableWrapper, HuggingFaceHubReaderDataPipe
+        >>> huggingface_reader_dp = HuggingFaceHubReaderDataPipe(IterableWrapper([dataset]), revision=main)
```
VitalyFedyunin, line 37: nit: If `dataset` is string here, can you please just put any of existing datasets here

torchdata/datapipes/iter/load/huggingface.py
```diff
@@ -0,0 +1,60 @@
...
+from typing import Any, Callable, Dict, Iterator, List, Optional, Tuple, Union
+
+from torch.utils.data.datapipes.utils.common import match_masks
+
+from torchdata.datapipes import functional_datapipe
+from torchdata.datapipes.iter import IterDataPipe
+from torchdata.datapipes.utils import StreamWrapper
+
+try:
+    import datasets
```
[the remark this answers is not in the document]
ejguan: I am just kidding. No need to do so.

test/test_huggingface_datasets.py
```diff
@@ -0,0 +1,56 @@
...
+            self.temp_dir.cleanup()
+            self.temp_sub_dir_2.cleanup()
+            self.temp_dir_2.cleanup()
+        except Exception as e:
+            warnings.warn(f"HuggingFace datasets was not able to cleanup temp dir due to {e}")
+
+
+    @skipIfNoDatasets
+    def test_huggingface_hubreader(self):
+        datapipe = HuggingFaceHubReader(dataset="lhoestq/demo1", revision="main")
```
VitalyFedyunin, line 50: Is it even working? As per API, you suppose to pass DataPipe with dataset names here. Also, it seems to be an unnecessary and rare scenario to load more than one dataset at once. Please consider dropping `source_datapipe` and replacing it with a singular `dataset` argument.

torchdata/datapipes/iter/load/huggingface.py
```diff
@@ -0,0 +1,59 @@
...
+        >>> path
+        Add test result here
+        >>> line
+        Add test result here b'BSD 3-Clause License'
+    """
+
+    source_datapipe: IterDataPipe[str]
+
+    def __init__(self, dataset: str, *, split : str = "train", revision : Optional[str] = None, data_files : Optional[Dict[str,str]] = None) -> None:
+        self.dataset = dataset
```
ejguan, line 50: Could we raise Error here when `datasets` is not available?
See ref: https://github.com/pytorch/data/blob/cd38927904836f6f67ce33bfaee094fff4078402/torchdata/datapipes/iter/load/iopath.py#L184-L189

torchdata/datapipes/iter/load/huggingface.py
```diff
@@ -0,0 +1,59 @@
...
+try:
+    import datasets
+except ImportError:
+    datasets = None
+
+def _get_response_from_huggingface_hub(dataset, split, revision, data_files) -> Tuple[Any, StreamWrapper]:
+    dataset = datasets.load_dataset(dataset, split, revision, data_files)
+    return dataset[0], StreamWrapper(dataset)
+
+@functional_datapipe("read_from_huggingface_hub")
```
ejguan, line 27: As the first argument is a string, we can't really support `functional_datapipe` in this case.

ejguan (commented):
Could you please also add `datasets` to https://github.com/pytorch/data/blob/main/test/requirements.txt to make sure the test case is carried out?

torchdata/datapipes/iter/load/huggingface.py
ejguan: If the first input is not `IterDataPipe`, there won't be the option for users to call `datapipe. read_from_huggingface_hub()`

So, adding `functional_datapipe` here would cause a bug.

torchdata/datapipes/iter/load/huggingface.py
```diff
@@ -0,0 +1,65 @@
...
+from torchdata.datapipes.iter import IterDataPipe
+from torchdata.datapipes.utils import StreamWrapper
+
+try:
+    import datasets
+except ImportError:
+    datasets = None
+
+def _get_response_from_huggingface_hub(dataset, split, revision, data_files) -> Tuple[Any, StreamWrapper]:
+    dataset = datasets.load_dataset(dataset, split, revision, data_files)
```
NivekT, line 24: n00b question:
1. Their API has an option `streaming=True` for an IterableDataset, do we want that?
2. How do people typically do something like `for batch in dataset`? In that case, do we want `yield from` instead in the DataPipe?

torchdata/datapipes/iter/load/huggingface.py
```diff
@@ -0,0 +1,65 @@
...
+        >>> reader_dp = huggingface_reader_dp
+        >>> it = iter(reader_dp)
+        >>> path, line = next(it)
+        >>> path
+        Add test result here
+        >>> line
+        Add test result here b'BSD 3-Clause License'
+    """
+
+    source_datapipe: IterDataPipe[str]
```
NivekT, line 46: Am I reading incorrectly or is this variable only used by `__len__` and not elsewhere?

msaroufim: You're right this is an artifact will fix

torchdata/datapipes/iter/load/huggingface.py
msaroufim: that's a good callout, I think using `streaming=True` by default is a good idea and users should be able to overwrite that behavior

A `yield` also sounds good to me, I was using this `    return dataset[0], StreamWrapper(dataset)` because that's what the other examples seemed to be using

torchdata/datapipes/iter/load/huggingface.py
```diff
@@ -0,0 +1,66 @@
...
+from torchdata.datapipes import functional_datapipe
+from torchdata.datapipes.iter import IterDataPipe
+from torchdata.datapipes.utils import StreamWrapper
+
+try:
+    import datasets
+except ImportError:
+    datasets = None
+
+def _get_response_from_huggingface_hub(dataset, split, revision, streaming, data_files) -> Tuple[Any, StreamWrapper]:
```
ninginthecloud, line 23: The types of function arguments are missing, let's update them. 
Also, I'm curious to know the reason that we choose these 5 arguments, the original HF load_dataset API has a lot of [arguments](https://github.com/huggingface/datasets/blob/4844b9c65e042598e0bd9f2e5bd5b5f784c618bc/src/datasets/load.py#L1590-L1606)

msaroufim: These seemed the most useful to me, I can add `**kwargs` so we can support all parameters

torchdata/datapipes/iter/load/huggingface.py
```diff
@@ -0,0 +1,66 @@
...
+        self.split = split
+        self.revision = revision
+        self.streaming = streaming
+        self.data_files = data_files
+
+    def __iter__(self) -> Iterator[Tuple[str, StreamWrapper]]:
+        yield _get_response_from_huggingface_hub(dataset=self.dataset, split=self.split, revision=self.revision, streaming=self.streaming, data_files=self.data_files)
+
+    def __len__(self) -> int:
+        return len([self.dataset])
```
ninginthecloud, line 66: will `len(dp)` always be 1 in this case considering we only input a single `dataset`?

msaroufim: yes `len(dp)` isn't particularly useful unfortunately here because the first input to `__init__` is a string

torchdata/datapipes/iter/load/huggingface.py
```diff
@@ -0,0 +1,66 @@
...
+from torchdata.datapipes.utils import StreamWrapper
+
+try:
+    import datasets
+except ImportError:
+    datasets = None
+
+def _get_response_from_huggingface_hub(dataset, split, revision, streaming, data_files) -> Tuple[Any, StreamWrapper]:
+    dataset = datasets.load_dataset(dataset, split, revision, streaming, data_files)
+    return dataset[0], StreamWrapper(dataset)
```
ninginthecloud, line 25: this is not a comment but more like a learning question for me:
What is StreamWrapper used for? dataset from `load_dataset` is already an iterable, could we return dataset or iter(dataset) directly?

torchdata/datapipes/iter/load/huggingface.py
ninginthecloud: how about returning the length of the HF [dataset](https://github.com/huggingface/datasets/blob/2f9a6a5bf3e25944b3620c870896dce0c8c0d52f/src/datasets/arrow_dataset.py#L1749-L1764)? However, it has different meaning compared to the one you implemented here. What do you think, @ejguan and @NivekT ~

ninginthecloud:
@msaroufim Thank you so much for working this PR! As you mentioned, there are many existing datasets in HF, this pipe will be quite handy for other researchers as well~

torchdata/datapipes/iter/load/huggingface.py
msaroufim: We seem to use it a bunch in the torchdata codebase to create iterable objects like so https://github.com/pytorch/data/blob/main/torchdata/datapipes/iter/load/online.py#L46

That said @NivekT had similar feedback to yours so maybe we can get rid of it

NivekT: We typically use `StreamWrapper` to wrap around streams that need to be closed during deletion.

https://github.com/pytorch/pytorch/blob/master/torch/utils/data/datapipes/utils/common.py#L172-L176

I am not familiar with HF dataset to be certain, my guess is that `StreamWrapper` isn't needed.

torchdata/datapipes/iter/load/huggingface.py
ejguan: For the current implementation, I think the length as 1 makes more sense since only one element is yielded from it. You can treat it as an element with a large batch size = dataset size.

An alternative option is doing the similar thing as the [`TracingArrowDataPipe`](https://fburl.com/code/vpcqnqdp). Basically, we load the HF dataset as an internal object, and delegate operation like `map`, `shuffle`, etc. from datapipe to the internal HF dataset. And, when users iterate over the DataPipe instance, we will return the iterator of the internal HF dataset. In this case, we can return the length of internal HF Dataset.

cc: @VitalyFedyunin Not sure if this is the best practice for HF Dataset.

torchdata/datapipes/iter/load/huggingface.py
msaroufim: Ok I don't think we need kwargs I added all parameters I could find listed on the main documentation https://huggingface.co/docs/datasets/loading

torchdata/datapipes/iter/load/huggingface.py
```diff
@@ -0,0 +1,80 @@
...
+                "to install the package"
+            )
+
+        self.dataset = dataset
+        self.split = split
+        self.revision = revision
+        self.streaming = streaming
+        self.data_files = data_files
+
+    def __iter__(self) -> Iterator[Tuple[str, StreamWrapper]]:
```
ninginthecloud, line 70: The return type does not match with return type of `_get_response_from_huggingface_hub()`.

ninginthecloud (approved):
@msaroufim Hi, Mark, the PR looks good to me now~ there's one minor type fix, you can see the comment for details.

torchdata/datapipes/iter/load/huggingface.py
NivekT: There are 16. @ninginthecloud Do you think having those 5 should be sufficient?

https://huggingface.co/docs/datasets/package_reference/loading_methods#datasets.load_dataset

NivekT (approved):
Overall, LGTM! Can you check the CI's linting error?
Also, is this serializable? Can you add it to `test_serialization`? I think it should work.

VitalyFedyunin (approved):
LGTM but please drop length and make in unavailable (see `filter` as example)

NivekT:
> Another TODO to self is `test_serialization.py` is failing because it's trying to import but can't find `datasets` - need to fix this

Let's add an entry here and we should be good to go:
https://github.com/pytorch/data/blob/7b5e7d92d57d4505b255392ca002f2500b860679/test/test_serialization.py#L75-L85

NivekT:
@ejguan Should we cherry-pick this or leave it for the next release?

ejguan:
> @ejguan Should we cherry-pick this or leave it for the next release?

I am fine with cherry-picking it to release.

PR #491, opened 2022-06-01
Con_disk_cache now handles situation when lock files remains between failed runs

VitalyFedyunin:
* __->__ #491

### DIFF
--- a/torchdata/datapipes/iter/util/cacheholder.py
+++ b/torchdata/datapipes/iter/util/cacheholder.py
@@ -242,10 +242,11 @@ class OnDiskCacheHolderIterDataPipe(IterDataPipe):
                         promise_fh.flush()
             else:
                 promise_filepath = _promise_filename(filepath)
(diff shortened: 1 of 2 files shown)


torchdata/datapipes/iter/util/cacheholder.py
```diff
@@ -223,16 +232,20 @@ def _cache_check_fn(data, filepath_fn, hash_dict, hash_type, extra_check_fn):
...
                     if not file_exists:
                         result = False
                         promise_fh.seek(0)
-                        promise_fh.write("[dataloader session uid]")
+                        promise_fh.write(str(uuid))
                         promise_fh.truncate()
                         promise_fh.flush()
+            else:
+                promise_filepath = _promise_filename(filepath)
+                with portalocker.Lock(promise_filepath, "r", flags=portalocker.LockFlags.SHARED) as promise_fh:
```
ejguan, line 245: IIUC, there can be a case that `promise` file has been removed when the promise if fulfilled.

PR #495, opened 2022-06-03
Improve fsspec DataPipe to accept extra keyword arguments

### LINKED ISSUE #494: fsspec requires to delegate `kwargs` to FileSystem to list/load file correctly
### 🐛 Describe the bug

Based on the current [implementation](https://github.com/pytorch/data/blob/6f1f81db751d28e675728ab36007be05db0614af/torchdata/datapipes/iter/load/fsspec.py#L69), the following code will raise `ConnectTimeoutError`.
```py
>>> import fsspec

>>> url = "s3://ai2-public-datasets/charades"
>>> fs, path = fsspec.core.url_to_fs(root)
>>> fs.ls(path)
botocore.exceptions.ConnectTimeoutError
```

However, if we provide extra arguments to `url_to_fs`, we should be able to successfully execute the following code:
 ```py
>>> import fsspec

>>> url = "s3://ai2-public-datasets/charades"
>>> fs, path = fsspec.core.url_to_fs(root, anon=True)  # The main difference here
>>> fs.ls(path)
[...]
```

It should be easily fixed by delegating extra key-word argument to all `FSSpecDataPipe`s.

### Versions

main branch

### PR DESCRIPTION
**Pull request by ejguan**
Fixes #494

### Changes

- Add `kwargs` to all `DataPipe`s using `fsspec`
- Fix a bug in `FSSpecFileLister` to prevent joining paths from the root path if the root path has already presented in the paths.
- Add/Fix more tests

I need this PR to test the performance between `fsspec` and native `s3`

### DIFF
--- a/torchdata/datapipes/iter/load/fsspec.py
+++ b/torchdata/datapipes/iter/load/fsspec.py
@@ -42,6 +42,8 @@ class FSSpecFileListerIterDataPipe(IterDataPipe[str]):
     Args:
         root: The root `fsspec` path directory or list of path directories to list files from
         masks: Unix style filter string or string list for filtering file name(s)
+        kwargs: Extra options that make sense to a particular storage connection,
+            e.g. host, port, username, password, etc.
 
     Example:
         >>> from torchdata.datapipes.iter import FSSpecFileLister
@@ -52,6 +54,7 @@ class FSSpecFileListerIterDataPipe(IterDataPipe[str]):
         self,
         root: Union[str, Sequence[str], IterDataPipe],
         masks: Union[str, List[str]] = "",
+        **kwargs,
     ) -> None:
         _assert_fsspec()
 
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

File: torchdata/datapipes/iter/load/fsspec.py [hunk 1]
```diff
@@ -87,8 +91,10 @@ def __iter__(self) -> Iterator[str]:
                     else:
                         if is_local:
                             abs_path = os.path.join(path, file_name)
-                        else:
+                        elif not file_name.startswith(path):
                             abs_path = posixpath.join(path, file_name)
+                        else:
+                            abs_path = file_name
```
NivekT, line 97: Just to confirm - this never happens if `is_local == True`?

NivekT (approved):

torchdata/datapipes/iter/load/fsspec.py
ejguan: Good question. I can't guarantee it since I am not aware of all use cases.Based on my experience on `fsspec`, I only encounter the problem when the input is `s3` url.
For local files, I would trust the test here https://github.com/pytorch/data/blob/main/test/test_fsspec.py#L56-L65. As long as it doesn't break, I think

PR #503, opened 2022-06-08
Always upload TorchData nightly release

ejguan:
Based on the offline discussion with Vitaly, we want to upload TorchData nightly releases even if tests are failing.

It's required by downstream library when PyTorch Core nightly updated but TorchData nightly not updated (still depends on the previous PT Core nightly)

Validated in: https://github.com/pytorch/data/actions/runs/2463948389
Even the tests are failing, the nightly releases are still uploaded.

### DIFF
--- a/packaging/torchdata/meta.yaml
+++ b/packaging/torchdata/meta.yaml
@@ -47,8 +47,6 @@ test:
     # The following packages are not on the default conda channel
     # - iopath
     # - rarfile
-  commands:
-    - pytest --no-header -v test --ignore=test/test_period.py --ignore=test/test_text_examples.py --ignore=test/test_audio_examples.py
 
 about:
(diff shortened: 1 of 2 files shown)


packaging/torchdata/meta.yaml
```diff
@@ -47,8 +47,6 @@ test:
     # The following packages are not on the default conda channel
     # - iopath
     # - rarfile
-  commands:
-    - pytest --no-header -v test --ignore=test/test_period.py --ignore=test/test_text_examples.py --ignore=test/test_audio_examples.py
```
ejguan, line 51: Moved test out of the step of conda build.

NivekT (approved):
This makes sense to me. Thanks!

Issue #293, opened 2022-03-11
[RFC] Make the TorchData Library Standalone from PyTorch Core Library


NivekT:
### 🚀 The feature

**Note that this is a request for comment; currently, there is no plan to make TorchData a standalone library.** We would like to solicit feedback from the community.

Proposal: Make the TorchData library standalone with little to no dependency on the PyTorch Core library (i.e. [`torch`](https://github.com/pytorch/pytorch)).

### Motivation, pitch

An argument for a standalone library is that, it will allow users to use all the data loading functionalities in this library without installing/using PyTorch. Datasets implemented using TorchData may become usable by other frameworks.

An argument against this change is - in order to certain DataLoader functionalities backward compatible with DataPipes, the `torch` library may need to become dependent on TorchData instead.

The list of arguments here is not comprehensive, feel free to leave a comment about potential use cases and how they will be impacted. 

### Alternatives

Leave the library as it is with dependency on `torch`.

josiahls:
I would really like this on my end. I'm writing a RL framework and part of the goal is to pickle a model and deploy it on a robotic system. I may or may not want torch actually installed there (maybe im using onnx?), but still have a pipeline for generically executing code.

ejguan:
Another feedback from our a downstream library (ray):
They previously provided an interface for users to transform `ray.Dataset` into `IterableDataset`. As we plan to collaborate to switch it to `IterDataPipe`, they would have a hard requirement to use `torch=1.11.0`. In order to provide a flexible requirements, they need to figure out how to fall back to use `IterableDataset` if `torch` doesn't meet the condition and `torchdata` doesn't exist in the environment.

If we make torchdata out-of-tree where DataLoader2 and related utility all go into torchdata, we potentially could set a loose requirements over the version of `torch`.

abhi-mosaic:
I'm working on a downstream library (composer) that has the exact same issue -- we would love to allow users to build datasets with `torchdata` without hard requirements on the `torch` version.

ejguan:
I almost forget one of the most important benefits by making TorchData standalone. If we have some changes landed into PyTorch Core that, a dependent changes in TorchData have to wait until the nightly release is updated for PyTorch Core.
On the other hand, if we have BC breaking changes in PyTorch Core, we have to open another PR in TorchData to incorporate such BC breaking then make sure the nightly release shipped for both repos. Otherwise, there is a risk that our downstream libraries like TorchVision/TorchText have red CI for a while.

[...]

PR #509, opened 2022-06-10
Validation of  `fn` on `input_col` for Datapipes

### LINKED ISSUE #362: Add validation between `input_col` and `fn` for Callable DataPipes
### 🚀 The feature

Using `inspect` to validate the `input_col` matching `fn` input for `Mapper`, `Flatmapper` and other DataPipe using both `input_col` and `fn` as the arguments.
- [`Mapper`](https://github.com/pytorch/pytorch/blob/6dc71461e11122494a303214a72b12adb353dc4d/torch/utils/data/datapipes/iter/callable.py#L55-L57)
- [`Filter`](https://github.com/pytorch/pytorch/blob/ec591087fb95525b430f2bbcdd023567e2fb1cc9/torch/utils/data/datapipes/iter/selecting.py#L13-L47)
- [`Flatmap`](https://github.com/pytorch/data/blob/732307a1dcea8686bff9a855bcd67c32d2087f8e/torchdata/datapipes/iter/transform/callable.py#L93-L123)
- [`BatchMap`](https://github.com/pytorch/data/blob/732307a1dcea8686bff9a855bcd67c32d2087f8e/torchdata/datapipes/iter/transform/callable.py#L15-L59)
- etc

### Motivation, pitch

This would help users to identify the wrong arguments at the construction time for such `IterDataPipe` rather than at iteration time.

### Alternatives

_No response_

### Additional context

_No response_

xiurobert:
Fixes #362 

### Changes

- Document input_col behavior on `FlatMapperIterDataPipe`
- Checks the arguments of mapping function against `input_col` in `FlatMapperIterDataPipe`

### DIFF
--- a/torchdata/datapipes/iter/transform/callable.py
+++ b/torchdata/datapipes/iter/transform/callable.py
@@ -96,7 +96,6 @@ class BatchMapperIterDataPipe(IterDataPipe[T_co]):
         assert batch_size > 0, "Batch size is required to be larger than 0!"
         self.batch_size = batch_size
         self.input_col = input_col
-        ensure_fn_works(fn, input_col)
 
     def _apply_fn(self, batch):
         if self.input_col is None:
--- a/torchdata/datapipes/iter/transform/callable.py
+++ b/torchdata/datapipes/iter/transform/callable.py
@@ -132,11 +132,20 @@ class FlatMapperIterDataPipe(IterDataPipe[T_co]):
 
     def _ensure_fn_works_on_input(self, fn: Callable):
         sig = inspect.signature(fn)
-        if len(sig.parameters) != self.input_col + 1:
-            raise TypeError(
-                f"The function {fn.__name__} takes {len(sig.parameters)} arguments, "
-                f"but {self.input_col + 1} are required."
-            )
+
+        if isinstance(self.input_col, int):
+            if len(sig.parameters) != self.input_col:
+                raise TypeError(
+                    f"The function {fn.__name__} takes {len(sig.parameters)} "
+                    f"arguments, "
+                    f"but {self.input_col} are required."
+                )
+        elif isinstance(self.input_col, (list, tuple)):
+            if len(sig.parameters) != len(self.input_col):
+                raise TypeError(
+                    f"The function {fn.__name__} takes {len(sig.parameters)} arguments, "
+                    f"but {len(self.input_col)} are required."
+                )
 
     def _apply_fn(self, data):
         if self.input_col is None:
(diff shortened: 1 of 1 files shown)


xiurobert:
Oh, hold on. This will break test_flatmap_iterdatapipe

xiurobert:
This breaks `test_map_batches_iterdatapipe`, specifically 683 (since there is only 1 argument)

xiurobert:
@ejguan, should I modify `fn_2_cols` or is that not desired?

ejguan:
Could you please open a PR in PyTorch core by creating the validation function first?
- [`Mapper`](https://github.com/pytorch/pytorch/blob/77b6885a225db0e4dfe331d3e98a5a456c672c2c/torch/utils/data/datapipes/iter/callable.py#L17)
- [`Filter`](https://github.com/pytorch/pytorch/blob/dd620c45752ab746bdd6bc47e8a2425a2f79a702/torch/utils/data/datapipes/iter/selecting.py#L14)
Then, TorchData can import such validation function from PyTorch Core and reuse it.

xiurobert:
Oh, good idea. Any specific file I should put it in?

ejguan:
You can put it in https://github.com/pytorch/pytorch/blob/master/torch/utils/data/datapipes/utils/common.py

xiurobert:
Are the CI failures normal?

torchdata/datapipes/iter/transform/callable.py
```diff
@@ -65,6 +96,7 @@ def __init__(
         assert batch_size > 0, "Batch size is required to be larger than 0!"
         self.batch_size = batch_size
         self.input_col = input_col
+        ensure_fn_works(fn, input_col)
```
ejguan, line 99: Let's remove this part for now. I need to double check with the behavior of `map_batches`

ejguan (commented):
The Error is related to `map_batches`. Let's remove the check for `map_batches` for now.

The logic of `ensure_fn_works` seems reasonable. Like I said before, we can move this function to PyTorch core to prevent duplicate implementations.

torchdata/datapipes/iter/transform/callable.py
xiurobert: I will take out both the function definition and the checks then until core merges my checks

xiurobert:
If #293 goes through this function may need to be copied back over here

xiurobert:
@ejguan Okay to continue working on this now that core has merged?

ejguan:
> @ejguan Okay to continue working on this now that core has merged?

Yeah. You can do that. But, you might wait until my PR landed as your PR breaks a few tests for TorchVision and I fixed them in my PR.

xiurobert:
Let me close and reopen this now that the validator function has been merged.

PR #510, opened 2022-06-10
[releng] Use PyTorch RC binary when testing on release branch

ejguan:
Currently the CI on release branch uses nightly PyTorch release. This PR changes it to use RC PyTorch release when the target branch of PR is the release branch or a commit is pushed to release branch.

See the CI result of the PR for release branch: https://github.com/pytorch/data/runs/6837219249?check_suite_focus=true#step:8:31

PR #514, opened 2022-06-10
Re-land Use PyTorch RC release on release branch (#510)

ejguan:
Currently the CI tests on release branch use the nightly PyTorch release. This PR changes tests to use RC PyTorch release when the target branch of PR is the release branch or a commit is pushed to release branch.

Reverted PR: #510

See the CI result for when the PR is landed (nightly channel): https://github.com/pytorch/data/runs/6867662390?check_suite_focus=true#step:8:3
See the CI result of the PR for release branch (test channel): https://github.com/pytorch/data/runs/6867788289?check_suite_focus=true#step:8:3

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -4,6 +4,7 @@ on:
     branches:
       - main
       - release/*
+    tags:
   pull_request:
     types: [opened, synchronize, reopened, labeled]
     branches:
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yml
```diff
@@ -30,6 +31,16 @@ jobs:
           - 3.8
           - 3.9
     steps:
+      - name: Get PyTorch Channel
+        shell: bash
+        run: |
+          if [[ "${{ github.base_ref }}" == release/* ]] || [[ "${{ github.ref }}" == refs/heads/release/* ]] || [[ "${{ github.ref }}" == refs/tags/v* ]]; then
```
ejguan, line 37: Re-land due to `github.base_ref` can be an empty variable. Then, when it's empty, the bash command will fail.

NivekT: noob question: Does the last condition mean that if the branch has a tag that starts with 'v', it will choose `torch_test` as the channel?

NivekT (approved):
LGTM! Thanks!

.github/workflows/ci.yml
ejguan: Yeah.

ejguan: Yep. It's one step during release. I haven't created one for 0.4.0 yet because I will create the first rc at the end of today to incorporate the timeline of cherry-pick. 
See all tags: https://github.com/pytorch/data/tags

PR #518, opened 2022-06-15
Pass through kwargs to json.loads call

ananthsub:
Summary:
The comment of this class suggests the keyword arguments in the constructor will be passed through to the json.loads call. However, `self.kwargs` is not actually passed through to the call.

Updated the example in the README too

Differential Revision: D37162608


NivekT (approved):
LGTM! Thanks for fixing this.

Issue #530, opened 2022-06-22
Multiprocessing and batching / collation


NicolasHug:
It looks like collation works differently depending on whether `MultiProcessingReadingService` is used. Perhaps I am not using this  properly?

```py
import torchdata
from torchdata.dataloader2 import DataLoader2, MultiProcessingReadingService


dp = torchdata.datapipes.iter.IterableWrapper(range(40))
dp = dp.batch(10)
dp = dp.collate()

dl = DataLoader2(dp)
print(list(dl))
# Gives
# [tensor([0, 1, 2, 3, 4, 5, 6, 7, 8, 9]), tensor([10, 11, 12, 13, 14, 15, 16, 17, 18, 19]), tensor([20, 21, 22, 23, 24, 25, 26, 27, 28, 29]), tensor([30, 31, 32, 33, 34, 35, 36, 37, 38, 39])]
# as expected



rs = MultiProcessingReadingService(num_workers=4)
dp = dp.sharding_filter()
dl = DataLoader2(dp, reading_service=rs)
print(list(dl))
# Gives
# [tensor([[0, 1, 2, 3, 4, 5, 6, 7, 8, 9]]), tensor([[10, 11, 12, 13, 14, 15, 16, 17, 18, 19]]), tensor([[20, 21, 22, 23, 24, 25, 26, 27, 28, 29]]), tensor([[30, 31, 32, 33, 34, 35, 36, 37, 38, 39]])]
# Notice the double brackets `[[ ... ]]

NicolasHug:
This is probably related to the default `collate_fn` of the `DataLoader` in the reading service?

https://github.com/pytorch/data/blob/8682f6e09d6facc9d6105c59c9966058c182065e/torchdata/dataloader2/reading_service.py#L119-L130

This is unlikely to be a reasonable fix, but passing `collate_fn=lambda x:x[0]` will lead to consistent results.

ejguan:
> This is unlikely to be a reasonable fix, but passing `collate_fn=lambda x:x[0]` will lead to consistent results.

It would be better when `MultiProcessingReadingService` should a special `collate_fn` to detect the type of output. If it's a single tensor, `collate_fn` should directly return it. Otherwise, like a list of Tensors, we should call default `collate_fn`.



> I think this should be addressed once `MultiProcessingReadingService` stop using the old `DataLoader`.

[...]

PR #531, opened 2022-06-22
[DataLoader2] Fix issue related to duplicate collation

NivekT:
* __->__ #531

Since we want users to always provide their own collation in the DataPipe graph, we add this function to avoid duplicate collation. It will no longer be needed once this reading service no longer relies on DLv1.

Separately, the [linter issue](https://github.com/pytorch/data/issues/364) is modified to track the fact that we want users to always provide a `Collator`.

Fixes part of https://github.com/pytorch/data/issues/530

Differential Revision: [D37348504](https://our.internmc.facebook.com/intern/diff/D37348504)

### DIFF
--- a/torchdata/dataloader2/reading_service.py
+++ b/torchdata/dataloader2/reading_service.py
@@ -88,6 +88,10 @@ class CheckpointableReadingServiceInterface(ReadingServiceInterface):
         pass
 
 
+def _collate_no_op(batch):
+    return batch[0]
+
+
 class MultiProcessingReadingService(ReadingServiceInterface):
     num_workers: int
     pin_memory: bool
@@ -126,6 +130,8 @@ class MultiProcessingReadingService(ReadingServiceInterface):
             multiprocessing_context=self.multiprocessing_context,
             prefetch_factor=self.prefetch_factor,
             persistent_workers=self.persistent_workers,
+            # TODO: `collate_fn` is necessary until we stop using DLv1 https://github.com/pytorch/data/issues/530
+            collate_fn=_collate_no_op,
         )
         return IterableWrapper(self.dl_)  # type: ignore[return-value]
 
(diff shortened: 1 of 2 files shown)


NivekT:
@nivekt has imported this pull request.  If you are a Facebook employee, you can view this diff [on Phabricator](https://www.internalfb.com/diff/D37348504).

torchdata/dataloader2/reading_service.py
```diff
@@ -126,6 +130,8 @@ def initialize(self, datapipe: IterDataPipe) -> IterDataPipe:
             multiprocessing_context=self.multiprocessing_context,
             prefetch_factor=self.prefetch_factor,
             persistent_workers=self.persistent_workers,
+            # TODO: `collate_fn` is necessary until we stop using DLv1 https://github.com/pytorch/data/issues/530
```
VitalyFedyunin, line 133: Please fix batch size, just in case =)

VitalyFedyunin (approved):
Simple test would be nice.

NivekT:
@nivekt has imported this pull request.  If you are a Facebook employee, you can view this diff [on Phabricator](https://www.internalfb.com/diff/D37348504).

ejguan:
> @ejguan @VitalyFedyunin One thought came to mind, should we make passing in a `ReadingService` mandatory? Or defaults to `MultiProcessingReadingService(num_workers=0)`?
> 
> It depends if there is a case where the user doesn't want to pass in a `ReadingService` but I can't think of any (and I assume they'd want the behavior of `MultiProcessingReadingService(num_workers=0)` if they don't provide one .

If users only want to run single process code but also want the DL2 feature of serialization, they might use None. Even though it might be less significant use case, we shouldn't add extra burden on users by requiring them to create a no-op `ReadingService`. And I can't come with any harm to allow None `ReadingService`.

PR #534, opened 2022-06-22
Update documentation for S3 DataPipes

### LINKED ISSUE #523: Document how to create a DataLoader when reading data from S3
### 📚 The doc issue

I find the provided example [here](https://github.com/pytorch/data/blob/main/torchdata/datapipes/iter/load/README.md#example) a bit confusing. 

```
from torchdata.datapipes.iter import S3FileLister, S3FileLoader

s3_prefixes = ['s3://bucket-name/folder/', ...]
dp_s3_urls = S3FileLister(s3_prefixes)
dp_s3_files = S3FileLoader(s3_urls) # outputs in (url, StreamWrapper(BytesIO))
# more datapipes to convert loaded bytes, e.g.
datapipe = StreamWrapper(dp_s3_files).parse_csv(delimiter=' ')

for d in datapipe: # Start loading data
    pass
```

First, I think there is a mistake in the example: `s3_urls` should be `dp_s3_urls`?
Second, it is not clear why `parse_csv(delimiter=' ')` is being used.
Last, I can't access my data after creating the `datapipe`. It would be great to have an example similar to [this one of the old plugin](https://github.com/aws/amazon-s3-plugin-for-pytorch/blob/master/examples/s3_cv_transform.py)

I believe that an example of how to load a S3 folder containing images into a `torch.utils.data.DataLoader` would be very useful for new users (like me).

Provide an example that starts with a S3 url of a folder with some images, and produce a dataloader with such images.

ejguan:
Fixes #523

- Make in-line doc as the single source of truth
- Add `sharding` and `shuffle` to the example for distributed training
- Rephrase each arguments based on the offline discussion with @ydaiming

### DIFF
--- a/torchdata/datapipes/iter/load/README.md
+++ b/torchdata/datapipes/iter/load/README.md
@@ -12,12 +12,12 @@ S3 IO datapipes are included when building with flag `BUILD_S3=1`. The following
 source with S3 datapipes.
 
 ```bash
-pip uninstall torchdata -y
-git clone https://github.com/pytorch/data.git
-cd data
-python setup.py clean
 BUILD_S3=1 python setup.py install
 ```
+
+We also offer nightly and official (>=0.4.0) TorchData releases integrated with `AWSSDK` on the most of platforms.
+Please check the [link](https://github.com/pytorch/data/tree/main/packaging#awssdk) for the list of supported platforms
+with the pre-assembled binaries.
 
 If you'd like to use customized installations of `pybind11` or `aws-sdk-cpp`, you may set the following flags when
 building from source.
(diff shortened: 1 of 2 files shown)


torchdata/datapipes/iter/load/README.md
```diff
@@ -12,10 +12,6 @@ S3 IO datapipes are included when building with flag `BUILD_S3=1`. The following
 source with S3 datapipes.
 
 ```bash
-pip uninstall torchdata -y
-git clone https://github.com/pytorch/data.git
-cd data
-python setup.py clean
 BUILD_S3=1 python setup.py install
```
NivekT, line 15: No change needed. Just a question, I assume this isn't relevant if people install from `pip` or `conda`?

NivekT (approved):
LGTM!

torchdata/datapipes/iter/load/README.md
ejguan: Users don't need to uninstall `torchdata` and re-install with `BUILD_S3`. I will add a few words about torchdata releases with pre-assembled AWSSDK.

torchdata/datapipes/iter/load/README.md
```diff
@@ -31,43 +31,12 @@ USE_SYSTEM_LIBS=1 # uses both pre-installed pybind11 and aws-sdk-cpp
...
-dp_s3_urls = S3FileLister(s3_prefixes)
-dp_s3_files = S3FileLoader(s3_urls) # outputs in (url, StreamWrapper(BytesIO))
-# more datapipes to convert loaded bytes, e.g.
-datapipe = StreamWrapper(dp_s3_files).parse_csv(delimiter=' ')
-
-for d in datapipe: # Start loading data
-    pass
-```
+- [`S3FileLister`](https://pytorch.org/data/main/generated/torchdata.datapipes.iter.S3FileLister.html#s3filelister)
+- [`S3FileLister`](https://pytorch.org/data/main/generated/torchdata.datapipes.iter.S3FileLoader.html#s3fileloader)
```
enric1994, line 39: @ejguan  This should be `S3FileLoader`

ejguan: Good catch! Thanks.

torchdata/datapipes/iter/load/s3io.py
```diff
@@ -77,31 +72,34 @@ class S3FileLoaderIterDataPipe(IterDataPipe[Tuple[str, StreamWrapper]]):
...
-        ...     pass
+        >>> from torchdata.datapipes.iter import IterableWrapper, S3FileLoader
+        >>> dp_s3_urls = IterableWrapper(['s3://bucket-name/folder/', ...]).list_files_by_s3()
+        # In order to make sure data are shuffled and sharded in the
+        # distributed environment, `shuffle`  and `sharding_filter`
+        # are required. For detail, please check our tutorial in:
+        # https://pytorch.org/data/main/tutorial.html#working-with-dataloader
+        >>> sharded_s3_urls = dp_s3_urls.shuffle().sharding_filter()
+        >>> dp_s3_files = S3FileLoader(sharded_s3_urls)
+        >>> for url, fd in datapipe: # Start loading data
```
enric1994, line 97: @ejguan Should be `dp_s3_files` instead of `datapipe`

Issue #517, opened 2022-06-14
AIStore Datapipes (AISFileListerIterDataPipe and AISFileLoaderIterDataPipe) 


gaikwadabhishek:
### 🚀 The feature

Add a new file loader and lister for AIStore, similar to [s3io](https://github.com/pytorch/data/blob/main/torchdata/datapipes/iter/load/s3io.py) and [fsspec](https://github.com/pytorch/data/blob/main/torchdata/datapipes/iter/load/fsspec.py).

For introduction and background, see PyTorch Blog “[Efficient PyTorch I/O library for Large Datasets, Many Files, Many GPUs](https://pytorch.org/blog/efficient-pytorch-io-library-for-large-datasets-many-files-many-gpus/).”

### Motivation, pitch

[AIStore](https://github.com/NVIDIA/aistore) (AIS for short) is a highly available lightweight object storage system that specifically focuses on petascale deep learning. As a reliable redundant storage, AIS supports n-way mirroring and erasure coding. But it is not purely – or not only – a storage system: it’ll shuffle user datasets and run custom extract-transform-load workloads.

AIS is an elastic cluster that can grow and shrink at runtime and can be ad-hoc deployed, with or without Kubernetes, anywhere from a single Linux machine to a bare-metal cluster of any size.

AIS fully supports Amazon S3, Google Cloud, and Microsoft Azure backends, providing a unified namespace across multiple connected backends and/or other AIS clusters, and [more](https://github.com/NVIDIA/aistore#features). Getting started with AIS will take only a few minutes (prerequisites boil down to having a Linux with a disk) and can be done either by running a prebuilt [all-in-one docker image](https://github.com/NVIDIA/aistore/tree/master/deploy) or directly from the open-source.

We are hoping that once integrated with TorchData, AIS can prove to be useful to the community.

### Alternatives

There are numerous ways to load TorchData pipelines: s3io (for objects on Amazon S3), fsspec (for files), and many more. The difference that AIS can maybe contribute is that it can be used both as a standalone reliable/scalable storage and/or be deployed in front of any of the [supported backends](https://github.com/NVIDIA/aistore/blob/master/docs/overview.md#at-a-glance), including (but not limited to) Amazon S3.
 
[...]

PR #545, opened 2022-06-24
adding aistore datapipe

gaikwadabhishek:
Fixes #{[517](https://github.com/pytorch/data/issues/517)}

### Changes
- Added `aisio.py` (Iterable Datapipe for AIStore backends)
- Added unit tests in ~~`test/test_local_io.py`~~ `test/test_aistore.py`
- Added GitHub action for running AIStore in ~~`CI.yml`~~ `.github/aistore_ci.yml` workflow

### Questions to maintainers
- We are unsure about the documentation generated on PyTorch to refer it in `README.md`, so I have tentatively added a URL similar to s3io functions. (see ```torchdata/datapipes/iter/load/README.md```)

Signed-off-by: Abhishek Gaikwad <gaikwadabhishek1997@gmail.com>

### DIFF
--- a/.github/workflows/aistore_ci.yml
+++ b/.github/workflows/aistore_ci.yml
@@ -1,0 +1,59 @@
+name: Run AIStore Datapipe Test
+on:
+  push:
+    branches:
+      - main
+      - release/*
+    tags:
+  pull_request:
+    types: [opened, synchronize, reopened, labeled]
+    branches:
+      - main
+      # For PR created by ghstack
+      - gh/*/*/base
+      - release/*
+
+jobs:
+  test:
+    if: ${{ github.repository_owner == 'pytorch' }}
+    runs-on: ${{ matrix.os }}
+    strategy:
+      fail-fast: false
+      matrix:
+        os:
+          - macos-latest
+          - ubuntu-latest
+        python-version:
+          - 3.7
+          - 3.8
+          - 3.9
+    steps:
+      - name: Get PyTorch Channel
+        shell: bash
+        run: |
+          if [[ "${{ github.base_ref }}" == release/* ]] || [[ "${{ github.ref }}" == refs/heads/release/* ]] || [[ "${{ github.ref }}" == refs/tags/v* ]]; then
+            PT_CHANNEL="https://download.pytorch.org/whl/test/cpu/torch_test.html"
+          else
+            PT_CHANNEL="https://download.pytorch.org/whl/nightly/cpu/torch_nightly.html"
+          fi
+          echo "::set-output name=value::$PT_CHANNEL"
+        id: pytorch_channel
+      - name: Setup Python ${{ matrix.python-version }}
+        uses: actions/setup-python@v2
+        with:
+          python-version: ${{ matrix.python-version }}
+      - name: Check out source repository
+        uses: actions/checkout@v2
+      - name: Install dependencies
+        run: |
+          pip3 install -r requirements.txt
+          pip3 install --pre torch -f "${{ steps.pytorch_channel.outputs.value }}"
+      - name: Run AIStore local deployment
+        uses: NVIDIA/aistore@master
+      - name: Build TorchData
+        run: |
+          python setup.py install
+      - name: Install test requirements
+        run: pip3 install -r test/requirements_aistore.txt
+      - name: Run AIStore DataPipe tests with pytest
+        run: pytest --no-header -v test/test_aistore.py
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -80,18 +80,18 @@ jobs:
         if: ${{ ! contains(github.event.pull_request.labels.*.name, 'ciflow/slow') }}
         run:
           pytest --no-header -v test --ignore=test/test_period.py --ignore=test/test_text_examples.py
-          --ignore=test/test_audio_examples.py
+          --ignore=test/test_audio_examples.py --ignore=test/test_aistore.py
       - name: Run DataPipes tests with pytest (including slow tests)
         if: ${{ contains(github.event.pull_request.labels.*.name, 'ciflow/slow') }}
         run:
           pytest --no-header -v test --ignore=test/test_period.py --ignore=test/test_text_examples.py
-          --ignore=test/test_audio_examples.py
+          --ignore=test/test_audio_examples.py --ignore=test/test_aistore.py
         env:
           PYTORCH_TEST_WITH_SLOW: 1
       - name: Run DataPipes period tests with pytest
         if: ${{ contains(github.event.pull_request.labels.*.name, 'ciflow/period') }}
         run:
           pytest --no-header -v test/test_period.py --ignore=test/test_text_examples.py
-          --ignore=test/test_audio_examples.py
+          --ignore=test/test_audio_examples.py --ignore=test/test_aistore.py
         env:
           PYTORCH_TEST_WITH_SLOW: 1
--- a/test/test_aistore.py
+++ b/test/test_aistore.py
@@ -1,0 +1,101 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
+# All rights reserved.
+#
+# This source code is licensed under the BSD-style license found in the
+# LICENSE file in the root directory of this source tree.
+
+import random
+import string
+import tempfile
+import unittest
+
+from torchdata.datapipes.iter import AISFileLister, AISFileLoader
+
+try:
+    from aistore.client.api import Client
+    from aistore.client.errors import AISError, ErrBckNotFound
+
+    AIS_CLUSTER_ENDPT = "http://localhost:8080"
+
+    HAS_AIS = Client(AIS_CLUSTER_ENDPT).is_aistore_running()
+except (ImportError, ConnectionError):
+    HAS_AIS = False
+skipIfNoAIS = unittest.skipIf(not HAS_AIS, "AIS not running or library not installed")
+
+
+@skipIfNoAIS
+class TestDataPipeLocalIO(unittest.TestCase):
+    def setUp(self):
+        # initialize client and create new bucket
+        self.client = Client(AIS_CLUSTER_ENDPT)
+        letters = string.ascii_lowercase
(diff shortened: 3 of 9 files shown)


.github/workflows/ci.yml
```diff
@@ -69,6 +69,9 @@ jobs:
           pip3 install --pre torch -f "${{ steps.pytorch_channel.outputs.value }}"
           pip3 install cmake ninja
           echo "/home/runner/.local/bin" >> $GITHUB_PATH
+      - name: Run AIStore local deployment
+        if: matrix.os != 'windows-latest'
+        uses: NVIDIA/aistore@master
```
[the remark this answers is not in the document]
gaikwadabhishek: **Running AIStore as a Docker Container**
Generally, AIStore is a storage cluster that requires a volume (ability to mount using `-v` in docker). In Github workflow environments, when we try to mount the `dev/root` as a volume to the container we receive an error during the startup which says that `dev/root` doesn't have underlying block storage. We can probably overcome this by running it from `/` or `/tmp` but we at AIStore don't recommend this. Currently, we are running it as a local deployment from scratch as a [Github action](https://github.com/NVIDIA/aistore/blob/master/action.yml).
**Running AIStore in different pipeline**
We can run AIStore as a separate workflow and import it into the CI workflow to run the python tests. But then we need to have a reliable mechanism to synchronize the testing pipeline with the AIS cluster starting up (in a separate pipeline). The latter may take anywhere between tens of milliseconds to maybe 10-20s depending on the load of the machine we run in. I have tried to simulate the [torch/data pipeline of this PR](https://github.com/gaikwadabhishek/data/runs/7032258015). 
 
Let me know what you guys think, we can work on this.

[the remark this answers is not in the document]
ejguan: The reason I want to have a separate workflow is we can execute tests for AIStore within that workflow without touching others. With separate workflow, I am able to disable the test easily using GHA whenever we find constant failing tests.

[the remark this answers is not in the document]
ejguan: > We can run AIStore as a separate workflow and import it into the CI workflow to run the python tests.

You don't need to do this. Since you AIStore doesn't require AWSSDK installed, you can basically copy paste the most of parts from `ci.yml` by removing all AWS related steps like:
- https://github.com/pytorch/data/blob/75f31dc99dcc3159ee9a3df73d3862c76754823c/.github/workflows/ci.yml#L44-L49
- https://github.com/pytorch/data/blob/75f31dc99dcc3159ee9a3df73d3862c76754823c/.github/workflows/ci.yml#L54-L61
- https://github.com/pytorch/data/blob/75f31dc99dcc3159ee9a3df73d3862c76754823c/.github/workflows/ci.yml#L70-L71
- Remove BUILD_S3=1. https://github.com/pytorch/data/blob/75f31dc99dcc3159ee9a3df73d3862c76754823c/.github/workflows/ci.yml#L75-L76
- Modify the tests steps to only execute the tests for AIStore.

setup.py
```diff
@@ -195,4 +195,7 @@ def remove_extension(pattern):
             "build_ext": CMakeBuild,
             "clean": clean,
         },
+        extras_require={
+            "aistore": ["aistore"],
+        },
```
ejguan, line 200: Since `aistore` is an optional dependency, we don't need to add it here.

.github/workflows/ci.yml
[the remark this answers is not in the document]
gaikwadabhishek: Okay, I have made the changes. Do you want to trigger this workflow on a specific label? Currently the condition is similar to the main `ci.yml` -
```if:
      ${{ github.repository_owner == 'pytorch' && (github.event.action != 'labeled' ||
      startsWith(github.event.label.name, 'ciflow')) }}```

**Review by NivekT — COMMENTED** [review]
This PR mostly looks good to me. I will let @ejguan comment on the proper CI setup.

Question: Are there AWS/GCP/Azure remote setups that your library currently test against for every build? We might want to something similar here for different DataPipes (not just `aistore`).

**Comment by msaroufim** [comment]
> This PR mostly looks good to me. I will let @ejguan comment on the proper CI setup.
> 
> Question: Are there AWS/GCP/Azure remote setups that your library currently test against for every build? We might want to something similar here for different DataPipes (not just `aistore`).

Let's talk about this offline more @NivekT I was thinking of something similar for benchmarks where I can setup some machine configuration an instance type with some specific configuration for example HDD vs SSD and then run a script directly on it. There's a few options like `ray up`, cloudformation, ec2 launch, eks, teraform and I'm in the process of figuring out what's most convenient. In your case is the multi cloud requirement a must? In which case teraform may be the obvious thing to leverage more

File: .github/workflows/aistore_ci.yml [hunk 3]
```diff
@@ -0,0 +1,62 @@
...
+      - main
+      # For PR created by ghstack
+      - gh/*/*/base
+      - release/*
+
+jobs:
+  test:
+    if:
+      ${{ github.repository_owner == 'pytorch' && (github.event.action != 'labeled' ||
+      startsWith(github.event.label.name, 'ciflow')) }}
```
ejguan, line 20: Like you said, no need to run the workflow for a specific label

.github/workflows/aistore_ci.yml
```diff
@@ -0,0 +1,62 @@
...
+        uses: actions/setup-python@v2
+        with:
+          python-version: ${{ matrix.python-version }}
+      - name: Check out source repository
+        uses: actions/checkout@v2
+      - name: Install dependencies
+        run: |
+          pip3 install -r requirements.txt
+          pip3 install --pre torch -f "${{ steps.pytorch_channel.outputs.value }}"
+          echo "/home/runner/.local/bin" >> $GITHUB_PATH
```
ejguan, line 53: nit: you can remove this line for $GITHUB_PATH

test/test_aistore.py
```diff
@@ -0,0 +1,101 @@
...
+from torchdata.datapipes.iter import AISFileLister, AISFileLoader
+
+try:
+    from aistore.client.api import Client
+    from aistore.client.errors import AISError, ErrBckNotFound
+
+    AIS_CLUSTER_ENDPT = "http://localhost:8080"
+
+    HAS_AIS = Client(AIS_CLUSTER_ENDPT).is_aistore_running()
+except ImportError or ConnectionError:
```
ejguan, line 21: Noob question because I am not familiar with this syntax. Should it be
`except (ImportError, ConnectionError):`?

test/test_aistore.py
```diff
@@ -0,0 +1,101 @@
...
+
+        # create 10 objects in the `/temp` dir
+        for i in range(num_objs):
+            object_body = "test string" * random.randrange(1, 10)
+            content = object_body.encode("utf-8")
+            obj_name = f"temp/obj{ i }"
+            with tempfile.NamedTemporaryFile() as file:
+                file.write(content)
+                file.flush()
+                self.client.put_object(self.bck_name, obj_name, file.name)
```
ejguan, line 45: Would the client copy the tempfile to the bucket? IIRC, `tempfile` will be removed from the fs on the completion of the context.

torchdata/datapipes/iter/load/aisio.py
```diff
@@ -0,0 +1,126 @@
...
+    -   Input must be a list and direct URLs are not supported.
+    -   length is -1 by default, all calls to len() are invalid as
+        not all items are iterated at the start.
+    -   This internally uses AIStore Python SDK.
+
+    Args:
+        source_datapipe(IterDataPipe[str]): a DataPipe that contains URLs/URL
+                                            prefixes to objects on AIS
+        length(int): length of the datapipe
+        url(str): AIStore endpoint
```
ejguan, line 50: Do you mind change the order of these two arguments to the same order as the signature.

torchdata/datapipes/iter/load/aisio.py
```diff
@@ -0,0 +1,126 @@
...
+
+    Note:
+    -   This function also supports files from multiple backends (`aws://..`, `gcp://..`, etc)
+    -   Input must be a list and direct URLs are not supported.
+    -   This internally uses AIStore Python SDK.
+
+    Args:
+        source_datapipe(IterDataPipe[str]): a DataPipe that contains URLs/URL prefixes to objects
+        length(int): length of the datapipe
+        url(str): AIStore endpoint
```
ejguan, line 97: Ditto

ejguan (commented):
Thank you for your contribution. Generally looks good to me. And, could you please sign the CLA before we can merge the PR?

test/test_aistore.py
gaikwadabhishek: Yes, the file created will be added to the AIS Bucket. Tempfile are deleted as soon as they are closed. In `tearDown()` we are destroying the bucket and all the objects within.

test/test_aistore.py
gaikwadabhishek: fixed

gaikwadabhishek:
> Thank you for your contribution. Generally looks good to me. And, could you please sign the CLA before we can merge the PR?

Hi @ejguan @NivekT 
I am still waiting for my FB CLA to be approved. I work for NVIDIA and I've been added to the corporate CLA by my employer. We are waiting for the changes to reflect here. We have also mailed to cla@fb.com regarding this and are waiting for their reply.

gaikwadabhishek:
@ejguan fixed the `mypi.ini` configuration to ignore the missing import of `aistore.`. Can you please retrigger the workflow

ejguan:
FYI: The failures from TorchVision CI are unrelated to your PR.

msaroufim:
@gaikwadabhishek your CLA application should have been approved just retrigger CI with an empty commit and the Facebook CLA check should pass

msaroufim:
@zpao as FYI looks like CLA Check is still red

gaikwadabhishek:
Hi @ejguan @NivekT @msaroufim,
Thanks for your support. 
CLA is approved now. Can we merge this?

msaroufim (approved):

VitalyFedyunin (approved):

Issue #538, opened 2022-06-23
Warn about pickle-ablity when using `dp.map(some_local_function)` ?


NicolasHug:
`torchdata` issues a warning about pickle when we use lambdas (which is great!)
Another kind of function that isn't compatible with pickle are local functions. Would it be possible to throw the same warning there?

```py
import torchdata
import pickle

def make_dp():

    def f(x):  # local function, not pickleable
        return x

    return torchdata.datapipes.iter.IterableWrapper(range(40)).map(f)

dp = make_dp()  # no warning

f = "/tmp/dp"
pickle.dump(dp, open(f, "wb"))  # fails
```

ejguan:
It makes sense to raise warning for local function. We might be able to use `"<locals>" in fn.__qualname__` to determine it. Sending a patch.

PR #547, opened 2022-06-24
Raise warning for unpickable local function

ejguan:
Summary:
Fixes https://github.com/pytorch/data/issues/538
- Improve the validation function to raise warning about unpickable function when either lambda or local function is provided to DataPipe.
- The inner function from functools.partial object is extracted as well for validation
- Mimic the behavior of pickle module for local lambda function: It would only raise Error for the local function rather than lambda function. So, we will raise warning about local function not lambda function.
```py

>>> import pickle
>>> def fn():
...     lf = lambda x: x
...     pickle.dumps(lf)
>>> pickle.dumps(fn)
AttributeError: Can't pickle local object 'fn.<locals>.<lambda>'
```

This Diff also fixes the Error introduced by https://github.com/pytorch/pytorch/pull/79344

Differential Revision: D37417556


ejguan:
> The CI failure is due to [pytorch/pytorch#80232](https://github.com/pytorch/pytorch/pull/80232) isn't in nightly?

@NivekT You are right. As long as the tests in phabricator pass, we should be fine.

PR #556, opened 2022-06-28
Fix binary path for official release

ejguan:
Found the issue during official release
See: https://github.com/pytorch/data/runs/7097431966?check_suite_focus=true


NivekT (approved):
LGTM!

PR #559, opened 2022-06-29
Make graph `traverse` working with unhashable `DataPipe`

ejguan:
Summary:
This Diff removes the requirement for `traverse` function that `DataPipe` needs to be hash-able. `traverse` function now is using `id` of `DataPipe` instance rather than `DataPipe` itself as the key for both `cache` and graph.

But, it requires the changes of type of `DataPipeGraph` from `Dict[DataPipe, "DataPipeGraph"]` to `Dict[int, Tuple[DataPipe, "DataPipeGraph"]]`.

Differential Revision: D37354153

Ref PR in PyTorch: https://github.com/pytorch/pytorch/pull/80509


torchdata/dataloader2/_graph_utils.py
```diff
@@ -1,18 +0,0 @@
...
-
-from torch.utils.data.graph import traverse
-
-from torchdata.datapipes.iter import IterDataPipe
-from torchdata.datapipes.map import MapDataPipe
-
-__all__ = ["DataPipe", "DataPipeGraph", "traverse"]
-
-DataPipe = Union[IterDataPipe, MapDataPipe]
-DataPipeGraph = Dict[DataPipe, "DataPipeGraph"]  # type: ignore[misc]
```
ejguan, line 18: Moved back to PyTorch Core.

NivekT (approved):
LGTM! I think it will be good if we can add some tests for the linter, but not the most urgent thing.

ejguan:
I do create a few tests in test/test_linter.py

PR #562, opened 2022-06-29
[BE] Unify Map style DataPipes to support input_col and output_col syntax

VitalyFedyunin:
* #652
* __->__ #562

Differential Revision: [D40146676](https://our.internmc.facebook.com/intern/diff/D40146676)


torchdata/datapipes/iter/util/common.py
```diff
@@ -0,0 +1,16 @@
...
+# All rights reserved.
+#
+# This source code is licensed under the BSD-style license found in the
+# LICENSE file in the root directory of this source tree.
+
+from torchdata.datapipes.iter import Mapper
+
+
+class MapTemplateIterDataPipe(Mapper):
+    def __init__(self, source_datapipe, input_col=None, output_col=None) -> None:
```
NivekT, line 11: 1. Let's add a docstring for the purpose of this class.
2. Add the class `MapTemplate` to RST documnetation file as well.

NivekT (commented):
Should we also add one for `flatmap`?

VitalyFedyunin:
UPD: Temporary make a copy of `FlatMapper`
UPD: Rewrote JsonParser as example.

test/test_local_io.py
```diff
@@ -317,6 +317,23 @@ def is_nonempty_json(path_and_stream):
         ]
         self.assertEqual(expected_res, list(json_dp))
 
+        def as_dict(data):
+            return {"file_name": data[0], "stream": data[1]}
+
+        dict_dp = datapipe_nonempty.map(as_dict)
+        parsed_dp = dict_dp.parse_json_files(input_col="stream", output_col="json")
```
VitalyFedyunin, line 324: This is the most important part of the PR. IMO it will simplify some of the domains code.

torchdata/datapipes/iter/util/common.py
```diff
@@ -0,0 +1,112 @@
...
+        fn = self._map
+        super().__init__(source_datapipe, fn=fn, input_col=input_col, output_col=output_col)
+
+    def _map(self, *args, **kwargs) -> T_co:
+        raise NotImplementedError
+
+
+@functional_datapipe("flatmap_proto")
+class FlatMapperProtoIterDataPipe(IterDataPipe[T_co]):
```
ejguan, line 29: Do you want to directly change the `FlatMapper` here?
https://github.com/pytorch/data/blob/14b2291f0f0488632e9689b9735c58a3b113d237/torchdata/datapipes/iter/transform/callable.py#L93-L94

torchdata/datapipes/iter/util/jsonparser.py
```diff
@@ -33,15 +33,13 @@ class JsonParserIterDataPipe(IterDataPipe[Tuple[str, Dict]]):
...
 
-    def __init__(self, source_datapipe: IterDataPipe[Tuple[str, IO]], **kwargs) -> None:
-        self.source_datapipe: IterDataPipe[Tuple[str, IO]] = source_datapipe
+    def __init__(self, source_datapipe: IterDataPipe, input_col=1, output_col=None, **kwargs) -> None:
+        super().__init__(source_datapipe, input_col=input_col, output_col=output_col)
         self.kwargs = kwargs
 
-    def __iter__(self) -> Iterator[Tuple[str, Dict]]:
-        for file_name, stream in self.source_datapipe:
+    def _map(self, stream: IO):
```
ejguan, line 40: Do you think we should change the type to `StreamWrapper` here?

torchdata/datapipes/iter/util/common.py
VitalyFedyunin: Yes I do, was making sure I cover all use cases before replacement.

torchdata/datapipes/iter/util/jsonparser.py
VitalyFedyunin: We are fine with any stream as input.

torchdata/datapipes/iter/util/common.py
VitalyFedyunin: I will update DP contributors doc and add docstrings

torchdata/datapipes/iter/util/common.py
```diff
@@ -0,0 +1,112 @@
...
+T_co = TypeVar("T_co", covariant=True)
+
+
+class MapTemplateIterDataPipe(Mapper[T_co]):
+    def __init__(self, source_datapipe, input_col=None, output_col=None) -> None:
+        fn = self._map
+        super().__init__(source_datapipe, fn=fn, input_col=input_col, output_col=output_col)
+
+    def _map(self, *args, **kwargs) -> T_co:
+        raise NotImplementedError
```
ejguan, line 25: BTW, we might want to use `abc. abstractmethod` here to guarantee to implement the method at the construct time.

torchdata/datapipes/iter/util/common.py
```diff
@@ -0,0 +1,112 @@
...
+        raise TypeError(f"{type(self).__name__} instance doesn't have valid length")
+
+
+class FlatMapTemplateIterDataPipe(FlatMapperProtoIterDataPipe[T_co]):
+    def __init__(self, source_datapipe, input_col=None, output_col=None, length: int = -1) -> None:
+        fn = self._flatmap
+        super().__init__(source_datapipe, fn=fn, input_col=input_col, output_col=output_col, length=length)
+
+    def _flatmap(self, *args, **kwargs) -> Iterable[T_co]:
+        raise NotImplementedError
```
ejguan, line 112: Ditto

torchdata/datapipes/iter/transform/callable.py
```diff
@@ -162,6 +165,115 @@ def __len__(self) -> int:
...
+        length: int = -1,
+    ) -> None:
+        super().__init__()
+        self.datapipe = datapipe
+        self.length = length
+
+        _check_unpickable_fn(fn)
+        self.fn = fn  # type: ignore[assignment]
+
+        self.input_col = input_col
```
ejguan, line 219: We can also validate `input_col` with `fn` using `torch.utils.data.datapipes.utils.common.validate_input_col`
https://github.com/pytorch/pytorch/blob/cfb9d0d23314fd28be118b6ca280ded55364e71c/torch/utils/data/datapipes/utils/common.py#L24

ejguan (approved):

torchdata/datapipes/iter/util/jsonparser.py
```diff
@@ -33,15 +33,13 @@ class JsonParserIterDataPipe(IterDataPipe[Tuple[str, Dict]]):
         [('1.json', ['foo', {'bar': ['baz', None, 1.0, 2]}]), ('2.json', {'__complex__': True, 'real': 1, 'imag': 2})]
     """
```
NivekT, line 34: Adding usage example for `input_col` argument

```suggestion
        [('1.json', ['foo', {'bar': ['baz', None, 1.0, 2]}]), ('2.json', {'__complex__': True, 'real': 1, 'imag': 2})]
    >>> datapipe_extra_col = datapipe1.map(lambda x: (x[0], "extra", x[1]))
    >>> # Use `input_col` if the stream is not in position 1
    >>> json_dp2 = datapipe_extra_col.parse_json_files(input_col=2)
    >>> list(json_dp2)
    [
        ("1.json", "extra", ["foo", {"bar": ["baz", None, 1.0, 2]}]),
        ("2.json", "extra", {"__complex__": True, "real": 1, "imag": 2}),
    ]
    """
```

torchdata/datapipes/iter/util/jsonparser.py
```diff
@@ -5,14 +5,14 @@
...
 from torchdata.datapipes import functional_datapipe
-from torchdata.datapipes.iter import IterDataPipe
+from torchdata.datapipes.iter import IterDataPipe, MapTemplate
 
 
 @functional_datapipe("parse_json_files")
-class JsonParserIterDataPipe(IterDataPipe[Tuple[str, Dict]]):
+class JsonParserIterDataPipe(MapTemplate):
     r"""
     Reads from JSON data streams and yields a tuple of file name and JSON data (functional name: ``parse_json_files``).
```
NivekT, line 18: Need to update docstring to add new arguments

NivekT (approved):
nit: This PR is fine (with nit comments) but there is something unsatisfying about how `input_col` and `output_col` work in the case where I want to select an element from a tuple/list?

A common return pattern that we have is `(Path, Stream)` from archives/streams/etc. Let's say I opened a file and now I have a DataPipe of tuples of `(Path, Stream)`, and I want to throw away the path and parse the stream. I will have to do something like this

```python
datapipe = FileOpener(json_files, mode="b")
datapipe = datapipe.parse_json_files()
datapipe = datapipe.map(lambda x: x[1])  # Ideally this should not be needed, and handled by `input_col`, `output_col`
```
I feel like ideally the `input_col` and `output_col` argument should allow for selection (unpack when the elements are list of length 1 or tuple of length 1).

We likely also want to allow the case where the input is not a list/tuple:
```python
datapipe = IterableWrapper([stream1, stream2])
datapipe = datapipe.parse_json_files()  # I think we should support this, but it currently doesn't work
```

What do you two think?

cc: @ejguan

torchdata/datapipes/iter/transform/callable.py
```diff
@@ -162,6 +165,115 @@ def __len__(self) -> int:
         raise TypeError(f"{type(self).__name__}'s length relies on the output of its function.")
 
 
+# TODO(VitalyFedyunin): Replacing FlatMapperIterDataPipe will require BC breaking change of input_col behaviour
```
NivekT, line 168: Is this because when `input_col` is specified`, the new version will try to return a list/tuple if the original input is a list/tuple?

ejguan:
> nit: This PR is fine (with nit comments) but there is something unsatisfying about how `input_col` and `output_col` work in the case where I want to select an element from a tuple/list?
> 
> Note that we do not have to address the below within this PR.
> 
> A common return pattern that we have is `(Path, Stream)` from archives/streams/etc. Let's say I opened a file and now I have a DataPipe of tuples of `(Path, Stream)`, and I want to throw away the path and parse the stream. I will have to do something like this
> 
> ```python
> datapipe = FileOpener(json_files, mode="b")
> datapipe = datapipe.parse_json_files()
> datapipe = datapipe.map(lambda x: x[1])  # Ideally this should not be needed, and handled by `input_col`, `output_col`
> ```
> 
> I feel like ideally the `input_col` and `output_col` argument should allow for selection (unpack when the elements are list of length 1 or tuple of length 1).
> 
> We likely also want to allow the case where the input is not a list/tuple:
> 
> ```python
> datapipe = IterableWrapper([stream1, stream2])
> datapipe = datapipe.parse_json_files()  # I think we should support this, but it currently doesn't work
> ```
> 
> What do you two think?
> 
> cc: @ejguan

I think we have discussed about it earlier about the expected behavior of the combination of `input_col` and `output_col`. Currently, when `input_col` is specified buy `output_col` is not, we will do in-place operation. In order to support selection, i can't find a good value for `output_col` to indicate that we want to select a single column.

> datapipe = datapipe.parse_json_files()  # I think we should support this, but it currently doesn't work

Do you mean you want it still works with `stream` input rather than a tuple of `(pathname, stream)`

NivekT:
> I think we have discussed about it earlier about the expected behavior of the combination of `input_col` and `output_col`. Currently, when `input_col` is specified buy `output_col` is not, we will do in-place operation. In order to support selection, i can't find a good value for `output_col` to indicate that we want to select a single column.

One potential solution is to have `output_col=[]` or `output_col=tuple()` as selection. There may be side effects that I currently cannot think of.

> Do you mean you want it still works with `stream` input rather than a tuple of `(pathname, stream)`

Yes, it may require some adjustment to how `input_col` and `MapTemplate` works to handle non-list/tuple inputs.

NivekT:
Sorry I accidentally edited your comment while quoting it.

ejguan:
> Yes, it may require some adjustment to how `input_col` and `MapTemplate` works to handle non-list/tuple inputs.

I think it's doable after this PR. We can let `JsonParser` be a `MapTemplate` with `input_col=1` by default to prevent BC breaking.
Users can specify `input_col=None` to make it directly working with `stream`.

PR #569, opened 2022-07-04
Raise Error when HTTPReader get 404 Response (#160)

### LINKED ISSUE #160: Raise Error when HTTPReader get 503 Response
### 🐛 Describe the bug

As I was testing the https://github.com/pytorch/data/blob/main/examples/text/CC100.ipynb, it would yield Error saying the object can not be decoded. I dug a little bit and found the root cause was HTTPResponse is a 503 Error.

But, I would expect the Error/Exception should be caught by HTTPReader
https://github.com/pytorch/pytorch/blob/2431218ee4ac1d2f016188c29869d817f6e3c70c/torch/utils/data/datapipes/iter/httpreader.py#L33-L38.

Not sure if this is related to streaming option. Needs to investigate.

### Versions

main

### Additional Context for New Contributors

See comment below

cc: @NivekT

diegoaichele:
Fixes #160 

### Changes

- Add line of requests raise_for_status() will only allow 200 accepted requests

### Comments
- This raise only run when the method __iter__ of HTTPReader is called.
- It is not only raise for the 503 error, it also runs for 404, 500,501,502 ...
- I simulate them using FastAPI with the next code.

```
from fastapi import FastAPI, HTTPException

app = FastAPI()

@app.get("/raise/{error_int}")
async def raise_error(error_int: int):
    raise HTTPException(status_code=error_int)
```


msaroufim (approved):

VitalyFedyunin (commented):
Can you please add a test to https://github.com/pytorch/data/blob/main/test/test_remote_io.py#L60 which tries to load an unexisting url?

diegoaichele:
@VitalyFedyunin 
feat: Add Test HTTPError exception on the test_http_reader_iterdatapipe
shorten the error message ([:54]), but the important thing is that the 404 response comes out

test/test_remote_io.py
```diff
@@ -82,6 +82,14 @@ def test_http_reader_iterdatapipe(self):
 
         # __len__ Test: returns the length of source DataPipe
         self.assertEqual(1, len(http_reader_dp))
+        
+        # Error Test: test if the Http Reader raises an error when the url is invalid
+        error_url = "https://github.com/pytorch/data/this/url/dont/exist"
+        http_error_dp = HttpReader(IterableWrapper([error_url]), timeout=timeout)
+        with self.assertRaises(Exception) as e:
```
VitalyFedyunin, line 89: Please use `assertRaisesRegex` it will reduce lines 91-92 to simple check.

VitalyFedyunin (approved):

NivekT (approved):

PR #577, opened 2022-07-07
Improved FAQ readability with collapsible Q&A.

lucaspar:
### Changes

+ Changed Q&A from plain text to collapsible sections in `README.md`.
+ Fixed typo in FAQ answer.

### DIFF
--- a/README.md
+++ b/README.md
@@ -175,49 +175,77 @@ There are several data loading implementations of popular datasets across differ
 
 ## Frequently Asked Questions (FAQ)
 
-Q: What should I do if the existing set of DataPipes does not do what I need?
+<details>
+<summary>
+What should I do if the existing set of DataPipes does not do what I need?
+</summary>
 
-A: You can
+You can
 [implement your own custom DataPipe](https://pytorch.org/data/main/tutorial.html#implementing-a-custom-datapipe). If you
 believe your use case is common enough such that the community can benefit from having your custom DataPipe added to
 this library, feel free to open a GitHub issue. We will be happy to discuss!
+</details>
 
-Q: What happens when the `Shuffler` DataPipe is used with DataLoader?
+<details>
+<summary>
+What happens when the <code>Shuffler</code> DataPipe is used with DataLoader?
+</summary>
 
-A. In order to enable shuffling, you need to add a `Shuffler` to your DataPipe line. Then, by default, shuffling will
+In order to enable shuffling, you need to add a <code>Shuffler</code> to your DataPipe line. Then, by default, shuffling will
 happen at the point where you specified as long as you do not set `shuffle=False` within DataLoader.
+</details>
 
-Q: What happens when the `Batcher` DataPipe is used with DataLoader?
+<details>
+<summary>
+What happens when the <code>Batcher</code> DataPipe is used with DataLoader?
+</summary>
 
-A: If you choose to use `Batcher` while setting `batch_size > 1` for DataLoader, your samples will be batched more than
+If you choose to use <code>Batcher</code> while setting <code>batch_size > 1</code> for DataLoader, your samples will be batched more than
 once. You should choose one or the other.
+</details>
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -175,49 +175,77 @@ There are several data loading implementations of popular datasets across differ
...
+You can
 [implement your own custom DataPipe](https://pytorch.org/data/main/tutorial.html#implementing-a-custom-datapipe). If you
 believe your use case is common enough such that the community can benefit from having your custom DataPipe added to
 this library, feel free to open a GitHub issue. We will be happy to discuss!
+</details>
 
-Q: What happens when the `Shuffler` DataPipe is used with DataLoader?
+<details>
+<summary>
+What happens when the <code>Shuffler</code> DataPipe is used with DataLoader?
```
ejguan, line 191: Do we need to use `<code>` here? I think "`" should be sufficient.

README.md
```diff
@@ -175,49 +175,77 @@ There are several data loading implementations of popular datasets across differ
...
+<summary>
+How is multiprocessing handled with DataPipes?
+</summary>
 
-A: Multi-process data loading is still handled by the `DataLoader`, see the
+Multi-process data loading is still handled by the `DataLoader`, see the
 [DataLoader documentation for more details](https://pytorch.org/docs/stable/data.html#single-and-multi-process-data-loading).
 As of PyTorch version >= 1.12.0 (TorchData version >= 0.4.0), data sharding is automatically done for DataPipes within
-the `DataLoader` as long as a `ShardingFiler` DataPipe exists in your pipeline. Please see the
+the `DataLoader` as long as a `ShardingFilter` DataPipe exists in your pipeline. Please see the
```
ejguan, line 225: Good catch.

ejguan (commented):
Thank you. Overall look good to me with one nit comment.

README.md
lucaspar: Yes - I wasn't going to change the backticks, but if you check the preview, backticks do not render as <code> does between summary tags, unless we add a new line after <summary>.

The problem with that is that the new line is rendered and doesn't look good, so <code> was necessary to make the font monospace.

ejguan: I see. Makes sense.

README.md
```diff
@@ -175,49 +175,77 @@ There are several data loading implementations of popular datasets across differ
...
+</details>
 
-Q: What happens when the `Batcher` DataPipe is used with DataLoader?
+<details>
+<summary>
+What happens when the <code>Batcher</code> DataPipe is used with DataLoader?
+</summary>
 
-A: If you choose to use `Batcher` while setting `batch_size > 1` for DataLoader, your samples will be batched more than
+If you choose to use <code>Batcher</code> while setting <code>batch_size > 1</code> for DataLoader, your samples will be batched more than
```
ejguan, line 203: If this only applies to summary, let's change this part. Then this PR is good to go.

lucaspar: Good call, new commit added.

README.md
```diff
@@ -175,49 +175,77 @@ There are several data loading implementations of popular datasets across differ
...
+</details>
 
-Q: What happens when the `Shuffler` DataPipe is used with DataLoader?
+<details>
+<summary>
+What happens when the <code>Shuffler</code> DataPipe is used with DataLoader?
+</summary>
 
-A. In order to enable shuffling, you need to add a `Shuffler` to your DataPipe line. Then, by default, shuffling will
+In order to enable shuffling, you need to add a <code>Shuffler</code> to your DataPipe line. Then, by default, shuffling will
```
ejguan, line 194: Sorry. There is another case.

ejguan:
For the lint Error, could you please fix it by following the instruction [here](https://github.com/pytorch/data/blob/main/CONTRIBUTING.md#code-style)?

lucaspar:
It's passing locally, so I think it looks good if anyone can approve running the workflows.

PR #585, opened 2022-07-11
[DRAFT] Debug lines for cache

VitalyFedyunin:
* __->__ #585

```python
import torchtext

train_iter, val_iter, test_iter = \
        torchtext.datasets.IWSLT2017("data/IWSLT2017", language_pair=('en', 'de'))

print(next(iter(train_iter)))

```


VitalyFedyunin:
Bunch of debug lines to see what is going on with cache keys. We can potentially fix it with additional requirements to file_name_fn and changing `contact` to some sort of `group_by`. But it is still error prone and as soon as we find good caching strategy we should redo 1:M or kill it.

PR #658, opened 2022-07-14
Fix buffer overflow for unzip with columns_to_skip

ejguan:
When `unzip` has `columns_to_skip`, we should only create `ChildDataPipe` corresponding to `instance_ids`. And `Unzipper` should also handle mapping `instance_ids` to 0-based indices.

This is one of the root causes for the issue reported internally. See: https://fburl.com/2k0et1gv

### DIFF
--- a/torchdata/datapipes/iter/util/unzipper.py
+++ b/torchdata/datapipes/iter/util/unzipper.py
@@ -78,7 +78,7 @@ class _UnZipperIterDataPipe(_ForkerIterDataPipe):
     def get_next_element_by_instance(self, instance_id: int):
         r"""
         Note:
-            Each element returned from datapipe is required to be a sequnce that can
+            Each element returned from the source datapipe is required to be a sequnce that can
             be subscribed with a column index
         """
         for return_val in super().get_next_element_by_instance(instance_id):
--- a/torchdata/datapipes/iter/util/unzipper.py
+++ b/torchdata/datapipes/iter/util/unzipper.py
@@ -4,6 +4,7 @@
 # This source code is licensed under the BSD-style license found in the
 # LICENSE file in the root directory of this source tree.
 
+from collections import deque
 from typing import List, Optional, Sequence, TypeVar
 
 from torch.utils.data.datapipes.iter.combining import _ChildDataPipe, _ForkerIterDataPipe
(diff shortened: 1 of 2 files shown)


ejguan:
I will add tests tmrw

torchdata/datapipes/iter/util/unzipper.py
```diff
@@ -65,11 +65,15 @@ def __new__(
...
+        return [_ChildDataPipe(container, i) for i in range(len(instance_ids))]
 
 
 class _UnZipperIterDataPipe(_ForkerIterDataPipe):
+    def __init__(self, datapipe: IterDataPipe, instance_ids: List[int], buffer_size: int = 1000):
+        super().__init__(datapipe, len(instance_ids), buffer_size)
+        self.instance_ids = instance_ids
+
     def get_next_element_by_instance(self, instance_id: int):
         for return_val in super().get_next_element_by_instance(instance_id):
```
NivekT, line 78: super nit: Let's add a comment for the method here so that we know `return_val` is a tuple

NivekT (approved):
LGTM! Just need tests. Thanks!

ejguan:
This PR requires https://github.com/pytorch/pytorch/pull/81502 landed and released.

torchdata/datapipes/iter/util/unzipper.py
```diff
@@ -65,11 +66,46 @@ def __new__(
...
+        if IterDataPipe.getstate_hook is not None:
+            return IterDataPipe.getstate_hook(self)
+
+        state = (
+            self.main_datapipe,
+            self.num_instances,
+            self.buffer_size,
+            self.instance_ids,
+        )
+        return state
```
NivekT, line 97: I think it would be better to call `super().__getstate__`  and add `self.instance_ids` to it, because now any changes to `Forker`'s `__getstate__` will need to repeat here as well.

torchdata/datapipes/iter/util/unzipper.py
```diff
@@ -65,11 +66,46 @@ def __new__(
...
+            self.num_instances,
+            self.buffer_size,
+            self.instance_ids,
+        ) = state
+        self._datapipe_iterator = None
+        self.buffer = deque()
+        self.child_pointers = [0] * self.num_instances
+        self.slowest_ptr = 0
+        self.leading_ptr = 0
+        self.end_ptr = None
```
NivekT, line 111: Similar to my other comment, extracting `self.instance_ids` then call `super().__setstate__` is probably better?

NivekT (approved):

torchdata/datapipes/iter/util/unzipper.py
NivekT: I can also make this change in a separate PR if this is landing already.

ejguan: Makes sense. Will change it in this PR.

torchdata/datapipes/iter/util/unzipper.py
ejguan: Fixed

PR #664, opened 2022-07-19
Type annotate dataloader2 train loop

dahsh:
Summary: * Type annotate dataloader2 train loop

Differential Revision: D37971478


NivekT (approved):
LGTM!

PR #667, opened 2022-07-20
updated aisio datapipe

gaikwadabhishek:
aisio.py changes to reflect the new aistore package (recent restructuring of APIs)

### Changes
- Fixed references of the latest APIs in AIStore 
- using request raw stream in AISFileLoader instead of getting the whole object
- URL fix for AIStore example on the examples page
- Changes in AIStore unit test

Signed-off-by: Abhishek Gaikwad <gaikwadabhishek1997@gmail.com>

### DIFF
--- a/test/requirements_aistore.txt
+++ b/test/requirements_aistore.txt
@@ -1,2 +1,2 @@
+aistore >= 1.0.2
 pytest
-aistore == 0.9.2
--- a/torchdata/datapipes/iter/load/aisio.py
+++ b/torchdata/datapipes/iter/load/aisio.py
@@ -4,11 +4,9 @@
 # This source code is licensed under the BSD-style license found in the
 # LICENSE file in the root directory of this source tree.
 
-from io import BytesIO
 from typing import Iterator, Tuple
 
 from torchdata.datapipes import functional_datapipe
-
 from torchdata.datapipes.iter import IterDataPipe
 from torchdata.datapipes.utils import StreamWrapper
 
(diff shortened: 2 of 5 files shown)


test/requirements_aistore.txt
```diff
@@ -1,2 +1,2 @@
+aistore >= 1.0.2
```
VirrageS, line 1: I would set this to specific version. Otherwise, if something changes in the AIStore's API the tests will break.

test/requirements_aistore.txt
```diff
@@ -1,2 +1,2 @@
+aistore == 1.0.2
```
NivekT, line 1: Question: what makes you choose `==` instead of `>=`?

torchdata/datapipes/iter/load/aisio.py
```diff
@@ -17,19 +15,37 @@
...
+except (ImportError, AttributeError):
+    AIS_VERSION_CHECK = False
+
 
 def _assert_aistore() -> None:
     if not HAS_AIS:
         raise ModuleNotFoundError(
-            "Package `aistore` is required to be installed to use this datapipe."
-            "Please run `pip install aistore` or `conda install aistore` to install the package"
+            "Package `aistore` (>=1.0.2) is required to be installed to use this datapipe."
```
NivekT, line 35:
```suggestion
            "Package `aistore` (==1.0.2) is required to be installed to use this datapipe."
```
nit: to be consistent with `requirements_aistore.txt`. Let me know if there is a reason to be different.

torchdata/datapipes/iter/load/aisio.py
```diff
@@ -17,19 +15,37 @@
...
     HAS_AIS = True
+
 except ImportError:
     HAS_AIS = False
 
+try:
+    import aistore
+    from packaging.version import parse as parse_version
+
+    AIS_VERSION_CHECK = parse_version(aistore.__version__) >= parse_version("1.0.2")
```
NivekT, line 26:
```suggestion
    AIS_VERSION_CHECK = parse_version(aistore.__version__) == parse_version("1.0.2")
```
nit: to be consistent with `requirements_aistore.txt`. Let me know if there is a reason to be different.

torchdata/datapipes/iter/load/aisio.py
```diff
@@ -17,19 +15,37 @@
...
+            "Package `aistore` (>=1.0.2) is required to be installed to use this datapipe."
+            "Please run `pip install --upgrade aistore` or `conda install aistore` to install the package"
             "For more info visit: https://github.com/NVIDIA/aistore/blob/master/sdk/python/"
         )
 
 
+def _assert_aistore_version() -> None:
+    if not AIS_VERSION_CHECK:
+        raise ImportError(
+            "AIStore version >= 1.0.2 required"
```
NivekT, line 44: nit: same reason

NivekT (approved):
Thanks for the fixes! Mostly SGTM! The main question I have is whether we want users to have `aistore == 1.0.2` or `aistore >= 1.0.2` and keeping that consistent throughout out messages to users.

test/requirements_aistore.txt
NivekT: It seems like we are anticipating more changes to AIStore's API in the future? Is that correct?

Is the plan to increase the version number here and adjust the our usages of the APIs as AIStore's APIs change?

gaikwadabhishek: Hi @NivekT, all versions after 1.0.2 should work. We didn't want the torch/data pipeline ever to fail if in case we were trying some experimental changes on our side. 
But we will take care of it. We changed requirements to always fetch the latest version as we don't want people to spend time resolving version conflicts.

PR #670, opened 2022-07-21
Add Dataloader2 training loop example with torch text

dahsh:
Summary:
* Add the uscase and example of DataLoader2 with open source datasets/datapipes.
TorchText provides several standard NLP datasets; here we use its SST2 OSS dataset
 in the DataLoader2 train loop.

* We will have more examples to showcase the advantages:
(1) The usage of the DLv2 with popular open source dataset.
(2) Integrate datasets/datapipes with different reading service.
(3) Datapipe manipulation for example batch, collate, map.
(4) Dist usage and examples with features such as sharding_filter for the sharding feature.
(5) Eventually add those examples to the pytorch tutorials.

Reviewed By: ejguan

Differential Revision: D37938017

### DIFF
--- a/examples/dataloader2/train_loop_torchtext.py
+++ b/examples/dataloader2/train_loop_torchtext.py
@@ -1,0 +1,119 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
+# All rights reserved.
+#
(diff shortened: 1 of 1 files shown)


NivekT (approved):
LGTM! Thanks!

PR #675, opened 2022-07-23
fix newline windows issue in generate_csv

### LINKED ISSUE #674: generate_csv in tutorial won't work correctly on windows
### 📚 The doc issue

The main tutorial has a helper function to generate csvs: https://github.com/pytorch/data/blob/main/docs/source/tutorial.rst

This code works fine for linux, but with windows the open function requires newline to be specified, otherwise the csv writer will insert \r\r\n where the default text mode translates each \n into \r\n/. This creates a csv file with empty lines between the data lines and then causes an issue in the row_processor since the row is empty and row[0] throws list index out of range error.

This can be fixed by specificying newline such as:
`writer = csv.DictWriter(open(f"sample_data{file_label}.csv", "w", newline=''), fieldnames=fieldnames)`

instead of:
`writer = csv.DictWriter(open(f"sample_data{file_label}.csv", "w"), fieldnames=fieldnames)`

dbish:
Fixes #674 

### Changes

updated tutorial in documentation to specify the newline explitly
`writer = csv.DictWriter(open(f"sample_data{file_label}.csv", "w", newline=''), fieldnames=fieldnames)`

instead of:
`writer = csv.DictWriter(open(f"sample_data{file_label}.csv", "w"), fieldnames=fieldnames)`

## Testing
- tested by running the generate_csv function before and after on a windows machine, before was creating newlines between lines and throwing errors, after the created files don't have newlines and full tutorial works

### DIFF
--- a/docs/source/tutorial.rst
+++ b/docs/source/tutorial.rst
@@ -73,7 +73,7 @@ For this example, we will first have a helper function that generates some CSV f
 
     def generate_csv(file_label, num_rows: int = 5000, num_features: int = 20) -> None:
         fieldnames = ['label'] + [f'c{i}' for i in range(num_features)]
-        writer = csv.DictWriter(open(f"sample_data{file_label}.csv", "w"), fieldnames=fieldnames)
+        writer = csv.DictWriter(open(f"sample_data{file_label}.csv", "w", newline=''), fieldnames=fieldnames)
         writer.writeheader()
         for i in range(num_rows):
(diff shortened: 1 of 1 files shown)


dbish:
Same problem discussed here with this as a solution: https://stackoverflow.com/questions/3348460/csv-file-written-with-python-has-blank-lines-between-each-row

ejguan (approved):

PR #680, opened 2022-07-25
Cloud benchmarks

msaroufim:
Per our offline discussion, I am separating out the cloud part of #422 which will allow us to provision an AWS infra from the command line for offline benchmarks and potentially integrate into CI and release decisions.

In followup PRs, I will make the template more interesting and revisit the logger discussion so we can build this. (This PR is the middle node)
Screen Shot 2022-07-25 at 10 14 40 AM

Here's a test that the template works

Screen Shot 2022-07-25 at 10 15 13 AM


NivekT (approved):
LGTM! Thanks for creating this tool!

Nit: can you run pre-commit to fix the lint issue?

PR #687, opened 2022-07-25
Fix target folder for official release

ejguan:
During preparing the minor release, the document folder is changed from `0.4.0` to `0.4`. For the official release, the workflow to extract target folder from tag for the official release should be modified correspondingly.

### DIFF
--- a/.github/workflows/_build_test_upload.yml
+++ b/.github/workflows/_build_test_upload.yml
@@ -432,6 +432,9 @@ jobs:
           elif [[ $TARGET_FOLDER == tags/* ]]; then
             TARGET_FOLDER=${TARGET_FOLDER:5}
           elif [[ $TARGET_FOLDER == v* ]] && [[ ${{ inputs.pre_dev_release }} == false ]]; then
+            if [[ $TARGET_FOLDER == v*.*.* ]]; then
+              TARGET_FOLDER=${TARGET_FOLDER%.*}
+            fi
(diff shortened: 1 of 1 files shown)


.github/workflows/_build_test_upload.yml
```diff
@@ -432,6 +432,9 @@ jobs:
           elif [[ $TARGET_FOLDER == tags/* ]]; then
             TARGET_FOLDER=${TARGET_FOLDER:5}
           elif [[ $TARGET_FOLDER == v* ]] && [[ ${{ inputs.pre_dev_release }} == false ]]; then
+            if [[ $TARGET_FOLDER == v*.*.* ]]; then
+              TARGET_FOLDER=${TARGET_FOLDER%.*}
+            fi
```
ejguan, line 437: Remove the trailing `.0` from the tag.
`v0.4.0` -> `v0.4`

NivekT (approved):

.github/workflows/_build_test_upload.yml
NivekT: Question: so we will have a single target folder of `v0.4` instead of two like `v0.4.0` and `v0.4.1`?

ejguan: Yeah. This is because minor release should not contain major feature changes unless significant bugs. So, there is no need to create a separate folder for docs.

PR #692, opened 2022-07-26
Enable release on Mac arm64

### LINKED ISSUE #676: osx-arm64 build missing from Anaconda pytorch channel
### 🐛 Describe the bug

Mac users on Apple silicon (M1, M2) cannot `conda install torchdata -c pytorch`, receiving "PackagesNotFoundError: The following packages are not available from current channels: -torchdata". This appears to be because torchdata was not built, packaged, and/or submitted to the pytorch channel. Is that something that can be done for future releases?

References:
* [torchdata](https://anaconda.org/pytorch/torchdata) shows only linux-64, osx-64, and win-64
* [torchtext](https://anaconda.org/pytorch/torchtext) for comparison, showing additional entries, including osx-arm64.

Many thanks.

### Versions

Any version.

ejguan:
Fixes #676

### Changes

- Update the workflow to compile AWSSDK and integrated on arm64 macos
  - Use conda to install python since this [action](https://github.com/actions/setup-python) doesn't work with arm64b macos
  - Exclude python 3.7 from arm64 macos due to the pre-compiled python binary is not available either on conda or python official website
  - Use `bash -l {0}` to make sure `conda` is available across steps
- Please check the manually triggered workflow: https://github.com/pytorch/data/runs/7528349378?check_suite_focus=true
  - Wheels: https://github.com/pytorch/data/runs/7528349378?check_suite_focus=true#step:4:1
  - Conda packages: https://github.com/pytorch/data/runs/7528462809?check_suite_focus=true#step:5:1

### DIFF
--- a/.github/workflows/_build_test_upload.yml
+++ b/.github/workflows/_build_test_upload.yml
@@ -62,22 +62,46 @@ jobs:
           - macos-latest
           - ubuntu-latest
           - windows-latest
+          - macos-m1-12
         python-version:
           - 3.7
           - 3.8
           - 3.9
           - "3.10"
+        # Python 3.7 is not officially provided on Apple Silicon
+        exclude:
+          - os: macos-m1-12
+            python-version: 3.7
     steps:
-      - name: Setup Python ${{ matrix.python-version }}
-        if: ${{ ! startsWith( matrix.os, 'ubuntu' ) }}
-        uses: actions/setup-python@v2
-        with:
-          python-version: ${{ matrix.python-version }}
       - name: Checkout Source Repository
         uses: actions/checkout@v2
         with:
           ref: ${{ inputs.branch }}
           submodules: recursive
+      - name: Setup Python ${{ matrix.python-version }}
+        if: ${{ startsWith( matrix.os, 'windows' ) }}
+        uses: actions/setup-python@v2
+        with:
+          python-version: ${{ matrix.python-version }}
+      - name: Install Conda on MacOS
+        if: ${{ startsWith( matrix.os, 'macos' ) }}
+        shell: bash -l {0}
+        run: |
+          mkdir -p ~/miniconda3
+          if ${{ startsWith( matrix.os, 'macos-m1' ) }}; then
+            curl https://repo.anaconda.com/miniconda/Miniconda3-latest-MacOSX-arm64.sh -o ~/miniconda3/miniconda.sh
+          else
+            curl https://repo.anaconda.com/miniconda/Miniconda3-latest-MacOSX-x86_64.sh -o ~/miniconda3/miniconda.sh
+          fi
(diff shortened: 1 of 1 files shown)


.github/workflows/_build_test_upload.yml
```diff
@@ -62,22 +62,46 @@ jobs:
...
       - name: Checkout Source Repository
         uses: actions/checkout@v2
         with:
           ref: ${{ inputs.branch }}
           submodules: recursive
+      - name: Setup Python ${{ matrix.python-version }}
+        if: ${{ startsWith( matrix.os, 'windows' ) }}
+        uses: actions/setup-python@v2
+        with:
+          python-version: ${{ matrix.python-version }}
```
NivekT, line 85: QQ: Was windows already setting up Python before? Was this not needed until now?

NivekT (approved):
LGTM! Just some questions:
1. `S3` is not compatible with Windows, correct?
2. We are building the same way regardless if it is MacOS x86_64 or arm64?

.github/workflows/_build_test_upload.yml
```diff
@@ -62,22 +62,46 @@ jobs:
...
           - 3.8
           - 3.9
           - "3.10"
+        # Python 3.7 is not officially provided on Apple Silicon
+        exclude:
+          - os: macos-m1-12
+            python-version: 3.7
     steps:
-      - name: Setup Python ${{ matrix.python-version }}
-        if: ${{ ! startsWith( matrix.os, 'ubuntu' ) }}
```
ejguan, line 72: Previously, we excluded `ubuntu` from this step because `ubuntu` used docker. Now, since `macos` would use conda to install python, `windows` becomes the only platform that would use this step.
cc: @NivekT

.github/workflows/_build_test_upload.yml
ejguan: https://github.com/pytorch/data/pull/692/files#r930410707

ejguan:
> 1. `S3` is not compatible with Windows, correct?

Actually, it is compatible. And, we do compile AWSSDK for PyPI wheels. However, for conda packages, I haven't figured out how to integrate AWSSDK with torchdata. I decide not to invest more time on it unless there are more users' request

> 2. We are building the same way regardless if it is MacOS x86_64 or arm64?

Yea, the same workflow.

.github/workflows/_build_test_upload.yml
```diff
@@ -62,22 +62,46 @@ jobs:
           - macos-latest
           - ubuntu-latest
           - windows-latest
+          - macos-m1-12
```
adamjstewart, line 65: Is this an internal Meta self-hosted runner or does GitHub Actions finally have public M1 runners?

ejguan: Our self-hosted runner

PR #695, opened 2022-07-27
Update version compatibility table

ejguan:
Update version matrix for 0.4.1

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -110,7 +110,7 @@ def _get_requirements():
 
 
 # Use new version of torch on main branch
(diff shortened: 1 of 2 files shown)


setup.py
```diff
@@ -110,7 +110,7 @@ def _get_requirements():
 
 
 # Use new version of torch on main branch
-pytorch_package_dep = "torch>1.11.0"
+pytorch_package_dep = "torch>1.12"
```
ejguan, line 113: It was incorrect before this PR, even though it was still working.

NivekT (approved):
Thanks for catching that.

PR #700, opened 2022-07-29
[DataLoader2] Implementing single iterator constraint

NivekT:
* __->__ #700

Changes:
* Imposing the single iterator constraint on `DataLoader2` by having its `__iter__` method returns a wrapper that can track if an iterator is valid or not
* Previously, `__iter__` returns `self`, but now it will return an instance of `_DataLoader2Iterator`
* Aside from the behavior related to reset (see below), we expect other behaviors to stay the same.

Prior behavior:
```python
dl = DataLoader2(IterableWrapper(range(10)))
it1 = iter(dl)
print(next(it1))  # 0
it2 = iter(dl)  # No reset here
print(next(it2))  # 1
print(next(it1))  # 2
```

New behavior with this PR:
```python
dl = DataLoader2(IterableWrapper(range(10)))
it1 = iter(dl)
print(next(it1))  # 0
it2 = iter(dl)  # DataLoader2 resets with the creation of a new iterator
print(next(it2))  # 0
print(next(it1))  # Raises exception, since it1 is no longer valid
```

Differential Revision: [D38262867](https://our.internmc.facebook.com/intern/diff/D38262867)

### DIFF
--- a/test/test_dataloader2.py
+++ b/test/test_dataloader2.py
@@ -9,6 +9,8 @@ import pickle
 import unittest
 from unittest import TestCase
 
+from torch.utils.data.graph import DataPipe
+
 from torchdata.dataloader2 import (
     DataLoader2,
     MultiProcessingReadingService,
@@ -20,14 +22,14 @@ from torchdata.datapipes.iter import IterableWrapper, IterDataPipe
 
 
 class TestReadingService(ReadingServiceInterface):
-    def initialize(self, dp: IterDataPipe) -> IterDataPipe:
+    def initialize(self, dp: DataPipe) -> DataPipe:
         return dp
 
 
 class DataLoader2Test(TestCase):
     def test_dataloader2(self) -> None:
         test_data_pipe = IterableWrapper(range(3))
-        data_loader = DataLoader2(datapipe=test_data_pipe)
+        data_loader: DataLoader2 = DataLoader2(datapipe=test_data_pipe)
 
         expected_batch = 0
         for batch in iter(data_loader):
@@ -36,12 +38,12 @@ class DataLoader2Test(TestCase):
 
     def test_dataloader2_shutdown(self) -> None:
         test_data_pipe = IterableWrapper(range(3))
-        data_loader = DataLoader2(datapipe=test_data_pipe)
+        data_loader: DataLoader2 = DataLoader2(datapipe=test_data_pipe)
         data_loader.shutdown()
 
     def test_dataloader2_state_dict(self) -> None:
         test_data_pipe = IterableWrapper(range(3))
-        data_loader = DataLoader2(datapipe=test_data_pipe)
+        data_loader: DataLoader2 = DataLoader2(datapipe=test_data_pipe)
 
         state = data_loader.state_dict()
         self.assertIsNotNone(state)
@@ -52,7 +54,7 @@ class DataLoader2Test(TestCase):
     def test_dataloader2_reading_service(self) -> None:
         test_data_pipe = IterableWrapper(range(3))
         reading_service = TestReadingService()
-        data_loader = DataLoader2(datapipe=test_data_pipe, reading_service=reading_service)
+        data_loader: DataLoader2 = DataLoader2(datapipe=test_data_pipe, reading_service=reading_service)
 
         expected_batch = 0
         for batch in iter(data_loader):
@@ -62,7 +64,7 @@ class DataLoader2Test(TestCase):
     def test_dataloader2_multi_process_reading_service(self) -> None:
         test_data_pipe = IterableWrapper(range(3))
         reading_service = MultiProcessingReadingService()
-        data_loader = DataLoader2(datapipe=test_data_pipe, reading_service=reading_service)
+        data_loader: DataLoader2 = DataLoader2(datapipe=test_data_pipe, reading_service=reading_service)
 
         expected_batch = 0
         for batch in iter(data_loader):
@@ -72,7 +74,7 @@ class DataLoader2Test(TestCase):
     def test_dataloader2_load_state_dict(self) -> None:
         test_data_pipe = IterableWrapper(range(3))
         reading_service = TestReadingService()
-        data_loader = DataLoader2(datapipe=test_data_pipe, reading_service=reading_service)
+        data_loader: DataLoader2 = DataLoader2(datapipe=test_data_pipe, reading_service=reading_service)
 
         batch = next(iter(data_loader))
         self.assertEqual(batch, 0)
@@ -84,7 +86,7 @@ class DataLoader2Test(TestCase):
         data_loader.shutdown()
 
         test_data_pipe_2 = IterableWrapper(range(5))
-        restored_data_loader = DataLoader2(datapipe=test_data_pipe_2, reading_service=reading_service)
+        restored_data_loader: DataLoader2 = DataLoader2(datapipe=test_data_pipe_2, reading_service=reading_service)
         restored_data_loader.load_state_dict(state)
 
         restored_data_loader_datapipe = restored_data_loader.datapipe
@@ -103,6 +105,27 @@ class DataLoader2Test(TestCase):
         )
 
         restored_data_loader.shutdown()
+
+    def test_dataloader2_reset(self) -> None:
+
+        test_data_pipe = IterableWrapper(range(10))
+        reading_services = [None, TestReadingService(), MultiProcessingReadingService(num_workers=1)]
+
+        for reading_service in reading_services:
+            data_loader: DataLoader2 = DataLoader2(datapipe=test_data_pipe, reading_service=reading_service)
+
+            # Functional Test: Ensure multiple sequential reads of DL2 is possible
+            self.assertEqual(list(range(10)), list(data_loader))  # [0, ..., 10]`
+            self.assertEqual(list(range(10)), list(data_loader))  # [0, ..., 10]`
+            self.assertEqual(list(range(10)), list(data_loader))  # [0, ..., 10]`
+
+            # Functional Test: Ensure that the creation of a new iterator invalidates the old one
+            it1 = iter(data_loader)
+            self.assertEqual(0, next(it1))
+            it2 = iter(data_loader)
+            self.assertEqual(0, next(it2))
+            with self.assertRaisesRegex(RuntimeError, "iterator has been invalidated"):
+                print(next(it1))
 
 
 class DataLoader2ConsistencyTest(TestCase):
--- a/torchdata/dataloader2/dataloader2.py
+++ b/torchdata/dataloader2/dataloader2.py
@@ -43,6 +43,38 @@ class ConcurrencySpec:
     persistent_workers: bool = False
 
 
+class _DataLoader2Iterator(Iterator):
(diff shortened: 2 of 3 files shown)


NivekT:
@nivekt has imported this pull request.  If you are a Facebook employee, you can view this diff [on Phabricator](https://www.internalfb.com/diff/D38262867).

torchdata/dataloader2/dataloader2.py
```diff
@@ -88,25 +121,16 @@ def __iter__(self) -> Iterator[T_co]:
...
-            raise StopIteration
-        try:
-            return next(self._datapipe_iter)  # type: ignore[arg-type]
-        except PauseIteration:
-            raise StopIteration
-        except StopIteration:
-            if self.reading_service is not None:
-                self.reading_service.finalize_iteration()
-            self._reset_iter = True
-            raise
```
NivekT, line 107: Since we are returning a wrapper in `__iter__`, I think it would be better for users to not call `__next__` directly on `DataLoader2`, but rather, call `next` on the `iterator` that created from `__iter__`. 

This should prevent the situation where `next` can be called on both `DataLoader2` and the `iterator` that has been created.

Thoughts?

torchdata/dataloader2/dataloader2.py
```diff
@@ -43,6 +43,38 @@ class ConcurrencySpec:
     persistent_workers: bool = False
 
 
+class _DataLoader2Iterator(Iterator):
+    def __init__(self, dataloader: "DataLoader2", iterator_id: int):
+        self.dataloader = dataloader
+        self.iterator_id = iterator_id
+
+    def __next__(self) -> T_co:
```
NivekT, line 51: Most of this came from `DataLoader2`

torchdata/dataloader2/dataloader2.py
```diff
@@ -43,6 +43,38 @@ class ConcurrencySpec:
...
+                "This may be caused multiple references to the same DataLoader2. "
+                "For feedback regarding this single iterator per DataLoader2 constraint, feel free "
+                "to comment on this issue: https://github.com/pytorch/data/issues/45."
+            )
+
+    def __getattr__(self, name):
+        """
+        To delegate operations such as `limit`, `pause`, `resume`, `__getstate__`
+        """
+        return getattr(self.dataloader, name)
```
ejguan, line 75: Why not directly delegate to `datapipe_iter`?

torchdata/dataloader2/dataloader2.py
```diff
@@ -43,6 +43,38 @@ class ConcurrencySpec:
...
 
 
+class _DataLoader2Iterator(Iterator):
+    def __init__(self, dataloader: "DataLoader2", iterator_id: int):
+        self.dataloader = dataloader
+        self.iterator_id = iterator_id
+
+    def __next__(self) -> T_co:
+        if self.iterator_id == self.dataloader.valid_iterator_id:
+            self.dataloader._reset_iter = True
```
ejguan, line 53: Why doe we need to `reset_iter` every single `next`? IIUC, you should implement a `__iter__` function with such logic and return `self`.

torchdata/dataloader2/dataloader2.py
```diff
@@ -88,25 +121,16 @@ def __iter__(self) -> Iterator[T_co]:
...
-            self._reset_iter = True
-            raise
+        self.valid_iterator_id = 0 if self.valid_iterator_id is None else self.valid_iterator_id + 1
+        return _DataLoader2Iterator(self, self.valid_iterator_id)
 
     def __getattr__(self, name: str) -> Any:
+        """
+        Delegate methods (e.g. `limit`, `pause`, `resume`, etc) to the iterator
+        created by the ReadingService and DataPipe.
+        """
```
ejguan, line 133: If we directly delegate to `datapipe_iter` from `dataloader_iter`, we don't need this function.

torchdata/dataloader2/dataloader2.py
```diff
@@ -43,6 +43,38 @@ class ConcurrencySpec:
...
+                "This iterator has been invalidated because another iterator has been created "
+                "from the same DataLoader2.\n"
+                "This may be caused multiple references to the same DataLoader2. "
+                "For feedback regarding this single iterator per DataLoader2 constraint, feel free "
+                "to comment on this issue: https://github.com/pytorch/data/issues/45."
+            )
+
+    def __getattr__(self, name):
+        """
+        To delegate operations such as `limit`, `pause`, `resume`, `__getstate__`
```
ejguan, line 73: And, please remove examples of `limit` and `pause` for now because those APIs are tied to internal services

torchdata/dataloader2/dataloader2.py
```diff
@@ -88,25 +121,16 @@ def __iter__(self) -> Iterator[T_co]:
                 self.reading_service.initialize_iteration()
 
             self._datapipe_iter = iter(self.datapipe)
```
ejguan, line 123: There is a problem that `self._datapipe_iter` is never updated when we want to have new `_DataLoader2Iterator`

ejguan (commented):
Let me know if those comments make sense to you.

test/test_dataloader2.py
```diff
@@ -104,6 +106,27 @@ def test_dataloader2_load_state_dict(self) -> None:
...
+            # Functional Test: Ensure multiple sequential reads of DL2 is possible
+            self.assertEqual(list(range(10)), list(data_loader))  # [0, ..., 10]`
+            self.assertEqual(list(range(10)), list(data_loader))  # [0, ..., 10]`
+            self.assertEqual(list(range(10)), list(data_loader))  # [0, ..., 10]`
+
+            # Functional Test: Ensure that the creation of a new iterator invalidates the old one
+            it1 = iter(data_loader)
+            self.assertEqual(0, next(it1))
+            it2 = iter(data_loader)
+            self.assertEqual(0, next(it2))
```
ejguan, line 126: Could you please run `next(it1)` more times before creating `it2`? I am worry about the result would always be `0`.
Same to `it2`, could you please run `list(it2)` to make sure the result is expected?

torchdata/dataloader2/dataloader2.py
NivekT: I am fine with this change, but I would flag that methods such as `state_dict` and `shutdown` will not be delegated to `DataLoader2`.

torchdata/dataloader2/dataloader2.py
NivekT: `self.dataloader._reset_iter` determines whether reset needs to happen when `iter(DataLoader2)` is called, so this is not "reset_iter" everything single `next`. 

Ideally, `self.dataloader._reset_iter = True` should only be set once the first time `next` is called. We can choose to set it once when `iter` is called (instead of the first time `next` is called). but that may lead to some unnecessary reset of `DataLoader2`.

torchdata/dataloader2/dataloader2.py
NivekT: I think it is? When `__iter__` of `DataLoader2` is called, it first checks if `self._reset_iter == True` (an indicator whether an active iterator has been read at least once). If that is `True`, `self._datapipe_iter` gets updated and `self._reset_iter` is set to `False`. 

Let's discuss offline to make sure I know what you mean.

torchdata/dataloader2/dataloader2.py
NivekT: Also, do we want to allow users to do `dl2.limit()`, ``dl2.resume()`, and etc? Or we want them to always invoke those from the iterator?

If we want the former, we will need to keep this method.

torchdata/dataloader2/dataloader2.py
ejguan: I see. Then, the question is do we need `_reset_iter`?
When I created this `_reset_iter` flag, it's used for mini-epoch. Only when `PauseIteration` is raised, we don't want to reset datapipe iterator but re-use the datapipe iterator since it's not depleted yet. Could we still achieve this behavior?

torchdata/dataloader2/dataloader2.py
ejguan: cc: @Miiira

NivekT: I think invoking from one single place (i.e. iterator) rather than multiple places is better and less error-prone.

The internal test is calling `.resume` on the iterator, so I think it is fine to remove the API from the `DataLoader2` class.

NivekT:
@nivekt has imported this pull request.  If you are a Facebook employee, you can view this diff [on Phabricator](https://www.internalfb.com/diff/D38262867).

torchdata/dataloader2/dataloader2.py
NivekT: Discussed offline

torchdata/dataloader2/dataloader2.py
Miiira: Yes we only need `limit`, `resume` on iterator. I'm good with removing this from `DataLoader2`. Maybe we also want to look at Lightning DataModule `train_dataloader` return type with this change

torchdata/dataloader2/dataloader2.py
```diff
@@ -43,6 +43,42 @@ class ConcurrencySpec:
...
+                "to comment on this issue: https://github.com/pytorch/data/issues/45."
+            )
+
+    def __getattr__(self, name):
+        """
+        To delegate operations to ``dataloader._datapipe_iter``.
+        """
+        if self.dataloader._datapipe_iter is None:
+            raise AttributeError
+        return getattr(self.dataloader_datapipe_iter, name)
```
Miiira, line 79: self.dataloader._datapipe_iter

torchdata/dataloader2/dataloader2.py
Miiira: Same question as @ejguan, are we still allowing resuing iterator when receiving a `PauseIteration`? We can use internal unit tests to verify that. 
For context, our reading service may need to support both mini-epoch and real multiple epochs training soon. But that should be a reading-service update to raise Pause/StopIteration based on some flag.

NivekT: Yes, I expect the iterator can be used as long as a new one isn't created. We should verify with internal test.

NivekT:
@nivekt has imported this pull request.  If you are a Facebook employee, you can view this diff [on Phabricator](https://www.internalfb.com/diff/D38262867).

test/test_dataloader2.py
```diff
@@ -19,15 +21,31 @@
...
+    def __iter__(self):
+        self.it = iter(self.dp)
+        return self
+
+    def __next__(self):
+        return next(self.it)
+
+    @staticmethod
+    def return_one():
+        return 1
```
NivekT, line 37: A method that `dataloader._datapipe_iter` will have

NivekT (commented):
I updated this PR (one fix) and added a test to make through attributes/methods are being passed through. I will be looking at internal tests again shortly.

NivekT:
@nivekt has imported this pull request.  If you are a Facebook employee, you can view this diff [on Phabricator](https://www.internalfb.com/diff/D38262867).

PR #706, opened 2022-08-01
ci: Don't install things to home directory, add workflow to test release pipelines

seemethere:
Noticed we were getting extra conda packges from the home directory on
the M1 runners. Making this PR to reduce the amount that we're using the
home directory.

Will probably make it unwritable in the future so this is a pre-emptive
step to that.

Also adds a workflow to test the release pipelines without actually releasing a new binary.

Signed-off-by: Eli Uriegas <eliuriegas@fb.com>

### DIFF
--- a/.github/workflows/_build_test_upload.yml
+++ b/.github/workflows/_build_test_upload.yml
@@ -32,7 +32,6 @@ on:
 
 jobs:
   get_release_type:
-    if: inputs.branch != ''
     runs-on: ubuntu-latest
     outputs:
       type: ${{ steps.get_release_type.outputs.type }}
@@ -43,13 +42,8 @@ jobs:
             RELEASE_TYPE=official
           elif [[ ${{ inputs.branch }} == release/* ]] && [[ ${{ inputs.pre_dev_release }} == true ]]; then
             RELEASE_TYPE=test
-          elif [[ ${{ inputs.branch }} == main ]] && [[ ${{ inputs.pre_dev_release }} == true ]]; then
-            RELEASE_TYPE=nightly
           else
-            echo "Invalid combination of inputs!"
-            echo "  branch: ${{ inputs.branch }}"
-            echo "  pre_dev_release: ${{ inputs.pre_dev_release }}"
-            exit 1
+            RELEASE_TYPE=nightly
           fi
           echo "Release Type: $RELEASE_TYPE"
           echo "::set-output name=type::$RELEASE_TYPE"
--- a/.github/workflows/_build_test_upload.yml
+++ b/.github/workflows/_build_test_upload.yml
@@ -55,8 +55,6 @@ jobs:
     needs: get_release_type
     runs-on: ${{ matrix.os }}
     container: ${{ startsWith( matrix.os, 'ubuntu' ) && 'pytorch/manylinux-cpu' || null }}
-    env:
-      MINICONDA_INSTALL_PATH_MACOS: ${{ runner.temp }}/miniconda
     strategy:
       fail-fast: false
       matrix:
@@ -85,6 +83,10 @@ jobs:
         uses: actions/setup-python@v2
         with:
           python-version: ${{ matrix.python-version }}
+      - name: Add temp runner environment variables
+        shell: bash -l {0}
+        run: |
+          echo "MINICONDA_INSTALL_PATH_MACOS=${RUNNER_TEMP}/miniconda" >> "${GITHUB_ENV}"
       - name: Install Conda on MacOS
         if: ${{ startsWith( matrix.os, 'macos' ) }}
         shell: bash -l {0}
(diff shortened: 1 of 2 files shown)


seemethere:
@ejguan how would I go about testing this?

NivekT (approved):

.github/workflows/_build_test_upload.yml
```diff
@@ -87,15 +89,15 @@ jobs:
...
           if ${{ startsWith( matrix.os, 'macos-m1' ) }}; then
-            curl https://repo.anaconda.com/miniconda/Miniconda3-latest-MacOSX-arm64.sh -o ~/miniconda3/miniconda.sh
+            curl https://repo.anaconda.com/miniconda/Miniconda3-latest-MacOSX-arm64.sh -o "${MINICONDA_INSTALL_PATH_MACOS}/miniconda.sh"
           else
-            curl https://repo.anaconda.com/miniconda/Miniconda3-latest-MacOSX-x86_64.sh -o ~/miniconda3/miniconda.sh
+            curl https://repo.anaconda.com/miniconda/Miniconda3-latest-MacOSX-x86_64.sh -o "${MINICONDA_INSTALL_PATH_MACOS}/miniconda.sh"
           fi
-          bash ~/miniconda3/miniconda.sh -b -u -p ~/miniconda3
+          bash "${MINICONDA_INSTALL_PATH_MACOS}/miniconda.sh" -b -u -p "${MINICONDA_INSTALL_PATH_MACOS}"
           rm -rf ~/miniconda3/miniconda.sh
```
ejguan, line 99:
```suggestion
          rm -rf "${MINICONDA_INSTALL_PATH_MACOS}/miniconda.sh"
```

.github/workflows/_build_test_upload.yml
```diff
@@ -294,26 +299,26 @@ jobs:
         uses: conda-incubator/setup-miniconda@v2
         with:
           python-version: ${{ matrix.python-version }}
-          activate-environment: conda_build_env
+          activate-environment: ${{ env.CONDA_ENV_PATH }}
```
ejguan, line 302: I am not sure about using path here. You might want to add a `CONDA_ENV_NAME` as `conda_build_env`. And, define `CONDA_ENV_PATH` by `"${{ MINICONDA_INSTALL_PATH_MACOS }}/${{ CONDA_ENV_NAME }}"`

.github/workflows/_build_test_upload.yml
```diff
@@ -294,26 +299,26 @@ jobs:
...
           if ${{ startsWith( matrix.os, 'macos-m1' ) }}; then
-            curl https://repo.anaconda.com/miniconda/Miniconda3-latest-MacOSX-arm64.sh -o ~/miniconda3/miniconda.sh
+            curl https://repo.anaconda.com/miniconda/Miniconda3-latest-MacOSX-arm64.sh -o "${MINICONDA_INSTALL_PATH_MACOS}/miniconda.sh"
           else
-            curl https://repo.anaconda.com/miniconda/Miniconda3-latest-MacOSX-x86_64.sh -o ~/miniconda3/miniconda.sh
+            curl https://repo.anaconda.com/miniconda/Miniconda3-latest-MacOSX-x86_64.sh -o "${MINICONDA_INSTALL_PATH_MACOS}/miniconda.sh"
           fi
-          bash ~/miniconda3/miniconda.sh -b -u -p ~/miniconda3
+          bash "${MINICONDA_INSTALL_PATH_MACOS}/miniconda.sh" -b -u -p "${MINICONDA_INSTALL_PATH_MACOS}"
           rm -rf ~/miniconda3/miniconda.sh
```
ejguan, line 314: Ditto

ejguan (commented):
Thank you for adding this PR. To test it, you have to push your changes to a branch on `torchdata` repo rather than you forked branch.
And, you might need to change all `ref`s ([here](https://github.com/pytorch/data/blob/a7745b9865ed590c252072c51026400ae64a656f/.github/workflows/_build_test_upload.yml#L79), [here](https://github.com/pytorch/data/blob/a7745b9865ed590c252072c51026400ae64a656f/.github/workflows/_build_test_upload.yml#L290)) from this workflow to your new branch name.
Then, you can manually trigger the nightly release workflow from https://github.com/pytorch/data/actions/workflows/nightly_release.yml.

ejguan:
@seemethere IIRC, m1 runner has `conda` built already. Could we just rely on the installed conda without install conda again to make sure it doesn't pollute the ci runner?

.github/workflows/_build_test_upload.yml
seemethere: Technically this should work as defined by their definitions for this particular variable:

```
  activate-environment:
    description:
      "Environment name (or path) to activate on all shells. Default is `test`
      which will be created in `$CONDA/envs/test`. If an empty string is used,
      no environment is activated by default (For `base` activation see the
      `auto-activate-base` option). If the environment does not exist, it will
      be created and activated. If `environment-file` is used and you want that
      to be the environment used, you need to explicitely provide the name of
      that environment on `activate-environment`. If using sh/bash/cmd.exe
      shells please read the IMPORTANT! section on the README.md! to properly
      activate conda environments on these shells."
    required: false
    default: "test"
```

From: https://github.com/conda-incubator/setup-miniconda/blob/main/action.yml#L79-L91

seemethere:
> @seemethere IIRC, m1 runner has `conda` built already. Could we just rely on the installed conda without install conda again to make sure it doesn't pollute the ci runner?

We could, but installing conda on its own in runner temp should be fine for now, if eventually you want to migrate to the local conda you can but my scope for this PR is just trying to reduce the usage of the home directory here

PR #714, opened 2022-08-04
Add benchmark from torchvision training references

NicolasHug:
Towards https://github.com/pytorch/data/issues/416

This is a modified and simplified version of the torchvision classification training reference that provides:

-   Distributed Learning (DDP) vs 1-GPU training
-    Datapipes (with DataLoader or torchdata.dataloader2) vs Iterable datasets (non-DP) vs MapStyle Datasets
-   Full training procedure or Data-loading only (with or without transforms) or Model training only (generating fake datasets)
-   Timing of data-loading vs model training
-   any classification model from torchvision

I removed a lot of non-essential features from the original reference, but I can simplify further. Typically I would expect the `MetricLogger` to disappear, or be trimmed down to its most essential bits.

### DIFF
--- a/benchmarks/torchvision_classification/train.py
+++ b/benchmarks/torchvision_classification/train.py
@@ -1,0 +1,346 @@
+import datetime
+import os
+import time
+import warnings
+
+import helpers
+import presets
+import torch
+import torch.utils.data
+import torchvision
+import utils
+from torch import nn
+from torchdata.dataloader2 import DataLoader2, MultiProcessingReadingService, adapter
+
+
+def train_one_epoch(model, criterion, optimizer, data_loader, device, epoch, args):
+    model.train()
+    metric_logger = utils.MetricLogger(delimiter="  ")
+    metric_logger.add_meter("lr", utils.SmoothedValue(window_size=1, fmt="{value}"))
+    metric_logger.add_meter("img/s", utils.SmoothedValue(window_size=10, fmt="{value}"))
+
+    header = f"Epoch: [{epoch}]"
+    for i, (image, target) in enumerate(metric_logger.log_every(data_loader, args.print_freq, header)):
+        if args.data_loading_only:
+            continue
+
+        start_time = time.time()
+        image, target = image.to(device), target.to(device)
+
+        output = model(image)
+        loss = criterion(output, target)
+
+        optimizer.zero_grad()
+        loss.backward()
+        optimizer.step()
+
+        acc1, acc5 = utils.accuracy(output, target, topk=(1, 5))
+        batch_size = image.shape[0]
+        metric_logger.update(loss=loss.item(), lr=optimizer.param_groups[0]["lr"])
+        metric_logger.meters["acc1"].update(acc1.item(), n=batch_size)
+        metric_logger.meters["acc5"].update(acc5.item(), n=batch_size)
+        metric_logger.meters["img/s"].update(batch_size / (time.time() - start_time))
+
+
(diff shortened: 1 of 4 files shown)


benchmarks/torchvision_classification/train.py
```diff
@@ -0,0 +1,346 @@
...
+            "Setting the world size to 1 is always a safe bet."
+        )
+
+    metric_logger.synchronize_between_processes()
+
+    print(f"{header} Acc@1 {metric_logger.acc1.global_avg:.3f} Acc@5 {metric_logger.acc5.global_avg:.3f}")
+    return metric_logger.acc1.global_avg
+
+
+def create_data_loaders(args):
```
NivekT, line 90: We likely have to accept path to the datasets as an argument.

benchmarks/torchvision_classification/train.py
```diff
@@ -0,0 +1,346 @@
...
+            raise ValueError("DataLoader2 only works with datapipes.")
+
+        # Note: we are batching and collating here *after the transforms*, which is consistent with DLV1.
+        # But maybe it would be more efficient to do that before, so that the transforms can work on batches??
+
+        train_dataset = train_dataset.batch(args.batch_size, drop_last=True).collate()
+        train_data_loader = DataLoader2(
+            train_dataset,
+            datapipe_adapter_fn=adapter.Shuffle(),
+            reading_service=MultiProcessingReadingService(num_workers=args.workers),
```
NivekT, line 175: Note that `MultiProcessingReadingService` will be updated to use a slightly different backend (i.e. what is currently `PrototypeMultiProcessingReadingService`.

I'm going to look into that part to see if there is any issue.

benchmarks/torchvision_classification/utils.py
```diff
@@ -0,0 +1,284 @@
...
+        return self.delimiter.join(loss_str)
+
+    def synchronize_between_processes(self):
+        for meter in self.meters.values():
+            meter.synchronize_between_processes()
+
+    def add_meter(self, name, meter):
+        self.meters[name] = meter
+
+    def log_every(self, iterable, print_freq, header=None):
```
NivekT, line 112: nit: Is it possible to break this into smaller pieces? Or just a few inline comments along the way would be helpful. Thanks!

benchmarks/torchvision_classification/train.py
```diff
@@ -0,0 +1,346 @@
...
+
+    if args.use_deterministic_algorithms:
+        torch.backends.cudnn.benchmark = False
+        torch.use_deterministic_algorithms(True)
+    else:
+        torch.backends.cudnn.benchmark = True
+
+    train_data_loader, val_data_loader, train_sampler = create_data_loaders(args)
+
+    num_classes = 1000  # I'm lazy. TODO change this
```
NivekT, line 206: What does this do? Define the number of classes to classify?

NivekT (approved):
Overall, LGTM! I think we can land this PR as it is and addresses the comments later. I am just leaving them as TODO now for our reference.

We should add a README with instructions to this folder. I am indifferent if whether it is added in this PR or in a later PR since it is not the most urgent.

benchmarks/torchvision_classification/train.py
NicolasHug: Yes, we need to pass this to the model's constructor. It needs to know how many classes there are because this directly impact the number of outputs of the model (one output (probability) per class).

msaroufim (approved):

NivekT (approved):

PR #715, opened 2022-08-04
Fix tutorial about shuffing before sharding

### LINKED ISSUE #709: Update tutorial about shuffling before sharding
### 📚 The doc issue

The [tutorial](https://pytorch.org/data/beta/tutorial.html#working-with-dataloader) needs to update the actual reason about shuffling before sharding. It's not accurate.
Shuffling before sharding is required to achieve global shuffling rather than only shuffling inside each shard.

_No response_

ejguan:
Fixes #709

Per title


NivekT (approved):
LGTM!

Do you also want to mention the exception for internal use case here? Something about data being sharded prior to loading? (I'm probably poorly paraphrasing)

PR #716, opened 2022-08-04
[Release 0.4.1] Install Conda in temp folder, add workflow to test release pipelines, change release type on release branch

ejguan:
Cherry-pick PR: #706 #717

Summary:
Noticed we were getting extra conda packges from the home directory on
the M1 runners. Making this PR to reduce the amount that we're using the
home directory.

Will probably make it unwritable in the future so this is a pre-emptive
step to that.

Also adds a workflow to test the release pipelines without actually releasing a new binary.

Signed-off-by: Eli Uriegas <eliuriegas@fb.com>

Pull Request resolved: https://github.com/pytorch/data/pull/706

Reviewed By: ejguan

Differential Revision: D38406845

Pulled By: seemethere

fbshipit-source-id: 1d312c3e1db34497c546a9b842ac7ba01d2a7ee3

### DIFF
--- a/.github/workflows/_build_test_upload.yml
+++ b/.github/workflows/_build_test_upload.yml
@@ -12,6 +12,10 @@ on:
       pytorch_version:
         required: true
         type: string
+      do-upload:
+        required: false
+        default: true
+        type: boolean
     secrets:
       PYTORCH_BINARY_AWS_ACCESS_KEY_ID:
         required: true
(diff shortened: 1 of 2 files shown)


NivekT (commented):
There are CI errors for both wheel and conda. Any idea why?

Looks like it is deal to `ImportError: cannot import name '_check_lambda_fn` which exists in `0.4.0` but I think has since been changed to `_check_unpickable_fn`. Perhaps there is a mismatch between the version of PT Core and TorchData?

I think we might also need to cherry pick #546 or correct the version of PT Core.

ejguan:
> There are CI errors for both wheel and conda. Any idea why?
> 
> Looks like it is deal to `ImportError: cannot import name '_check_lambda_fn` which exists in `0.4.0` but I think has since been changed to `_check_unpickable_fn`. Perhaps there is a mismatch between the version of PT Core and TorchData?
> 
> I think we might also need to cherry pick #546 or correct the version of PT Core.

Oh. Let me fix that. I think the test release workflow is using nightly PyTorch but it should use 1.12.1

NivekT (commented):
It [fails](https://github.com/pytorch/data/runs/7679998328?check_suite_focus=true) on m1 with Python 3.8. 

Looks like it is due to `subprocess.CalledProcessError: Command '['cmake', '--version']' died with <Signals.SIGABRT: 6>.` Is it just flaky?

ejguan:
> It [fails](https://github.com/pytorch/data/runs/7679998328?check_suite_focus=true) on m1 with Python 3.8.
> 
> Looks like it is due to `subprocess.CalledProcessError: Command '['cmake', '--version']' died with <Signals.SIGABRT: 6>.` Is it just flaky?

I guess it's flaky. The root cause was some missing system dependency. I think it's fine now as the test is passed in this PR [link](https://github.com/pytorch/data/runs/7696682398?check_suite_focus=true).

NivekT (approved):
LGTM! Thanks!

PR #717, opened 2022-08-04
Change release_type on release branch for testing workflow

ejguan:
When `input_branch` is empty, the testing workflow should determine which PyTorch release type should be used for testing PR.

You can find the testing result from the cherry-picking PR #716


NivekT (approved):
LGTM! Thanks!

Issue #712, opened 2022-08-02
Add Examples of Common Preprocessing Steps with IterDataPipe (such as splitting a data set into two)


NivekT:
### 📚 The doc issue

There are a few common steps that users often would like to do while preprocessing data, such as [splitting their data set](https://pytorch.org/docs/stable/data.html#torch.utils.data.random_split) into train and eval. There are documentation in PyTorch Core about how to do these things with `Dataset`. We should add the same to our documentation, specifically for `IterDataPipe`. Or create a link to PyTorch Core's documentation for reference when that is appropriate. This issue is driven by common questions we have received either in person or on the forum.

If we find that any functionality is missing for `IterDataPipe`, we should implement them.

ezeli:
Is there a general method? I currently implement an IterDataPipe that splits the dataset by index:
@functional_datapipe("index_split")
class IndexSpliterIterDataPipe(IterDataPipe):
    def __init__(self, source_dp, start_idx, end_idx) -> None:
        super().__init__()
        self.source_dp = source_dp
        self.start_idx = start_idx
        self.end_idx = end_idx
        assert self.end_idx > self.start_idx

    def __iter__(self):
        source_data = copy.deepcopy(self.source_dp)
        source_data = iter(source_data)
        for _ in range(self.start_idx):
            next(source_data)
        for _ in range(self.start_idx, self.end_idx):
            yield next(source_data)

    def __len__(self):
        return self.end_idx - self.start_idx

NivekT:
@ezeli I think having a custom DataPipe seems fine for your use case. I am open to add that to the library if there are more users who have a use case for it.

We currently have `.header(limit=10)` which can yield from the start up to the specified limit, but not the exact one that you'd like.

If you prefer to use built-in DataPipes instead, you can use one of:
* `dp = dp.enumerate().filter(filter_fn)` - if you want to discard every sample outside of the index
* `dp1, dp2 = dp.enumerate().demux(classifier_fn)` - if you want to split samples into two DataPipes

ezeli:
In fact, it is not advisable to split the dataset by demux. Because the training set, validation set, and test set are often very large, it is easy to exceed the buffer_size, and if the buffer_size is set too large, it will lose the meaning of IterData and will put pressure on the memory.

vincentFrancais:
I second what @ezeli said. Using demux I often exceed the buffer_size on my datasets. Furthermore it's not always straightforward to define a classifier function to split datasets according to one's needs.

[...]

PR #723, opened 2022-08-09
[WIP][DataPipe] Add RandomSplitter (with buffer)

NivekT:
* __->__ #723

This PR adds RandomSplitter with an implementation that uses a buffer through `demux`, thus allowing all child DataPipes to be used simultaneously. This may not work for the memory-bound cases.

TODO:
* Decide if we like a buffer-less version better. Or we can add both.
* Determines if the API related to randomness needs further extension (we might need to add `set_seed`)
* More tests.

See https://github.com/pytorch/data/issues/712 for related discussion.
See #724 for the version WITHOUT buffer.

### DIFF
--- a/docs/source/torchdata.datapipes.iter.rst
+++ b/docs/source/torchdata.datapipes.iter.rst
@@ -1,3 +1,4 @@
+
 Iterable-style DataPipes
 ==========================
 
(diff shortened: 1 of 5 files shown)


NivekT:
@ejguan @VitalyFedyunin This is WIP, we should discuss this but let me know if there is any initial reaction on how we would like to do `random_split`.

NivekT:
Closing this PR for now since we are moving forward with the buffer-less version. We can re-visit this if there is a need for it.

PR #724, opened 2022-08-09
[DataPipe] Add RandomSplitter (without buffer)

NivekT:
* __->__ #724

This PR adds RandomSplitter without a buffer. The upside is that this uses less memory (good for memory-bound cases) but the downside are 1) only one group can be iterated through at a time and 2) it skips over all the groups that do not match the target (which is potentially wasteful).

Implementation note:
* I decided against reusing `_ChildDataPipe` since its features are overly complicated for this use case.
* I also decided against having an option to change seed automatically after each iteration, because there are situations where the first iteration is for `test` and the second iteration is for `valid`. Changing seed will be confusing and causes inconsistency.

See #712 for related discussion.
See #723 for the version with buffer.

Differential Revision: [D38675266](https://our.internmc.facebook.com/intern/diff/D38675266)

### DIFF
--- a/tools/gen_pyi.py
+++ b/tools/gen_pyi.py
@@ -71,6 +71,7 @@ def gen_pyi() -> None:
         "dataframe": "torcharrow.DataFrame",
         "end_caching": "IterDataPipe",
         "unzip": "List[IterDataPipe]",
+        "random_split": "List[IterDataPipe]",
         "read_from_tar": "IterDataPipe",
         "read_from_xz": "IterDataPipe",
         "read_from_zip": "IterDataPipe",
--- a/torchdata/datapipes/iter/util/randomsplitter.py
+++ b/torchdata/datapipes/iter/util/randomsplitter.py
@@ -1,0 +1,77 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
+# All rights reserved.
+#
+# This source code is licensed under the BSD-style license found in the
+# LICENSE file in the root directory of this source tree.
+
+import random
+from typing import Dict, TypeVar, Union
+
+from torchdata.datapipes import functional_datapipe
+from torchdata.datapipes.iter import IterDataPipe
+
+K = TypeVar("K")
+
+
+@functional_datapipe("random_split")
+class RandomSplitterIterDataPipe(IterDataPipe):
+    r"""
+    Randomly split samples from a source DataPipe into groups(functional name: ``random_split``).
+    Since there is no buffer, only one group of samples can be accessed at any time.
+    Note that multiple iterations of this DataPipe will yield the same split for consistency across epochs.
+
+    Args:
+        source_datapipe: Iterable DataPipe being split
+        total_length: Length of the source
+        weights: Dict of weights; the length of this list determines how many output DataPipes there will be.
+        seed: random seed used to determine the split
+        target: key of the group that will be yielded in the current iteration
+
+    Example:
+        >>> from torchdata.datapipes.iter import IterableWrapper
+        >>> dp = IterableWrapper(range(10))
+        >>> train = dp.random_split(total_length=10, weights={"train": 0.5, "valid": 0.5}, seed=0, target="train")
+        >>> list(train)
+        [1, 3, 4, 5, 9]
+        >>> valid = dp.random_split(total_length=10, weights={"train": 0.5, "valid": 0.5}, seed=0, target="valid")
+        >>> list(valid)
+        [2, 8, 6, 7, 0]
+    """
+
+    def __init__(
+        self,
+        source_datapipe: IterDataPipe,
+        total_length: int,
+        weights: Dict[K, Union[int, float]],
+        seed,
+        target: K,
+    ):
+        self.source_datapipe = source_datapipe
+        self.total_length = total_length
+        self.seed = seed
+        self.norm_weights = self.normalize_weights(weights, total_length)
+        self.keys = list(weights.keys())
+        self.key_to_index = {k: i for i, k in enumerate(self.keys)}
+        self.weights = [self.norm_weights[k] for k in self.keys]
+        assert target in self.keys
+        self.target = target
+        self.rng = random.Random(self.seed)
+
+    def __iter__(self):
+        for sample in self.source_datapipe:
+            if self.draw() == self.target:
+                yield sample
+
+    def draw(self):
+        selected_key = self.rng.choices(self.keys, self.weights)[0]
+        self.weights[self.key_to_index[selected_key]] -= 1
+        return selected_key
+
+    @staticmethod
+    def normalize_weights(weights, total_length: int):
+        total_weight = sum(weights.values())
+        return {k: round(float(w) * total_length / total_weight) for k, w in weights.items()}
+
+    def reset(self):
+        self.rng = random.Random(self.seed)
+        self.weights = [self.norm_weights[k] for k in self.keys]
--- a/torchdata/datapipes/iter/util/randomsplitter.py
+++ b/torchdata/datapipes/iter/util/randomsplitter.py
@@ -10,14 +10,15 @@ from typing import Dict, TypeVar, Union
 from torchdata.datapipes import functional_datapipe
 from torchdata.datapipes.iter import IterDataPipe
 
-K = TypeVar("K")
+T = TypeVar("K")
 
 
 @functional_datapipe("random_split")
 class RandomSplitterIterDataPipe(IterDataPipe):
     r"""
     Randomly split samples from a source DataPipe into groups(functional name: ``random_split``).
-    Since there is no buffer, only one group of samples can be accessed at any time.
+    Since there is no buffer, only ONE group of samples (i.e. one child DataPipe) can be iterated through
+    at any time. Attempts to iterate through multiple of them simultaneously will fail.
     Note that multiple iterations of this DataPipe will yield the same split for consistency across epochs.
 
     Args:
@@ -25,26 +26,35 @@ class RandomSplitterIterDataPipe(IterDataPipe):
         total_length: Length of the source
         weights: Dict of weights; the length of this list determines how many output DataPipes there will be.
         seed: random seed used to determine the split
-        target: key of the group that will be yielded in the current iteration
 
     Example:
         >>> from torchdata.datapipes.iter import IterableWrapper
         >>> dp = IterableWrapper(range(10))
-        >>> train = dp.random_split(total_length=10, weights={"train": 0.5, "valid": 0.5}, seed=0, target="train")
+        >>> train, valid = dp.random_split(total_length=10, weights={"train": 0.5, "valid": 0.5}, seed=0)
         >>> list(train)
-        [1, 3, 4, 5, 9]
-        >>> valid = dp.random_split(total_length=10, weights={"train": 0.5, "valid": 0.5}, seed=0, target="valid")
+        [2, 3, 5, 7, 8]
         >>> list(valid)
-        [2, 8, 6, 7, 0]
+        [0, 1, 4, 6, 9]
     """
 
+    def __new__(
+        cls,
+        source_datapipe: IterDataPipe,
+        total_length: int,
+        weights: Dict[T, Union[int, float]],
+        seed,
+    ):
+        container = _RandomSplitterIterDataPipe(source_datapipe, total_length, weights, seed)
+        return [SplitterIterator(container, k) for k in list(weights.keys())]
+
+
+class _RandomSplitterIterDataPipe(IterDataPipe):
     def __init__(
         self,
         source_datapipe: IterDataPipe,
         total_length: int,
-        weights: Dict[K, Union[int, float]],
+        weights: Dict[T, Union[int, float]],
         seed,
-        target: K,
     ):
         self.source_datapipe = source_datapipe
         self.total_length = total_length
@@ -53,14 +63,7 @@ class RandomSplitterIterDataPipe(IterDataPipe):
         self.keys = list(weights.keys())
         self.key_to_index = {k: i for i, k in enumerate(self.keys)}
         self.weights = [self.norm_weights[k] for k in self.keys]
-        assert target in self.keys
-        self.target = target
         self.rng = random.Random(self.seed)
-
-    def __iter__(self):
-        for sample in self.source_datapipe:
-            if self.draw() == self.target:
-                yield sample
 
     def draw(self):
         selected_key = self.rng.choices(self.keys, self.weights)[0]
@@ -75,3 +78,15 @@ class RandomSplitterIterDataPipe(IterDataPipe):
     def reset(self):
         self.rng = random.Random(self.seed)
         self.weights = [self.norm_weights[k] for k in self.keys]
+
+
+class SplitterIterator:
+    def __init__(self, main_datapipe: _RandomSplitterIterDataPipe, target: T):
+        self.main_datapipe = main_datapipe
+        self.target = target
+
+    def __iter__(self):
+        self.main_datapipe.reset()
+        for sample in self.main_datapipe.source_datapipe:
+            if self.main_datapipe.draw() == self.target:
+                yield sample
--- a/docs/source/torchdata.datapipes.iter.rst
+++ b/docs/source/torchdata.datapipes.iter.rst
@@ -1,3 +1,4 @@
+
 Iterable-style DataPipes
 ==========================
 
@@ -178,6 +179,7 @@ A miscellaneous set of DataPipes with different functionalities.
     IterableWrapper
     MapToIterConverter
     OnDiskCacheHolder
+    RandomSplitter
     ShardingFilter
 
 Selecting DataPipes
--- a/test/test_iterdatapipe.py
+++ b/test/test_iterdatapipe.py
@@ -945,44 +945,6 @@ class TestIterDataPipe(expecttest.TestCase):
         # __len__ Test: length matches the length of the shortest input
         self.assertEqual(len(output_dp), 10)
 
-    # def test_random_splitter_iterdatapipe(self):
-    #
-    #     n_epoch = 2
-    #
-    #     # Functional Test: Split results are the same across epochs
-    #     dp = IterableWrapper(range(10))
-    #     train = dp.random_split(total_length=10, weights={"train": 0.5, "valid": 0.5}, seed=0, target="train")
-    #     results = []
-    #     for _ in range(n_epoch):
-    #         res = list(train)
-    #         self.assertEqual(5, len(res))
-    #         results.append(res)
-    #     self.assertEqual(results[0], results[1])
-    #     valid_res = list(dp.random_split(total_length=10, weights={"train": 0.5, "valid": 0.5}, seed=0, target="valid"))
-    #     self.assertEqual(5, len(valid_res))
-    #     self.assertEqual(list(range(10)), sorted(results[0] + valid_res))
-    #
-    #     # Functional Test: DataPipe can split into 3 DataPipes
-    #     dp = IterableWrapper(range(10))
-    #     train = dp.random_split(
-    #         total_length=10, weights={"train": 0.5, "valid": 0.2, "test": 0.2}, seed=0, target="train"
-    #     )
-    #     results = []
-    #     for _ in range(n_epoch):
-    #         res = list(train)
-    #         self.assertEqual(6, len(res))
-    #         results.append(res)
-    #     self.assertEqual(results[0], results[1])
-    #     valid_res = list(
-    #         dp.random_split(total_length=10, weights={"train": 0.5, "valid": 0.2, "test": 0.2}, seed=0, target="valid")
-    #     )
-    #     self.assertEqual(2, len(valid_res))
-    #     test_res = list(
-    #         dp.random_split(total_length=10, weights={"train": 0.5, "valid": 0.2, "test": 0.2}, seed=0, target="test")
-    #     )
-    #     self.assertEqual(2, len(test_res))
-    #     self.assertEqual(list(range(10)), sorted(results[0] + valid_res + test_res))
-
     def test_random_splitter_iterdatapipe(self):
 
         n_epoch = 2
--- a/test/test_iterdatapipe.py
+++ b/test/test_iterdatapipe.py
@@ -945,43 +945,86 @@ class TestIterDataPipe(expecttest.TestCase):
         # __len__ Test: length matches the length of the shortest input
         self.assertEqual(len(output_dp), 10)
 
+    # def test_random_splitter_iterdatapipe(self):
+    #
+    #     n_epoch = 2
+    #
+    #     # Functional Test: Split results are the same across epochs
+    #     dp = IterableWrapper(range(10))
+    #     train = dp.random_split(total_length=10, weights={"train": 0.5, "valid": 0.5}, seed=0, target="train")
+    #     results = []
+    #     for _ in range(n_epoch):
+    #         res = list(train)
+    #         self.assertEqual(5, len(res))
+    #         results.append(res)
+    #     self.assertEqual(results[0], results[1])
+    #     valid_res = list(dp.random_split(total_length=10, weights={"train": 0.5, "valid": 0.5}, seed=0, target="valid"))
+    #     self.assertEqual(5, len(valid_res))
+    #     self.assertEqual(list(range(10)), sorted(results[0] + valid_res))
+    #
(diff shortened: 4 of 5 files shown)


torchdata/datapipes/iter/util/randomsplitter.py
```diff
@@ -0,0 +1,77 @@
...
+        seed: random seed used to determine the split
+        target: key of the group that will be yielded in the current iteration
+
+    Example:
+        >>> from torchdata.datapipes.iter import IterableWrapper
+        >>> dp = IterableWrapper(range(10))
+        >>> train = dp.random_split(total_length=10, weights={"train": 0.5, "valid": 0.5}, seed=0, target="train")
+        >>> list(train)
+        [1, 3, 4, 5, 9]
+        >>> valid = dp.random_split(total_length=10, weights={"train": 0.5, "valid": 0.5}, seed=0, target="valid")
```
ejguan, line 36: I prefer the case without buffer. But, could we still do
`train, val = dp.random_split()` to make sure those sub-sets are mutually exclusive?

NivekT: That makes sense. I updated the PR to do that.

NivekT:
Offline: Discussion:

* This buffer-less version is likely better but we need more clear error message.
* Let's support both syntax - if "target" is provided, then return only one DataPipe. Otherwise, returns a list of DataPipes. Look at the first commit.
* We definitely want `set_seed` to allow changing of `seed`. 
* The default behavior should be same seed every epoch. We can have an argument to allow automatically changing of seed between epochs.

tools/gen_pyi.py
```diff
@@ -71,6 +71,7 @@ def gen_pyi() -> None:
         "dataframe": "torcharrow.DataFrame",
         "end_caching": "IterDataPipe",
         "unzip": "List[IterDataPipe]",
+        "random_split": "Union[IterDataPipe, List[IterDataPipe]]",
```
NivekT, line 74: Is this return type fine? It will either be a single `IterDataPipe` or a `List[...]` depending if `target` is specified.

torchdata/datapipes/iter/util/randomsplitter.py
```diff
@@ -0,0 +1,116 @@
...
+        self.target = target
+
+    def __iter__(self):
+        self.main_datapipe.reset()
+        for sample in self.main_datapipe.source_datapipe:
+            if self.main_datapipe.draw() == self.target:
+                yield sample
+
+    def __len__(self):
+        return self.main_datapipe.norm_weights[self.target]
```
NivekT, line 113: Length can be known in advance based on the given `weights`.

NivekT (commented):
Updated this PR based on our discussion.

Note: I decided against having an option to change seed automatically after each iteration, because there are situations where the first iteration is for test and the second iteration is for valid. Changing seed will be confusing and causes inconsistency. Users should use `set_seed` instead to update the seed when necessary.

NivekT:
@nivekt has imported this pull request.  If you are a Facebook employee, you can view this diff [on Phabricator](https://www.internalfb.com/diff/D38675266).

NivekT:
@nivekt has imported this pull request.  If you are a Facebook employee, you can view this diff [on Phabricator](https://www.internalfb.com/diff/D38675266).

NivekT:
@nivekt has imported this pull request.  If you are a Facebook employee, you can view this diff [on Phabricator](https://www.internalfb.com/diff/D38675266).

VitalyFedyunin:
Can we derive `total_length` from source Datapipe if possible?

NivekT:
@nivekt has imported this pull request.  If you are a Facebook employee, you can view this diff [on Phabricator](https://www.internalfb.com/diff/D38675266).

NivekT:
@nivekt has imported this pull request.  If you are a Facebook employee, you can view this diff [on Phabricator](https://www.internalfb.com/diff/D38675266).

NivekT:
> Can we derive `total_length` from source Datapipe if possible?

Updated the implementation to do that with an exception when it cannot infer length from the source.

torchdata/datapipes/iter/util/randomsplitter.py
```diff
@@ -0,0 +1,135 @@
...
+
+    def __new__(
+        cls,
+        source_datapipe: IterDataPipe,
+        weights: Dict[T, Union[int, float]],
+        seed,
+        total_length: Optional[int] = None,
+        target: Optional[T] = None,
+    ):
+        if total_length is None:
```
VitalyFedyunin, line 61: This might be an issue for DataPipes which have only runtime length. Can we add TODO line to review this later on.

VitalyFedyunin (commented):
Overall LGTM, but I notice there is no state saving/restoration phase, which might be needed considering seed operations.

torchdata/datapipes/iter/util/randomsplitter.py
NivekT: Yea, this is partially why made `total_length` mandatory, so that users have to make a conscious decision to call `len` or provide an integer if it is known.

I will add a note in the docstring to encourage users to provide one as well as add a TODO line.

NivekT (commented):
Updated PR to store RNG state

NivekT:
@nivekt has imported this pull request.  If you are a Facebook employee, you can view this diff [on Phabricator](https://www.internalfb.com/diff/D38675266).

torchdata/datapipes/iter/util/randomsplitter.py
```diff
@@ -0,0 +1,167 @@
...
+
+    def reset(self):
+        self._rng = random.Random(self._seed)
+        self.weights = [self.norm_weights[k] for k in self.keys]
+
+    def set_seed(self, seed):
+        """
+        Update the `seed`. The new `seed` will be used in the next iteration.
+        """
+        self._seed = seed
```
ejguan, line 117: Is that possible to unify seed for all `_RandomSplitterIterDataPipe` from the same source?
We might be able to do so by having the same parent-children pattern as `demux`. The parent class has `set_seed` API, which will be invoked by `DataLoader` per iteration. Then, the children DataPipe would take reference the same seed from parent.

With current implementation, there will be multiple `_RandomSplitterIterDataPipe` in the pipeline and technically each of them should have a different seed generated by `DataLoader`.

ejguan (commented):
I am also [revamping the logic for `Shuffler`](https://github.com/pytorch/pytorch/pull/83535) to make sure it's working proper in all scenarios. You might want to take a look.

torchdata/datapipes/iter/util/randomsplitter.py
```diff
@@ -0,0 +1,167 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
```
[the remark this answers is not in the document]
NivekT: Tbh unsure. I made a new file since the use cases seem different enough and we might add more splitting functionality (such as other ones available in core).

We can potentially consolidate based on the categories in the [doc](https://pytorch.org/data/beta/torchdata.datapipes.iter.html)? Or deal with it when we move things from Core to here?

torchdata/datapipes/iter/util/randomsplitter.py
NivekT: What does this part mean?

> Is that possible to unify seed for all `_RandomSplitterIterDataPipe` from the same source?

Currently, if you `set_seed` for one child, other children are impacted as well.

```python
train, valid = dp.random_split(total_length=10, weights={"train": 0.5, "valid": 0.5}, seed=0)
train.set_seed(100)
print(train.main_datapipe._seed)  # 100
print(valid.main_datapipe._seed)  # 100
```

Or do you mean when users call `.random_split()` multiple times at different points in the pipeline, we would like all those different calls to share the same seed?

torchdata/datapipes/iter/util/randomsplitter.py
```diff
@@ -0,0 +1,167 @@
...
+        >>> valid = dp.random_split(total_length=10, weights={"train": 0.5, "valid": 0.5}, seed=0, target='valid')
+        >>> list(valid)
+        [0, 1, 4, 6, 9]
+    """
+
+    def __new__(
+        cls,
+        source_datapipe: IterDataPipe,
+        weights: Dict[T, Union[int, float]],
+        seed,
```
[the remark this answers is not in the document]
NivekT: I have been thinking about that and I'm on the fence about what makes sense given the use case of `random_split` is different from `shuffle` in that most users probably want to pick one static `seed` and that forget about it for `random_split`. Whereas it is more important to control randomness for `shuffle`.

If we choose to remove `seed` from the argument, and the `seed` is `None`, would you suggest initializing with `self._seed = int(torch.empty((), dtype=torch.int64).random_().item())`?

I am inclined not to do that, and just initialize `random.Random()` with `None` so that the default seed isn't different between runs.

torchdata/datapipes/iter/util/randomsplitter.py
ejguan: > Currently, if you `set_seed` for one child, other children are impacted as well.

Ohhh, Sorry. This is exact what I expect. Just forget my comment above.

torchdata/datapipes/iter/util/randomsplitter.py
```diff
@@ -0,0 +1,167 @@
...
+                yield sample
+
+    def __len__(self):
+        return self.main_datapipe.norm_weights[self.target]
+
+    def set_seed(self, seed):
+        """
+        Update the `seed`. The new `seed` will be used in the next iteration.
+        """
+        self.main_datapipe.set_seed(seed)
```
ejguan, line 167: I was talking about the `set_seed` here. The way I think `apply_seed` should work is:
- Traverse the graph and list all datapipes
- Find the datapipes with `set_seed` api among the list of them
- Set seeds automatically by `DataLoader2` + `ReadingService`

If we keep `set_seed` to both parent and children, it would do seeding `num_of_children + 1` times. It's kind wasting time even thought the seed is still aligned across them

torchdata/datapipes/iter/util/randomsplitter.py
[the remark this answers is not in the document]
ejguan: Related to the [comment](https://github.com/pytorch/data/pull/724/files#r948363330). It depends on how you want to do this split.
If you think it's more like a static split, I am fine to keep seed as an argument. But, I think `set_seed` API should be removed from those DataPipe to prevent `DataLoader2` automatically reset seed every time. You can imagine this would cause splits are not mutually exclusive and collectively exhaustively with `train_dataloader`, `test_dataloader` and `valid_dataloader`.

torchdata/datapipes/iter/util/randomsplitter.py
NivekT: In terms of behavior, IIUC that would change the results of `random_split` for every epoch right? I would guess users prefer not to do that? Do you agree?

So we should either remove or rename `set_seed` to avoid `apply_seed` from using it.

torchdata/datapipes/iter/util/randomsplitter.py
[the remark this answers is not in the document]
NivekT: I am thinking the split should be "mostly static, with optional `set_seed` for special usages".

I think it would be nice to have an option to replace the `seed` but I don't feel strongly enough about it. Maybe remove the method `set_seed` is the best option for now.

cc: @VitalyFedyunin Any thoughts?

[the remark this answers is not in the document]
NivekT: I supposed there might be use cases where users just want to randomly sample x% of data with this DataPipe? In that use case they would want things to be different per epoch.

How about I add an argument `fixed_seed: bool = True`. If `True`, `set_seed` doesn't change the seed. If `False`, it does change.

torchdata/datapipes/iter/util/randomsplitter.py
NivekT: Let's fold this discussion to the other comment thread.

torchdata/datapipes/iter/util/randomsplitter.py
```diff
@@ -0,0 +1,176 @@
...
+    Randomly split samples from a source DataPipe into groups(functional name: ``random_split``).
+    Since there is no buffer, only ONE group of samples (i.e. one child DataPipe) can be iterated through
+    at any time. Attempts to iterate through multiple of them simultaneously will fail.
+
+    Note that by default, multiple iterations of this DataPipe will yield the same split for consistency across epochs.
+    You can invoke ``set_seed`` on the output(s) to update the seed whenever needed (such as per epoch).
+
+    Args:
+        source_datapipe: Iterable DataPipe being split
+        total_length: Length of the ``source_datapipe``, optional but providing an integer is highly encouraged,
```
ejguan, line 28: nit: Make comment follows the order of actual arguments.

torchdata/datapipes/iter/util/randomsplitter.py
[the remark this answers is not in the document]
ejguan: For the case of `fixed_seed = False`, I personally think users would want to have fixed splits within each epoch but different seeds for every epoch.

It becomes tricky for one existing example from pytorch lighting that they need multiple DataLoaders per epoch. But, each `DataLoader` would reset the seed per iteration. Then, we have to ask users to manually reset global RNG after iteration of each dataloader. Code as shown below:
```py
train_dp, val_dp = dp.random_split(total_length=10, weights={"train": 0.7, "valid": 0.3}, seed=0)
train_dataloader = DataLoader(train_dp)
val_dataloader = DataLoader(val_dp)

for epoch in range(3):
    torch.manual_seed(epoch)
    for d in train_dataloader:  # New seed for random_split
        model(d)

    torch.manual_seed(epoch)
    for d in val_dataloader:  # Same seed for random_split
        model(d)
```
`torch.manual_seed(epoch)` has to be placed before the iteration of each DataLoader to make sure the splits static within one epoch, because `apply_seed` will be invoked by `DataLoader2` at the beginning of each iteration.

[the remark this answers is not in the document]
NivekT: I was not expecting users to do `train` `val` split with `fixed_seed = False`, because I don't think they would want to use a different training set for every epoch, right?

I was thinking that if they use `fixed_seed = False`, they likely just want to randomly sample say 50% of the training data, which can be different per epoch and the rest will be dropped.

If my thinking is incorrect, then we should drop the argument `fixed_seed` entirely and go with the seed being static?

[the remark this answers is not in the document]
ejguan: > I was thinking that if they use `fixed_seed = False`, they likely just want to randomly sample say 50% of the training data, which can be different per epoch and the rest will be dropped. Perhaps we should just separate these two functionalities entirely.

I feel this use case is more like a sampling rather than splitting.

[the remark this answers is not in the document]
NivekT: I removed `fixed_seed` as well as renamed `set_seed` then. The intention here is that the `seed` should be static across epochs because I don't believe users want to change the seed by epoch (let me know if this is not true). We will allow `override_seed` but generally we don't expect users to use it.

Any comments?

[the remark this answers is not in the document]
ejguan: Sounds good. In terms of the name, could we do something including `epoch` to give users a hint to only call it per epoch?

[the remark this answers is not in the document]
NivekT: I am hesitate to do that in the name since technically they can use the method separately (not necessarily tied to epochs), but I will update the docstring to indicate that may be a common use case.

NivekT (commented):
I think this should be mostly consistent with `Shuffler` with the exception that the seed doesn't change per epoch.

torchdata/datapipes/iter/util/randomsplitter.py
```diff
@@ -0,0 +1,171 @@
...
+        weights: Dict[T, Union[int, float]],
+        seed,
+    ):
+        self.source_datapipe: IterDataPipe = source_datapipe
+        self.total_length = total_length
+        self._seed = seed
+        self.norm_weights = self.normalize_weights(weights, total_length)
+        self.keys = list(weights.keys())
+        self.key_to_index = {k: i for i, k in enumerate(self.keys)}
+        self.weights = [self.norm_weights[k] for k in self.keys]
```
ejguan, line 97: nit: I feel those variables can be simplified.
Why don't we just get a copy of `self.norm_weights`? `key_to_index` seems redundant to me

torchdata/datapipes/iter/util/randomsplitter.py
```diff
@@ -0,0 +1,171 @@
...
+        self.target = target
+
+    def __iter__(self):
+        self.main_datapipe.reset()
+        for sample in self.main_datapipe.source_datapipe:
+            if self.main_datapipe.draw() == self.target:
+                yield sample
+
+    def __len__(self):
+        return self.main_datapipe.norm_weights[self.target]
```
ejguan, line 163: nit: Would it be better to add a function call to get length rather than hard code the attribute name?

torchdata/datapipes/iter/util/randomsplitter.py
```diff
@@ -0,0 +1,171 @@
...
+
+    def draw(self):
+        selected_key = self._rng.choices(self.keys, self.weights)[0]
+        self.weights[self.key_to_index[selected_key]] -= 1
+        return selected_key
+
+    @staticmethod
+    def normalize_weights(weights, total_length: int):
+        total_weight = sum(weights.values())
+        return {k: round(float(w) * total_length / total_weight) for k, w in weights.items()}
```
ejguan, line 108: And, using `round` might lead to the sum of weights not equal to `total_length`.

ejguan (commented):
Overall, LGTM with a few comments

torchdata/datapipes/iter/util/randomsplitter.py
NivekT: I ended up doing that because [`random.choices`](https://docs.python.org/3/library/random.html#random.choices) (used in `draw`) take in two `list` as arguments.

It is more efficient to convert `self.keys` and `self.weights` to `list` once here rather than for every draw:
`selected_key = self._rng.choices(list(self.weights.keys()), list(self.weights.values()))[0]`

torchdata/datapipes/iter/util/randomsplitter.py
NivekT: Actually I realized the length cannot be known in advance, so I'm taking the method out entirely.

Example: 50/50 split for a DP of length 7. You don't know if `dp1` or `dp2` will have length 3 or 4.

torchdata/datapipes/iter/util/randomsplitter.py
NivekT: You are right. This also means we have to renormalize the weights after one of the keys run out. I have updated my implementation. Please verify.

NivekT:
@nivekt has imported this pull request.  If you are a Facebook employee, you can view this diff [on Phabricator](https://www.internalfb.com/diff/D38675266).

torchdata/datapipes/iter/util/randomsplitter.py
ejguan: I see. Then, `self.norm_weights` can also be a list since we can rely on `keys` and `key_to_index` to indicate the offset. Super nit. Feel free to ignore.

torchdata/datapipes/iter/util/randomsplitter.py
ejguan: Makes sense.

torchdata/datapipes/iter/util/randomsplitter.py
ejguan: I am unsure. If one key runs out, why do we need to renormalize? If one weight goes to 0, the corresponding key should never be returned by `random.choices` IMHO.

torchdata/datapipes/iter/util/randomsplitter.py
```diff
@@ -0,0 +1,185 @@
...
+    def normalize_weights(weights, total_length: int):
+        """
+        Given a ``List`` or a ``Dict`` of weights, normalize them according to ``total_length``.
+        """
+        if isinstance(weights, dict):
+            total_weight = sum(weights.values())
+            return {k: float(w) * total_length / total_weight for k, w in weights.items()}
+        elif isinstance(weights, list):
+            total_weight = sum(weights)
+            return [float(w) * total_length / total_weight for w in weights]
```
ejguan, line 122: I feel we can simplify it by assuming `weights` always being `list`. And, when it's a `dict`, it should be provided to this function via `weights.values()`

torchdata/datapipes/iter/util/randomsplitter.py
NivekT: Let's say we have `total_length=10` and `weights=[3.33, 3.33, 3.33]`.
We draw 4 '0's in a row, the remaining weights will be `weights=[-0.67, 3.33, 3.33]`.
We can set the negative value to 0 and get `weights=[0, 3.33, 3.33]`. Now the remaining number of draws is 6 but the weights sum up to `6.66`. 

It should not impact the result/output, but if we choose to normalize, it will bring the sum back to `6` to match the remaining number of draws. I can take it out since it is technically unnecessary computation.

NivekT: I think I will take it out, because it will greatly simplify the other functions. For example, `normalize_weights` will no longer have to consider the `list` case.

NivekT:
Thanks for the helpful comments. It is simpler than before now!

NivekT:
@nivekt has imported this pull request.  If you are a Facebook employee, you can view this diff [on Phabricator](https://www.internalfb.com/diff/D38675266).

torchdata/datapipes/iter/util/randomsplitter.py
ejguan: Then, how about we always convert `weights` to a list of integer. For example, we do `floor` over all the elements except the last one. And, the rest of elements send to the last key. Then, we don't need to normalize and the weight is kept as integer then no negative value will happen.

ejguan (approved):
Except the last [comment](https://github.com/pytorch/data/pull/724#discussion_r955406409), overall lgtm. Thank you.

torchdata/datapipes/iter/util/randomsplitter.py
NivekT: Two things:
1. I was wrong about not needing to normalize after a weight hitting negative. Actually it is necessary.
    Example: Splitting 10 elements to 9 DPs: [1.11, 1.11, 1.11, 1.11, 1.11, 1.11, 1.11, 1.11, 1.11]
    Without normalization, you may end up with 2 or more DPs with more than one element, while the ideal case is 8 
    DPs with 1 element and 1 DP with 2.
2. I considered your suggestion to `floor` all but the last one. It changes the weights from its true distributino.
    Example: 10 elements split into two DPs with weights [0.89, 0.11], most of the time you want 9 element in the first 
    and 1 element second DP.

Because of these complications, I am going to add a note, telling users that if they want certainty, please provide integer weights that sum up to `total_length`.

torchdata/datapipes/iter/util/randomsplitter.py
```diff
@@ -0,0 +1,196 @@
...
+            self.key_to_index,
+            self.weights,
+            rng_state,
+        ) = state
+        self._rng = random.Random()
+        self._rng.setstate(rng_state)
+
+    def get_length(self, target: T) -> int:
+        if not self._lengths:
+            if all(w.is_integer() for w in self.norm_weights) and sum(self.norm_weights) == self.total_length:
```
NivekT, line 164: We can potentially assert this during `__init__`, such that if the normalized weights aren't integer or cannot sum up to `total_length`, we will raise exception.

torchdata/datapipes/iter/util/randomsplitter.py
ejguan: IIUC, the current implementation is similar to doing `ceiling` to all keys until the last requested. That's why we have to do re-normalization when each time one key is depleted. It's fine but I feel like we can do it in advance without doing normalization multiple times.

NivekT:
@nivekt has imported this pull request.  If you are a Facebook employee, you can view this diff [on Phabricator](https://www.internalfb.com/diff/D38675266).

NivekT:
@nivekt has imported this pull request.  If you are a Facebook employee, you can view this diff [on Phabricator](https://www.internalfb.com/diff/D38675266).

PR #727, opened 2022-08-11
Implement DistribtuedReadingService

ejguan:
Add DistributedReadingService
- Single process
- Share shuffle seeds across distributed process
- Automatically distributed sharding

Add tests for both `DataLoader2` and `DataLoader`.
- Spawn processes
- Elastic training

### DIFF
--- a/test/test_distributed.py
+++ b/test/test_distributed.py
@@ -6,7 +6,6 @@
 
 
 import os
-import subprocess
 import unittest
 
 from functools import partial
@@ -162,16 +161,17 @@ class DistributedTest(TestCase):
     def test_elastic_training_dl2(self, backend) -> None:
         world_size = DEFAULT_WORLD_SIZE if backend != "nccl" else torch.cuda.device_count()
         nnodes = 1
-        subprocess.run(
+        from torch.distributed import run
+
+        run.main(
             [
-                "torchrun",
+                "--run_path",
                 f"--nnodes={nnodes}",
                 f"--nproc_per_node={world_size}",
                 abs_path("bin/elastic_training.py"),
                 "--" + backend,
                 "--dl2",
             ],
-            check=True,
         )
 
     @backend_parametrize
@@ -187,16 +187,17 @@ class DistributedTest(TestCase):
     def test_elastic_training_dl1(self, backend) -> None:
         world_size = DEFAULT_WORLD_SIZE if backend != "nccl" else torch.cuda.device_count()
         nnodes = 1
-        subprocess.run(
+        from torch.distributed import run
+
+        run.main(
             [
-                "torchrun",
+                "--run_path",
                 f"--nnodes={nnodes}",
                 f"--nproc_per_node={world_size}",
                 abs_path("bin/elastic_training.py"),
                 "--" + backend,
                 "--dl1",
             ],
-            check=True,
         )
 
 
--- a/test/test_distributed.py
+++ b/test/test_distributed.py
@@ -6,15 +6,19 @@
 
 
 import os
+import subprocess
 import unittest
 
+from functools import partial
 from unittest import TestCase
 
 import torch
 import torch.distributed as dist
 import torch.multiprocessing as mp
-from torch.testing._internal.common_utils import instantiate_parametrized_tests, parametrize
+from torch.testing._internal.common_utils import instantiate_parametrized_tests, IS_WINDOWS, parametrize
+from torch.utils.data import DataLoader
 
+from torchdata.dataloader2 import DataLoader2, DistributedReadingService
 from torchdata.datapipes.iter import IterableWrapper
 from torchdata.datapipes.iter.util.prefetch import PrefetchTimeoutError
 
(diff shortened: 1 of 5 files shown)


torchdata/dataloader2/reading_service.py
```diff
@@ -235,3 +240,84 @@ def finalize(self) -> None:
...
+        return datapipe
+
+    def initialize_iteration(self) -> None:
+        seed = self._share_seed()
+        self._seed_generator.manual_seed(seed)
+        assert self._datapipe is not None
+        self._datapipe = torch.utils.data.graph_settings.apply_shuffle_seed(
+            self._datapipe,
+            self._seed_generator,
+        )
```
ejguan, line 277: This should be changed after `initialize_iteration` is modified.

VitalyFedyunin:
Just to be clear, does the first version assumes 1 process per rank?

torchdata/dataloader2/reading_service.py
NivekT: I may be out of the loop but what is the planned change here?

NivekT (commented):
Question about the distributed design in general:
1. Are we expecting users to always use something like `torch.multiprocessing.spawn` to start distributed training? And that will properly start/clean up all the processes?
2. To what extent is the optimization from #555 compatible with this? Maybe it can work for every node?

ejguan:
> Just to be clear, does the first version assumes 1 process per rank?

Yeah. I think our next step is to support mixed reading service (distributed reading service + multiprocessing reading service)

ejguan:
> 1. Are we expecting users to always use something like `torch.multiprocessing.spawn` to start distributed training? And that will properly start/clean up all the processes?

Nope. I can add a different test for elastic training.

torchdata/dataloader2/reading_service.py
ejguan: `initialize_iteration` should accept the RNG from `DataLoader2` to set randomness over the pipeline within RS.

test/test_distributed.py
```diff
@@ -15,6 +15,7 @@
 import torch.multiprocessing as mp
 from torch.testing._internal.common_utils import instantiate_parametrized_tests, parametrize
 
+from torchdata.dataloader2 import DataLoader2, DistributedReadingService
 from torchdata.datapipes.iter import IterableWrapper
 from torchdata.datapipes.iter.util.prefetch import PrefetchTimeoutError
```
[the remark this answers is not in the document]
ejguan: Oh, yes. Thanks for catching it. Will added it in this PR!

ejguan:
This PR should be ready to review. cc: @NivekT @VitalyFedyunin 
In terms of multiprocessing + distributed, I will figure out the design of mixed reading services first then add such feature.

ejguan:
Actually, there is one thing left, which is doc and tutorial for DistributedReadingService.

ejguan:
I will wait until https://github.com/pytorch/pytorch/pull/83741 is landed and released into nightly because it will also use the updated API.

ejguan:
Failing DataPipe tests because the nightly binaries for mac haven't been updated yet.

NivekT (approved):
LGTM!

nit comment:
IIUC, the difference between this and `_test_distributed_rs` is DL1 vs DL2?

- If so, we can potential remove the duplicate code (not urgent).
- Alternatively, we can just label `_test_distributed_dl` and `_test_distributed_dataloader` (and `elastic_dl`/`elastic_training`) so it will be obvious from the first glance that those two are mostly the same but testing different version of DL.

test/test_distributed.py
```diff
@@ -24,10 +27,25 @@
...
 if not dist.is_available():
+    import sys
+
     print("Distributed not available, skipping tests", file=sys.stderr)
     sys.exit(0)
 
 
+_backends = ["gloo"]
+#  if dist.is_mpi_available():
+#      _backends.append("mpi")
```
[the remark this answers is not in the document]
ejguan: Fixed

ejguan:
> IIUC, the difference between this and `_test_distributed_rs` is DL1 vs DL2?
> 
> * If so, we can potential remove the duplicate code (not urgent).
> * Alternatively, we can just label `_test_distributed_dl` and `_test_distributed_dataloader` (and `elastic_dl`/`elastic_training`) so it will be obvious from the first glance that those two are mostly the same but testing different version of DL.

Fixed

torchdata/dataloader2/reading_service.py
```diff
@@ -235,3 +240,70 @@ def finalize(self) -> None:
...
+            raise RuntimeError("Torch Distributed is required to be initialized")
+        self._world_size = dist.get_world_size()
+        self._rank = dist.get_rank()
+        self._pg = dist.new_group(backend="gloo", timeout=timedelta(seconds=self._timeout))
+        torch.utils.data.graph_settings.apply_sharding(
+            datapipe,
+            self._world_size,
+            self._rank,
+        )
+        datapipe = datapipe.fullsync(self._timeout)
```
VitalyFedyunin, line 280: super nit: Maybe in future check if `fullsync` already in pipeline (some other cascaded RS added it, or user did it manually).

ejguan: It's a good point. It should be a minimum change with the existing graph functions. Will send a patch shortly.

ejguan:
I will land this PR until PyTorch nightly is updated.

Issue #533, opened 2022-06-22
`len(dataloader)` in distributed setting is different with datapipes and with map-style datasets


NicolasHug:
In a distributed setting, `len(dataloader)` will return:

-  `len(dataset) // (batch_size * num_GPUs)` if `dataset` is a map-style dataset
- `len(dataset) // batch_size` if `dataset` is a datapipe

This discrepancy makes it a bit difficult to work with torchvision's training recipes, where we often need the size of the dataloader.

Below is an illustration of this discrepancy - you can run the snippet (even without a GPU) with `torchrun --nproc_per_node 4 script.py`

```py
# Run this with e.g. `torchrun --nproc_per_node 4 script.py`
import torch.utils.data as data
import torch.distributed as dist
import torchdata


def replace_print():
    import builtins as __builtin__
    builtin_print = __builtin__.print
    def print(*args, **kwargs):
        if dist.get_rank() == 0:
            builtin_print(f"[GPU 0]", *args, **kwargs)

    __builtin__.print = print


# Setting up DDP - you can ignore this
dist.init_process_group(backend="gloo")
replace_print()
dist.barrier()


size = 800
dp = torchdata.datapipes.iter.IterableWrapper(range(size)).sharding_filter()
dl = data.DataLoader(dp, batch_size=10, num_workers=4, drop_last=True)
print(f"with dp, {len(dl) = }")
# Gives : 80

ds = list(range(size))
dl = data.DataLoader(ds, batch_size=10, num_workers=4, drop_last=True, sampler=data.DistributedSampler(ds, shuffle=False))
print(f"with mapstyle, {len(dl) = }")
# Gives: 20

```

ejguan:
Thank you for opening the issue. It's kind easy to fix but need to consider all the use cases.
If users don't specify `sharding_filter` in the pipeline, the length should be `len(dataset) * num_GPUs // batch_size`.

I do want to understand when you need the size of `dataloader`? is this related to the meta data for each Dataset?

NicolasHug:
>  If users don't specify sharding_filter in the pipeline, the length should be len(dataset) * num_GPUs // batch_size.

I agree. Interestingly, with map-style datasets, `len(dataset)` is equal to `len(dataset) // batch_size` if users don't pass `sampler=DistributedSampler()`, which is equivalent to not calling `.sharding_filter()`. But I think `len(dataset) * num_GPUs // batch_size` as you proposed makes more sense.

> I do want to understand when you need the size of dataloader? 

[...]

PR #728, opened 2022-08-11
[DataLoader] Add __len__ to DataLoader2

### LINKED ISSUE #549: DataLoader2.__len__() ?
This is somewhat related to https://github.com/pytorch/data/issues/533

As described in https://github.com/pytorch/data/issues/533#issuecomment-1163381945, we like to check the `len()` of the DataLoader in torchvision in our logging utils.

Are there plans to implement `__len__()` on `DataLoader2`?

NivekT:
* __->__ #728

Adding `__len__` to `DataLoader2`. The main idea is to delegate to reading service to determine what the right length is; this will allow returning length only after the DataPipe is initialized (e.g. sharding is applied) and the correct length to be returned.

For `MultiprocessingReadingService`, we will likely have methods to return one of these or both (undecided):
1. A Dict of `n` elements: `{worker_id: int = worker_dp_length: int}`   
    - `n = num_workers`
    - This will allow users to decide and compute the custom length value that they desire.
2.  A `total_len: int` that is the sum of all the `worker_dp_length`
    - This make less sense if users choose to not shard (and duplicate the original data across workers)

For `DistributedReadingService`, it gets more complicated and we will discuss that separately. 

The implementation currently doesn't reflect those ideas. This will be updated but currently is not a priority.

Fixes #549

Differential Revision: [D38999743](https://our.internmc.facebook.com/intern/diff/D38999743)


torchdata/dataloader2/dataloader2.py
```diff
@@ -130,6 +133,9 @@ def __iter__(self) -> Iterator[T_co]:
         self.valid_iterator_id = 0 if self.valid_iterator_id is None else self.valid_iterator_id + 1
         return DataLoader2Iterator(self, self.valid_iterator_id)
 
+    def __len__(self):
+        return len(self.datapipe)
+
```
ejguan, line 138: Have you tested how it returns when sharding_filter presents? At least in MP environment.

NivekT (commented):
Forgot to post these comments

torchdata/dataloader2/dataloader2.py
NivekT: No I have not but I can do that. See my other comment about potentially delegating to ReadingService for a similar reason.

torchdata/dataloader2/dataloader2.py
```diff
@@ -78,6 +78,9 @@ def __getattr__(self, name):
             raise AttributeError
         return getattr(self.dataloader._datapipe_iter, name)
 
+    def __len__(self):
+        return len(self.dataloader)
```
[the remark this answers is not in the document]
ejguan: I think it's fine to keep it as it is as long as the length from `DL2` is correct.

torchdata/dataloader2/dataloader2.py
```diff
@@ -130,6 +133,9 @@ def __iter__(self) -> Iterator[T_co]:
         self.valid_iterator_id = 0 if self.valid_iterator_id is None else self.valid_iterator_id + 1
         return DataLoader2Iterator(self, self.valid_iterator_id)
 
+    def __len__(self):
+        return len(self.datapipe)
```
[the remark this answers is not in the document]
ejguan: I think we still need to rely on `self.datapipe` because RS won't have aware of length of `self.datapipe` before `__iter__` is invoked (as `reading_service.initialize(datapipe)`). And, after `initialize(datapipe)`, the `self.datapipe` should have been modified and the corresponding changes should be reflected at length.

But, this behavior becomes super weird to me. The length varies before and after `__iter__` is invoked.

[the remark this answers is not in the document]
ejguan: Could we do something like adapting `self.datapipe` when `__len__` is called but `self.datapipe` has not been modified by `self.datapipe`?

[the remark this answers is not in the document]
NivekT: I am going to change it to `len(self._datapipe_before_reading_service_adapt)` because I am guessing that what you mean. It makes sense since that variable also shouldn't be impacted by ReadingService but it would be impacted by adapters within `__init__`, but that is a little bit dependent on the implementation of ReadingService. I think if you `.apply_sharding` within child process and not the main one, the length should not change.

Relevant issue about what is the proper `__len__` to return: https://github.com/pytorch/data/issues/533

torchdata/dataloader2/dataloader2.py
NivekT: `sharding_filter` has not impact on the length, because length doesn't change until `apply_sharding` and I think that is done within processes and not the main one? Correct me if I am wrong on that.

NivekT:
Offline discussion:
* Initialize when `__len__` is called. First, check if RS is initialized, if not, initialize RS, then call `self.ReadingService.get_datapipe_length`(delegating to RS).
* Need to change interface of RS to support these operations.

VitalyFedyunin:
Please re-request review when discussed changes applied.

test/test_dataloader2.py
```diff
@@ -52,6 +56,33 @@ def test_dataloader2(self) -> None:
...
+            # Call `__len__` before initialization
+            self.assertEqual(10, len(data_loader_sharding))
+            _it = iter(data_loader_sharding)
+            # Length should stay the same after calling `__iter__`
+            self.assertEqual(10, len(data_loader_sharding))
+            list(_it)
+            self.assertEqual(10, len(data_loader_sharding))
+
+            # Functional Test: Case with filter
+            data_loader_filter: DataLoader2 = DataLoader2(datapipe=filter_dp, reading_service=rs)
```
NivekT, line 82: Since `rs` is being re-used, this would've failed without the previous PR that deepcopies.

torchdata/dataloader2/reading_service.py
```diff
@@ -145,6 +154,10 @@ def init_datapipe_process(num_workers, worker_id, datapipe):
         torch.utils.data.graph_settings.apply_sharding(datapipe, num_workers, worker_id)
 
     def initialize(self, datapipe: DataPipe) -> DataPipe:
+        try:
+            self._length = len(datapipe)
+        except TypeError:
+            pass
```
NivekT, line 160: Two things:
1. This is  taking the `len` before any `apply_sharding` happens. I think that is the right number that we want to return for multiprocessing. Agree?
2. This is only ran once (instead of per iteration) as I don't think users want different length per iteration. Is this a reasonable assumption?

NivekT:
@nivekt has imported this pull request.  If you are a Facebook employee, you can view this diff [on Phabricator](https://www.internalfb.com/diff/D38999743).

NivekT:
@nivekt has imported this pull request.  If you are a Facebook employee, you can view this diff [on Phabricator](https://www.internalfb.com/diff/D38999743).

NivekT:
@nivekt has imported this pull request.  If you are a Facebook employee, you can view this diff [on Phabricator](https://www.internalfb.com/diff/D38999743).

NivekT:
@nivekt has imported this pull request.  If you are a Meta employee, you can view this diff [on Phabricator](https://www.internalfb.com/diff/D38999743).

NivekT:
@nivekt has imported this pull request.  If you are a Meta employee, you can view this diff [on Phabricator](https://www.internalfb.com/diff/D38999743).

torchdata/dataloader2/reading_service.py
```diff
@@ -365,6 +390,10 @@ def initialize(self, datapipe: DataPipe) -> DataPipe:
         if not isinstance(datapipe, FullSync):
             datapipe = datapipe.fullsync(self._timeout)
         self._datapipe = datapipe
+        try:
+            self._length = len(datapipe)
+        except TypeError:
+            self._length = None
```
NivekT, line 396: Is this the right time/place to request length for distributed?

SeanLiaoy:
Is there any update? I think it's very helpful for using Trainers like pytorch-lightning or huggingface Transformers.

NivekT:

Issue #648, opened 2022-07-13
Chainer/Concater from single datapipe?


NicolasHug:
The `Concater` datapipe takes multiple DPs as input. Is there a class that would take a **single** datapipe of iterables instead? Something like this:

```py
class ConcaterIterable(IterDataPipe):
    def __init__(self, source_datapipe):
        self.source_datapipe = source_datapipe

    def __iter__(self):
        for iterable in self.source_datapipe:
            yield from iterable
```

Basically:

[`itertools.chain` ](https://docs.python.org/3/library/itertools.html#itertools.chain)== `Concater`
[`itertools.chain.from_iterable`](https://docs.python.org/3/library/itertools.html#itertools.chain.from_iterable) == `ConcaterIterable`

Maybe a neat way of implementing this would be to keep a single `Concater` class, which would fall back to the  `ConcaterIterable` behaviour if it's passed only one DP as input?

-----

Details: I need this for my benchmarking on manifold where each file is a big pickle archive of multiple images. My DP builder looks like this:

```py
def make_manifold_dp(root, dataset_size):
    handler = ManifoldPathHandler()
    dp = IoPathFileLister(root=root)
    dp.register_handler(handler)

    dp = dp.shuffle(buffer_size=dataset_size).sharding_filter()

    dp = IoPathFileOpener(dp, mode="rb")
    dp.register_handler(handler)

    dp = PickleLoaderDataPipe(dp)
    dp = ConcaterIterable(dp)  # <-- Needed here!
    return dp
```

NicolasHug:
BTW, this is a NIT, but has it been considered to rename `Concater` into `Chainer` to be a bit more consistent with `itertools`?

VitalyFedyunin:
You can try to use `.unbatch()` for it, it is not so generic but might work in your case.

However proper solution would be to add new DataPipe. And I would rather call it `flatten`

VitalyFedyunin:
Note: we already used `flatten` for horizontal/column operations. Perhaps we need rows_flatten or other (way) better name.

NicolasHug:
> You can try to use `.unbatch()` for it, it is not so generic but might work in your case.
> 
> However proper solution would be to add new DataPipe. And I would rather call it `flatten`


Looks like  `unbatch()` doesn't work for datapipes that contain datapipes, so in https://github.com/pytorch/data/issues/732 I still had to resort to something like `ConcaterIterable` above

VitalyFedyunin:
[...]

PR #730, opened 2022-08-13
added slice + flatten for iterable datapipes

### LINKED ISSUE #656: Ability to manipulate columns and fields
### 🚀 The feature

- [x] Add ability to drop specific column / field

For `list` and `tuple`
```python
list(dp) # [ (0, 1, 2), (3, 4, 5), (6,7, 8) ]
dp = dp.drop(1)
list(dp) # [ (0, 2), (3, 5), (6, 8) ]
```

Similar for `dict` types
```python
list(dp) # [ {a:1, b:2}, {a:3, b:4}, {a:5, b:6}]
dp = dp.drop('a')
list(dp) # [ {b:2}, {b:4}, {b:6}]
```

- [ ] Add ability to slice fields

For `list` and `tuple`
```python
list(dp) # [ (0, 1, 2, 0, 1, 2), (3, 4, 5, 3,4,5), (6,7, 8,6,7,8) ]
dp = dp[1:-1]
list(dp) # [ ( 1, 2, 0, 1), ( 4, 5, 3,4), (7, 8,6,7) ]
```

For `dict`
```python
list(dp) # [ {a:1, b:2, c:3, d:4}, {a:3, b:4, c:3, d:4}, {a:5, b:6, c:3, d:4}]
dp = dp['a','b']
list(dp) #  [ {a:1, b:2}, {a:3, b:4}, {a:5, b:6}]
```

- [ ] Add ability to flatten structures

```python
list(dp) # [ (1, (2,3), 4), (5, (6,7), 8) ]
dp = dp.flatten(1)
list(dp) # [ (1, 2, 3,4), (5,6,7,8) ]
```

Note: No arguments `flatten()` should cover use-case from #648
Note: Exception if structures are different length 

```python
list(dp) # [ { a: 1, b: {e: 2, f:3 }, c: 4} , { a: 1, b: {e: 2, f:3 }, c: 4} ]
dp = dp.flatten('b')
list(dp) # [ { a: 1, e: 2, f:3 , c: 4}  , { a: 1, e: 2, f:3, c: 4}]
```
Note: Exception if keys overlaps

CC @NivekT @ejguan

dbish:
creating your pull request.

Fixes #656 

### Changes

- added new iterdatapipe called islice for selecting a slice of an iterable item returned (ex: list, tuple, dict)
- added indexing as part of this islice functionality, allowing a list of indexes (or keys) to also be used in place of a slice for selection
- also added flatten iterdatapipe to allow for an item in an iterable to be flattened one level if an index or list of indices are provided
- added default for flatten (no args) that yields all items for all iterables in the provided datapipe chained together

### DIFF
--- a/test/test_iterdatapipe.py
+++ b/test/test_iterdatapipe.py
@@ -993,6 +993,74 @@ class TestIterDataPipe(expecttest.TestCase):
         drop_dp = input_dp.drop([0, 1])
         self.assertEqual(3, len(drop_dp))
 
+    def test_islice_iterdatapipe(self):
+        # tuple tests
+        input_dp = IterableWrapper([(0, 1, 2), (3, 4, 5), (6, 7, 8)])
+
+        # Functional Test: slice with no stop and no step for tuple
+        islice_dp = input_dp.islice(1)
+        self.assertEqual([(1, 2), (4, 5), (7, 8)], list(islice_dp))
+
+        # Functional Test: slice with no step for tuple
+        islice_dp = input_dp.islice(0, 2)
+        self.assertEqual([(0, 1), (3, 4), (6, 7)], list(islice_dp))
+
+        # Functional Test: slice with step for tuple
+        islice_dp = input_dp.islice(0, 2, 2)
+        self.assertEqual([(0,), (3,), (6,)], list(islice_dp))
+
+        # Functional Test: filter with list of indices for tuple
+        islice_dp = input_dp.islice([0, 1])
+        self.assertEqual([(0, 1), (3, 4), (6, 7)], list(islice_dp))
+
+        # list tests
+        input_dp = IterableWrapper([[0, 1, 2], [3, 4, 5], [6, 7, 8]])
+
+        # Functional Test: slice with no stop and no step for list
+        islice_dp = input_dp.islice(1)
+        self.assertEqual([[1, 2], [4, 5], [7, 8]], list(islice_dp))
+
+        # Functional Test: slice with no step for list
+        islice_dp = input_dp.islice(0, 2)
+        self.assertEqual([[0, 1], [3, 4], [6, 7]], list(islice_dp))
+
+        # Functional Test: filter with list of indices for list
+        islice_dp = input_dp.islice(0, 2)
+        self.assertEqual([[0, 1], [3, 4], [6, 7]], list(islice_dp))
+
+        # dict tests
+        input_dp = IterableWrapper([{"a": 1, "b": 2, "c": 3}, {"a": 3, "b": 4, "c": 5}, {"a": 5, "b": 6, "c": 7}])
+
+        # Functional Test: slice with no stop and no step for dict
+        islice_dp = input_dp.islice(1)
+        self.assertEqual([{"b": 2, "c": 3}, {"b": 4, "c": 5}, {"b": 6, "c": 7}], list(islice_dp))
+
+        # Functional Test: slice with no step for dict
+        islice_dp = input_dp.islice(0, 2)
+        self.assertEqual([{"a": 1, "b": 2}, {"a": 3, "b": 4}, {"a": 5, "b": 6}], list(islice_dp))
+
+        # Functional Test: slice with step for dict
+        islice_dp = input_dp.islice(0, 2, 2)
+        self.assertEqual([{"a": 1}, {"a": 3}, {"a": 5}], list(islice_dp))
+
+        # Functional Test: filter with list of indices for dict
+        islice_dp = input_dp.islice(["a", "b"])
+        self.assertEqual([{"a": 1, "b": 2}, {"a": 3, "b": 4}, {"a": 5, "b": 6}], list(islice_dp))
+
+        # __len__ Test:
+        input_dp = IterableWrapper([[0, 1, 2], [3, 4, 5], [6, 7, 8]])
+        islice_dp = input_dp.islice(0, 2)
+        self.assertEqual(3, len(islice_dp))
+
+        # Reset Test:
+        n_elements_before_reset = 2
+        input_dp = IterableWrapper([[0, 1, 2], [3, 4, 5], [6, 7, 8]])
+        islice_dp = input_dp.islice([2])
+        expected_res = [[2], [5], [8]]
+        res_before_reset, res_after_reset = reset_after_n_next_calls(islice_dp, n_elements_before_reset)
+        self.assertEqual(expected_res[:n_elements_before_reset], res_before_reset)
+        self.assertEqual(expected_res, res_after_reset)
+
 
 if __name__ == "__main__":
     unittest.main()
--- a/test/test_iterdatapipe.py
+++ b/test/test_iterdatapipe.py
@@ -993,71 +993,59 @@ class TestIterDataPipe(expecttest.TestCase):
         drop_dp = input_dp.drop([0, 1])
         self.assertEqual(3, len(drop_dp))
 
-    def test_islice_iterdatapipe(self):
+    def test_slice_iterdatapipe(self):
         # tuple tests
         input_dp = IterableWrapper([(0, 1, 2), (3, 4, 5), (6, 7, 8)])
 
         # Functional Test: slice with no stop and no step for tuple
-        islice_dp = input_dp.islice(1)
-        self.assertEqual([(1, 2), (4, 5), (7, 8)], list(islice_dp))
+        slice_dp = input_dp.slice(1)
+        self.assertEqual([(1, 2), (4, 5), (7, 8)], list(slice_dp))
 
         # Functional Test: slice with no step for tuple
-        islice_dp = input_dp.islice(0, 2)
-        self.assertEqual([(0, 1), (3, 4), (6, 7)], list(islice_dp))
+        slice_dp = input_dp.slice(0, 2)
+        self.assertEqual([(0, 1), (3, 4), (6, 7)], list(slice_dp))
 
         # Functional Test: slice with step for tuple
-        islice_dp = input_dp.islice(0, 2, 2)
-        self.assertEqual([(0,), (3,), (6,)], list(islice_dp))
+        slice_dp = input_dp.slice(0, 2, 2)
+        self.assertEqual([(0,), (3,), (6,)], list(slice_dp))
 
         # Functional Test: filter with list of indices for tuple
-        islice_dp = input_dp.islice([0, 1])
-        self.assertEqual([(0, 1), (3, 4), (6, 7)], list(islice_dp))
+        slice_dp = input_dp.slice([0, 1])
+        self.assertEqual([(0, 1), (3, 4), (6, 7)], list(slice_dp))
 
         # list tests
         input_dp = IterableWrapper([[0, 1, 2], [3, 4, 5], [6, 7, 8]])
 
         # Functional Test: slice with no stop and no step for list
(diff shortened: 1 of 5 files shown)


torchdata/datapipes/iter/transform/callable.py
```diff
@@ -205,3 +206,162 @@ def __len__(self) -> int:
         if isinstance(self.datapipe, Sized):
             return len(self.datapipe)
         raise TypeError(f"{type(self).__name__} instance doesn't have valid length")
+
+
+@functional_datapipe("islice")
```
ejguan, line 211: `islice` might not be the best name. The behavior is different from [`itertolls.islice`](https://docs.python.org/3/library/itertools.html#itertools.islice). We might make users confused by the behavior.

torchdata/datapipes/iter/transform/callable.py
```diff
@@ -205,3 +206,162 @@ def __len__(self) -> int:
...
+                if isinstance(self.index, list):
+                    new_item = [x for i, x in enumerate(old_item) if i in self.index]  # type: ignore[assignment]
+                else:
+                    new_item = old_item[self.index : self.stop : self.step]  # type: ignore[assignment]
+            elif isinstance(old_item, dict):
+                if isinstance(self.index, list):
+                    new_item = {k: v for (k, v) in old_item.items() if k in self.index}  # type: ignore[assignment]
+                else:
+                    new_keys = list(old_item.keys())[self.index : self.stop : self.step]
+                    new_item = {k: v for (k, v) in old_item.items() if k in new_keys}  # type: ignore[assignment]
```
ejguan, line 262: For the case of `dict`, could we align the behavior no matter `self.index` is list or slice? TBH, supporting slice for `dict` is a little bit weird to me.

torchdata/datapipes/iter/transform/callable.py
```diff
@@ -205,3 +206,162 @@ def __len__(self) -> int:
...
+        index: Union[int, List[Hashable]],
+        stop: Optional[int] = None,
+        step: Optional[int] = None,
+    ) -> None:
+        super().__init__()
+        self.datapipe = datapipe
+
+        self.index = index
+        self.stop = stop
+        self.step = step
```
ejguan, line 243: Could we add a check here to validate list of indices should be mutually exclusive from `stop` and `step`?

torchdata/datapipes/iter/transform/callable.py
dbish: slice didn't seem quite right either, since it only slices the iterables within the datapipe, instead of the entire datapipe, but you're right the islice itertools version is a bit different too. Any other good names we might pick instead?

torchdata/datapipes/iter/transform/callable.py
dbish: that's fair. dict is ordered nowadays by default, so it seemed ok to have a proper slice out of it with that expectation. I'm also fine with just having it be expected that dicts can only have the direct indexing filter. Whichever you all think users would want and understand.

VitalyFedyunin: Dicts insertion ordering is only guaranteed in python >= 3.7. Generally, range slicing of dicts is the extremely confusing feature and I would rather drop it. Key slicing `islice(["a", "b"])` is still useful and we should keep it.

dbish: Noted. Will drop it then and throw an error instead for dicts if a range is provided

torchdata/datapipes/iter/__init__.py
```diff
@@ -69,6 +69,7 @@
     BatchMapperIterDataPipe as BatchMapper,
     DropperIterDataPipe as Dropper,
     FlatMapperIterDataPipe as FlatMapper,
+    FlattenIterDataPipe as Flatten,
```
dbish, line 72: probably should be "Flattener" instead of "Flatten"

torchdata/datapipes/iter/transform/callable.py
```diff
@@ -283,3 +284,84 @@ def __len__(self) -> int:
...
+                        old_item[index]
+                except (IndexError, KeyError):
+                    warnings.warn(
+                        "At least one index in the filter is not present in the item being returned,"
+                        " please be aware that expected columns/keys may be missing."
+                    )
+
+                yield new_item  # type: ignore[misc]
+
+            else:
```
dbish, line 358: @VitalyFedyunin does this look like the default behavior you wanted? One level of automatic flattening but more like an iterator created across that level, so you get one item at a time from that flattened list? It's what was described by @NicolasHug here: https://github.com/pytorch/data/issues/648 (and probably why he wanted to name it chainer).

VitalyFedyunin: The proposed solution looks like `unbatch` to me. 

For default behaviour I would rather do something like:

```python
list(dp) # [ (1, (2,3), 4), (5, (6,7), 8) ]
dp = dp.flatten()
list(dp) # [ (1, 2, 3,4), (5,6,7,8) ]
```

Meaning expand all nested fields by checking something like:

```
from collections.abc import Iterable

if not isinstance(x, Iterable):
  x = [x]
```

VitalyFedyunin: We probably want to exclude `str` class. Do avoid expanding strings to characters.

dbish: I can make that change. It won't solve the problem described in Nicolas's request though at least how I've read what he was looking for.

VitalyFedyunin: It will not. He needs `unbatch` or no-op `flatmap`.

torchdata/datapipes/iter/transform/callable.py
dbish: @VitalyFedyunin just slice or should we call it something else?

VitalyFedyunin: `slice` is good.

torchdata/datapipes/iter/transform/callable.py
```diff
@@ -205,3 +206,162 @@ def __len__(self) -> int:
...
+        for old_item in self.datapipe:
+            if self.indices:
+                if isinstance(old_item, dict):
+                    new_item = {}  # type: ignore[assignment]
+                    for k, v in old_item.items():
+                        if k in self.indices:
+                            for k_sub, v_sub in v.items():
+                                new_item[k_sub] = v_sub
+                        else:
+                            new_item[k] = v
```
ejguan, line 331: Wdyt about adding a warning when `k_sub` has already been presented in `old_item`?

dbish: yep, good idea. i can do that

torchdata/datapipes/iter/transform/callable.py
NivekT: @VitalyFedyunin What makes you prefer that default behavior instead of what is implemented?

I am used to this being the default:
```python
list(dp)  # [(1, (2, 3), 4), (5, (6, 7), 8)]
dp = dp.flatten()
list(dp)  # [1, (2, 3), 4, 5, (6, 7), 8]
```
Here is the [`numpy/scipy` implementation](https://numpy.org/doc/stable/reference/generated/numpy.ndarray.flatten.html) for reference.

I agree that we can have additional arguments `indices` to unpack specific indices rather than the whole element. Just curious about the choice of default here.

dbish: Updated default per Vitaly's feedback. Lmk if you two decide we should change this. The newest commit shows the new default behavior so you can compare to the previous option.

torchdata/datapipes/iter/transform/callable.py
```diff
@@ -205,3 +205,185 @@ def __len__(self) -> int:
...
+            flatten_all = True
+        for old_item in self.datapipe:
+            if isinstance(old_item, dict):
+                new_item = {}  # type: ignore[assignment]
+                for k, v in old_item.items():
+                    if k in self.indices:
+                        pass
+                    if (flatten_all or (k in self.indices)) and isinstance(v, dict):
+                        for k_sub, v_sub in v.items():
+                            if k_sub not in new_item:
```
ejguan, line 345: I think it's better to check `if k not in old_item` because there might be a chance that the the key is ordered behind `k_sub`. For example of `{"a": 1, "sub": {"a": 11, "b": 12}, "b": 2}, we only raise warning for `a` but not `b`.

torchdata/datapipes/iter/transform/callable.py
```diff
@@ -205,3 +205,185 @@ def __len__(self) -> int:
...
+                                new_item[k_sub] = v_sub
+                            else:
+                                warnings.warn(
+                                    "Flattener tried to insert the same key twice into the same dict,"
+                                    "the second key,value pair has been dropped."
+                                )
+                    else:
+                        if k not in new_item:
```
ejguan, line 353: ditto

NivekT (commented):
I chatted with Vitaly and understand what the intention is with the default behavior when nothing is passed in.

We also chatted about changing the name from `flatten` to `flatten_sample` but that is probably too much and potentially more confusing.

I would recommend:
1) Being explicit in the docstring about the `flatten` operation is being applied at the element/sample level.
2) Adding one example in docstring with the default behavior when nothing is passed in. This can be copied over from one of the unit tests.

Otherwise, LGTM!

test/test_iterdatapipe.py
```diff
@@ -993,6 +993,159 @@ def test_drop_iterdatapipe(self):
...
+
+        # Functional Test: slice with no stop and no step for list
+        slice_dp = input_dp.slice(1)
+        self.assertEqual([[1, 2], [4, 5], [7, 8]], list(slice_dp))
+
+        # Functional Test: slice with no step for list
+        slice_dp = input_dp.slice(0, 2)
+        self.assertEqual([[0, 1], [3, 4], [6, 7]], list(slice_dp))
+
+        # Functional Test: filter with list of indices for list
```
VitalyFedyunin, line 1027: Same as previous lines?

test/test_iterdatapipe.py
```diff
@@ -993,6 +993,159 @@ def test_drop_iterdatapipe(self):
...
+        # Functional Test: flatten all iters in the datapipe one level (no argument)
+        input_dp = IterableWrapper([[0, [1, 2]], [3, [4, 5]], [6, [7, 8]]])
+        flatten_dp = input_dp.flatten()
+        self.assertEqual([[0, 1, 2], [3, 4, 5], [6, 7, 8]], list(flatten_dp))
+
+        # Functional Test: string test, flatten all iters in the datapipe one level (no argument)
+        input_dp = IterableWrapper([["zero", ["one", "2"]], ["3", ["4", "5"]], ["6", ["7", "8"]]])
+        flatten_dp = input_dp.flatten()
+        self.assertEqual([["zero", "one", "2"], ["3", "4", "5"], ["6", "7", "8"]], list(flatten_dp))
+
```
VitalyFedyunin, line 1091: Nit: Probably include test and mention in the doc, that we are not supporting nesting levels > 1

VitalyFedyunin (approved):
Overall good. 

As part of the exercise you can consider including `assertWarns` tests for borderline situations:

https://docs.python.org/3/library/unittest.html#unittest.TestCase.assertWarns

PR #740, opened 2022-08-16
Add nightly benchmark and trigger in keyword

msaroufim:
This PR runs Nicolas' benchmark suite nightly and stores the results in a github artifact so anyone on the team can inspect them. It's also possible to trigger the benchmark suite manually by typing `RUN_BENCHMARK` in the PR description which is how this benchmark run was actually triggered

Eventually we'd need to add more benchmarks, log them to the same directory, maybe do a dashboard and of course integrate it into an AWS infra but at least this way the team has a way of easily running benchmarks now without needing to manually spin up machine

h/t @xuzhao9

### DIFF
--- a/.github/workflows/benchmark.yml
+++ b/.github/workflows/benchmark.yml
@@ -28,15 +28,10 @@ jobs:
         echo "::set-output name=value::$PT_CHANNEL"
       id: pytorch_channel
     - name: Setup additional system libraries
-      if: startsWith( matrix.os, 'ubuntu' )
       run: |
         sudo add-apt-repository multiverse
         sudo apt update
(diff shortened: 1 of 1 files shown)


msaroufim:
@NicolasHug is there a setup script I need to run beforehand? https://github.com/pytorch/data/runs/7868534954?check_suite_focus=true#step:9:23

NivekT:
> @NicolasHug is there a setup script I need to run beforehand? https://github.com/pytorch/data/runs/7868534954?check_suite_focus=true#step:9:23

That only works on the internal AI cluster. I will message you more details.

NivekT (commented):
I think we need an extra step to get the relevant data. Let's talk offline in our meeting.

PR #745, opened 2022-08-18
Unify implementation of in_batch_shuffle with `IterDataPipe.shuffle`

### LINKED ISSUE #519: `in_batch_shuffle` currently doesn't use shared RNG
Currently, `in_batch_shuffle` doesn't use the RNG shared across processes. This can be problematic if it is used before prior to sharding, resulting in certain samples not being used in training while others may be used by multiple processes.

https://github.com/pytorch/data/blob/100d086413873a5c224842da4c2cd55cb634317f/torchdata/datapipes/iter/transform/bucketbatcher.py#L19-L20

Note that `in_batch_shuffle` is also used by `bucketbatch`.

https://github.com/pytorch/data/blob/100d086413873a5c224842da4c2cd55cb634317f/torchdata/datapipes/iter/transform/bucketbatcher.py#L50-L51

Moreover, we may want to provide the ability to `set_seed` for those two DataPipes to enable reproducibility.

ejguan:
Fixes #519

This PR takes reference from the modification from [my PR](https://github.com/pytorch/pytorch/pull/83535) for `IterDataPipe.shuffle`.
- Add `set_seed`
- Add `set_shuffle`
- Make `RNG` lazily seeded
- Make serialization works

There will be another PR in PyTorch core to make `shuffle` related functions ([`apply_shuffle_settings`](https://github.com/pytorch/pytorch/blob/38348362608a47371c65d7fd52db138b4c6a5d65/torch/utils/data/graph_settings.py#L48), [`apply_shuffle_seed`](https://github.com/pytorch/pytorch/blob/38348362608a47371c65d7fd52db138b4c6a5d65/torch/utils/data/graph_settings.py#L69)) working with all `DataPipe` that has `set_seed` and `set_shuffle` APIs rather than only `iter.Shuffler` or `map.Shuffler`.

### DIFF
--- a/torchdata/datapipes/iter/transform/bucketbatcher.py
+++ b/torchdata/datapipes/iter/transform/bucketbatcher.py
@@ -61,10 +61,11 @@ class InBatchShufflerIterDataPipe(IterDataPipe[DataChunk[T_co]]):
                 yield DataChunk(new_batch)
 
     def reset(self) -> None:
-        if self._enabled and self._seed is None:
-            self._seed = int(torch.empty((), dtype=torch.int64).random_().item())
-        self._rng.seed(self._seed)
-        self._seed = None
+        if self._enabled:
+            if self._seed is None:
+                self._seed = int(torch.empty((), dtype=torch.int64).random_().item())
+            self._rng.seed(self._seed)
+            self._seed = None
 
     def __len__(self) -> int:
         return len(self.datapipe)
--- a/torchdata/datapipes/iter/transform/bucketbatcher.py
+++ b/torchdata/datapipes/iter/transform/bucketbatcher.py
@@ -10,6 +10,8 @@ import random
 from functools import partial
 from typing import Callable, Iterator, List, Optional, TypeVar
 
+import torch
+
 from torchdata.datapipes import DataChunk, functional_datapipe
 from torchdata.datapipes.iter import IterDataPipe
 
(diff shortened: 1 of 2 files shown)


ejguan:
https://github.com/pytorch/pytorch/pull/83741 will update the seed of `in_batch_shuffle` per epoch in DataLoader

NivekT (commented):
Let's follow up on this after https://github.com/pytorch/pytorch/pull/83535 is finalized.

ejguan:
This PR is ready to review. cc: @NivekT @VitalyFedyunin

torchdata/datapipes/iter/transform/bucketbatcher.py
```diff
@@ -37,15 +39,61 @@ class InBatchShufflerIterDataPipe(IterDataPipe[DataChunk[T_co]]):
...
+        else:
+            for batch in self.datapipe:
+                new_batch = self._rng.sample(batch, len(batch))
+                yield DataChunk(new_batch)
+
+    def reset(self) -> None:
+        if self._enabled and self._seed is None:
+            self._seed = int(torch.empty((), dtype=torch.int64).random_().item())
+        self._rng.seed(self._seed)
+        self._seed = None
```
NivekT, line 67: No strong opinion about this feel free to ignore:
Should we skip over these two lines as well if `self._enabled == False`?

The practical difference is that `self._rng` and `self._seed` will not change unless `self._enabled == True`.
Tbh I doubt we will run into a case like this but anyway

NivekT (approved):
LGTM. We can land this after if anything needs to change for https://github.com/pytorch/pytorch/pull/83741.

torchdata/datapipes/iter/transform/bucketbatcher.py
NivekT: Actually `shuffler` is already doing that so we can keep it that way.

ejguan: This is actually a good catch considering a super corner case that `self._enabled = False` and `self._seed is not None` and `seed` will be eliminated by this `reset` function.

Thank you. But, I might open a separate PR to change both of them.

ejguan: > But, I might open a separate PR to change both of them.

Forget about it. I will change it right away

NivekT (approved):

PR #746, opened 2022-08-18
[DataLoader] Deep copy ReadingService during DL2 initialization

NivekT:
* #728
* __->__ #746

By creating a deep copy of the input `ReadingService` during the initialization of `DataLoader2`, we allow the instance of `ReadingService` to store state without worrying about the users passing the same `ReadingService` to initialize additional DataLoaders.

For instance, `PrototypeMultiProcessingReadingService` stores `processes` and `datapipes` as instance variables. If a single object of that ReadingService is used to initialize multiple DataLoaders, it can led to conflicts and incorrect states. This should no longer be an issue with this change.

Relevant test are added in subsequent PR related to lengths and eventually `PrototypeMultiprocessingReadingService`.

Differential Revision: [D38868731](https://our.internmc.facebook.com/intern/diff/D38868731)

----

**BC Breaking Note**:

Previously, if a ReadingService is being passed to multiple DataLoaders, the ReadingService's state can be altered by any of those DataLoaders. This use case is not intended and can lead to silent incorrectness.

For instance, `PrototypeMultiProcessingReadingService` stores `processes` and `datapipes` as instance variables. If a single object of that ReadingService is used to initialize multiple DataLoaders, it can led to conflicts and incorrect states. 

Consider the code snippet below:
```python
dp = IterableWrapper([0, 1, 2, 3, 4])  #.shuffle()
rs = PrototypeMultiProcessingReadingService(num_workers=2)
dl1 = DataLoader2(dp, reading_service=rs)
dl2 = DataLoader2(dp, reading_service=rs)
next(iter(dl1))
print(f"Number of processes that exist in `dl1`'s RS after initializing `dl1`: {len(dl1.reading_service.processes)}")
print(next(iter(dl2)))
# Note that we are still examining `dl1.read_service` below
print(f"Number of processes that exist in `dl1`'s RS after initializing `dl2`: {len(dl1.reading_service.processes)}")
```
Prior to this change, since the reading service objects are identical within each DataLoader2, initializing `dl2` will change the state of the reading service that exists in `dl1`.
```
Number of processes that exist in `dl1`'s RS after initializing `dl1`: 2
Number of processes that exist in `dl1`'s RS after initializing `dl2`: 4
```
After this change, the reading service that exist in each DataLoader2 is independent and will not impact each other.
```
Number of processes that exist in `dl1`'s RS after initializing `dl1`: 2
Number of processes that exist in `dl1`'s RS after initializing `dl2`: 2
```


torchdata/dataloader2/dataloader2.py
```diff
@@ -97,7 +111,7 @@ def __init__(
             self.datapipe_adapter_fns = datapipe_adapter_fn
         else:
             self.datapipe_adapter_fns = [datapipe_adapter_fn]
-        self.reading_service = reading_service
+        self.reading_service = deepcopy(reading_service)
```
ejguan, line 114: Let's try to import your PR to internal to validate if this is working properly with internal RS.

NivekT:
@nivekt has imported this pull request.  If you are a Facebook employee, you can view this diff [on Phabricator](https://www.internalfb.com/diff/D38868731).

VitalyFedyunin:
We should put this information into the ReadingService interface file too. To make sure that RS developers know about potential deep copy.

torchdata/dataloader2/dataloader2.py
VitalyFedyunin: +1 on checking existing implementations.

VitalyFedyunin:
One more though, we deepcopy RS, but we pickle-unpickle DP. I think we need to pick consistent approach.

ejguan:
> One more though, we deepcopy RS, but we pickle-unpickle DP. I think we need to pick consistent approach.

Totally agree. And, I support `deepcopy` because it's more readable and even working with unpickable objects.

NivekT:
> One more though, we deepcopy RS, but we pickle-unpickle DP. I think we need to pick consistent approach.

Can you elaborate on this? What make a consistent approach matter here?

NivekT:
Also, it seems like some internal ReadingService [cannot be deep-copied](https://fburl.com/testinfra/935swgsk).

ejguan:
> Also, it seems like some internal ReadingService [cannot be deep-copied](https://fburl.com/testinfra/935swgsk). One alternative is for users to pass in a constructor instead of the object itself, but that will change the interface quite a bit.

It seems the Error comes from construction of process group in `__init__` function. You might be able to move them into `initialize` function to make it working.

NivekT:
@nivekt has imported this pull request.  If you are a Facebook employee, you can view this diff [on Phabricator](https://www.internalfb.com/diff/D38868731).

NivekT:
@nivekt has imported this pull request.  If you are a Facebook employee, you can view this diff [on Phabricator](https://www.internalfb.com/diff/D38868731).

ejguan (approved):
Approving from OSS side. We might want @Miiira double check on the Diff.

And, could you please add a comment to `ReadingServiceInterface` that it's required to be deepcopied / picklable before `initialize` function.

NivekT:
@nivekt has imported this pull request.  If you are a Facebook employee, you can view this diff [on Phabricator](https://www.internalfb.com/diff/D38868731).

NivekT:
@nivekt has imported this pull request.  If you are a Meta employee, you can view this diff [on Phabricator](https://www.internalfb.com/diff/D38868731).

PR #749, opened 2022-08-22
Added default no-op function for flatmap

### LINKED ISSUE #738: Allow `flatmap` to have no-op option which will vertically flatten datapipe
```python
list(dp) # [[0,1,2],[3,4,5]]
dp = dp.flatmap()
list(dp) # [0,1,2,3,4,5]
```

This feels similar to `unbatch`, but should work with any iterable.

--

Looks like our final solution would be to allow `flatmap` to have no-op. 
Meanwhile, you can use:
```python
dp = dp.flatmap(fn = lambda x: x)
```

_Originally posted by @VitalyFedyunin in https://github.com/pytorch/data/issues/648#issuecomment-1217196626_

mobikoz:
Fixes #738 

### Changes

in datapipe's common functions:
- added _no_op_fn definition 

in FlatMapperIterDataPipe class
- added _no_op_fn method use for default callable use

### DIFF
--- a/torchdata/datapipes/iter/transform/callable.py
+++ b/torchdata/datapipes/iter/transform/callable.py
@@ -99,6 +99,7 @@ class FlatMapperIterDataPipe(IterDataPipe[T_co]):
 
     Note:
         The output from ``fn`` must be a Sequence. Otherwise, an error will be raised.
+        If no ``fn``, source DataPipe will be just flattened vertically.
 
     Args:
         datapipe: Source IterDataPipe
@@ -116,9 +117,11 @@ class FlatMapperIterDataPipe(IterDataPipe[T_co]):
     datapipe: IterDataPipe
     fn: Callable
 
-    def __init__(self, datapipe: IterDataPipe, fn: Callable, input_col=None) -> None:
+    def __init__(self, datapipe: IterDataPipe, fn: Callable = None, input_col=None) -> None:
         self.datapipe = datapipe
 
+        if fn is None:
+            fn = self._no_op_fn
         _check_unpickable_fn(fn)
         self.fn = fn  # type: ignore[assignment]
         self.input_col = input_col
@@ -138,6 +141,9 @@ class FlatMapperIterDataPipe(IterDataPipe[T_co]):
 
     def __len__(self) -> int:
         raise TypeError(f"{type(self).__name__}'s length relies on the output of its function.")
+
+    def _no_op_fn(self, e):
+        return e
 
 
 @functional_datapipe("drop")
--- a/torchdata/datapipes/iter/transform/callable.py
+++ b/torchdata/datapipes/iter/transform/callable.py
@@ -9,6 +9,7 @@ from typing import Callable, Hashable, Iterator, List, Optional, Set, Sized, Typ
 
 from torch.utils.data import functional_datapipe, IterDataPipe
 from torch.utils.data.datapipes.utils.common import _check_unpickable_fn
+from torchdata.datapipes.utils.common import _no_op_fn
 
 T_co = TypeVar("T_co", covariant=True)
 
@@ -99,7 +100,7 @@ class FlatMapperIterDataPipe(IterDataPipe[T_co]):
 
     Note:
         The output from ``fn`` must be a Sequence. Otherwise, an error will be raised.
-        If no ``fn``, source DataPipe will be just flattened vertically.
+        If ``fn`` is ``None``, source DataPipe will be just flattened vertically.
 
     Args:
         datapipe: Source IterDataPipe
@@ -113,6 +114,11 @@ class FlatMapperIterDataPipe(IterDataPipe[T_co]):
         >>> flatmapped_dp = source_dp.flatmap(fn)
         >>> list(flatmapped_dp)
         [0, 0, 1, 10, 2, 20, 3, 30, 4, 40]
+        >>>
+        >>> source_dp = IterableWrapper([[1, 2, 3], [4, 5, 6]])
+        >>> flatmapped_dp = source_dp.flatmap()
+        >>> list(flatmapped_dp)
+        [1, 2, 3, 4, 5, 6]
     """
     datapipe: IterDataPipe
     fn: Callable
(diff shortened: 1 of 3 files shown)


torchdata/datapipes/iter/transform/callable.py
```diff
@@ -139,6 +142,9 @@ def __iter__(self) -> Iterator[T_co]:
     def __len__(self) -> int:
         raise TypeError(f"{type(self).__name__}'s length relies on the output of its function.")
 
+    def _no_op_fn(self, e):
+        return e
```
NivekT, line 146: We likely want this as a static method or a function defined outside of the class

torchdata/datapipes/iter/transform/callable.py
```diff
@@ -99,6 +99,7 @@ class FlatMapperIterDataPipe(IterDataPipe[T_co]):
 
     Note:
         The output from ``fn`` must be a Sequence. Otherwise, an error will be raised.
+        If no ``fn``, source DataPipe will be just flattened vertically.
```
NivekT, line 102:
```suggestion
        If ``fn`` is ``None``, source DataPipe will be just flattened vertically.
```
super nit

NivekT (commented):
Thanks for contributing this PR. Just some nit comments.

mobikoz:
Waiting for company CLA signature, should be shortly

NivekT (commented):
One last thing - can you add a unit test around here?

https://github.com/pytorch/data/blob/main/test/test_iterdatapipe.py#L692

Thanks!

mobikoz:
Unit test added, also I fixed _no_op_fn to always return iterables.

torchdata/datapipes/utils/common.py
```diff
@@ -22,3 +22,13 @@ def validate_pathname_binary_tuple(data: Tuple[str, IOBase]):
...
+
+
+def _no_op_fn(*args):
+    """
+    No-operation function, returns passed arguments , always as iterable.
+    """
+    if len(args) == 1 and isinstance(args[0], (list, tuple)):
+        return args[0]
+    else:
+        return args
```
NivekT, line 34:
```suggestion
    if len(args) == 1:
        try:
            return iter(args[0])
        except TypeError:
            pass
    return args
```
@mobikoz What do you think of this instead? This will allow objects that are not `list` or `tuple` to be unpacked (e.g. `numpy.array`).

cc: @ejguan

NivekT (commented):
See comment about `_no_op_fn`, let me know what you think.

torchdata/datapipes/utils/common.py
ejguan: Can I know why do we want to differentiate the cases between size 1 and other sizes?
```py
list(dp)  # [[1], [2], [3], [4],]
list(dp.flatmap())  # I would expect the result as [1, 2, 3, 4]
```

BTW, if this function is not commonly used by other DataPipes, we can put it in the `callable.py` file.

NivekT: > Can I know why do we want to differentiate the cases between size 1 and other sizes?

As far as I observe, because the input to the function is `*args`, there are two main cases:
1. `input_col=None`
2. `input_col` only specifies one column

mobikoz: @ejguan   This differentiation is the consequence of using variable length *args for the sake of input_col parameter use.
If not handled, in case of given example: 
'
list(dp)  # [[1], [2], [3], [4],]
list(dp.flatmap())  #  result will be [[1], [2], [3], [4],]
'
This is because args would be always of type tuple, in this case for eg. ([1],) for the first element.

ejguan: I see, makes sense. Thanks for the explanation.

NivekT: I guess the only thing that is does is that if `arg[0]` isn't iterable (can't be unpacked), it returns `args`.

But in this case that is a kind of silent failure, which I think is worse. So, my bad. Let's do this instead:
```python
    if len(args) == 1:
        return args[0]
    return args
```

Sorry about this @mobikoz. After this, we should be about to merge the change. Thanks!

NivekT (approved):
See comment for one last change and we will merge. Thank you!

torchdata/datapipes/utils/common.py
mobikoz: I believe that currently committed implementation is correct - situation when arg[0] is not iterable seems to be common and TypeError will be rised. 

Consider example
`list(dp) # [[1, 2, 3], [4, 5, 6]]`
`list(dp.flatmap(input_col=1)) # TypeError will be raised`

NivekT: My point is that the `TyperError` should be raised for your example (but it doesn't in the current implementation because it catches the exception). 
Otherwise, it would be a silent failure, because the DataPipe promises to unpack the element but it actually didn't and just returned the element itself.

mobikoz: Currently it will work like that:
`list(dp) # [[1, 2, 3], [4, 5, 6]]`
`list(dp.flatmap(input_col=1) # [2, 5]`


`list(dp) # [1, 2, 3, 4, 5, 6]`
`list(dp.flatmap(input_col=1) # [1, 2, 3, 4, 5, 6]`

Isn't it expected behavior ? 
Just double checking if I understand correctly.

mobikoz: btw it is analogous behavior to calling flatmap with some fn, for eg 
`def fn(e):`
`    return` [e, e * 10]`

NivekT: I might be wrong but you might want to do this if all you want is to select an index?

```python
from operator import itemgetter
dp = IterableWrapper([[1, 2, 3], [4, 5, 6]]).map(itemgetter(1))
list(dp)  # [2, 5]
```

The case I worry about is something like this:
```python
dp = IterableWrapper([[1, [2, 2], 3], [4, [5, 5], 6], [7, 8, 9], [10, [11, 11], 12]]).flatmap(input_col=1)
```
The 3rd list will silently unpack to `8` even though it is different from the other lists.

@ejguan any thoughts on which behavior is better?

ejguan: How about this?
```py
list(dp) # [[1, [2], 3], [4, [5], 6]]
list(dp.flatmap(input_col=1) # I expect [2, 5] rather than [[2], [5]]
```

I feel like it's fine to remove `iter` from the currently implementation as long as we make it clear in the comment that we would only allow the output being unpackable by `yield from`

mobikoz: Pushed corrected implementation

PR #765, opened 2022-09-08
add lazy test to iter2map_converter

### LINKED ISSUE #753: `to_map_datapipe` doesn't behave lazily loading in multiprocessing
### 🐛 Describe the bug

At this [line](https://github.com/pytorch/data/blob/719616a1b4791034da3d888357e3ef62c70806e3/torchdata/datapipes/iter/util/converter.py#L118-L119), when it's sent to child process, it will automatically load all data into memory. It leads to multiple problems:
- Data is not lazily loaded. It would take more time and more memory during the construction of multiprocessing
- The sharding is not going to be applied properly, as we are doing sharding after the worker processes are constructed.

Solution is straightforward. We should just remove it those lines from `to_map_datapipe`.

### Versions

main

DongyuXu77:
Fixes #753 

This PR adds a test to ensure the converter behaves lazily in multiprocessing by checking the._map attribute after loading data.


test/test_iter2map_converter.py
```diff
@@ -0,0 +1,41 @@
...
+
+from torchdata.dataloader2 import (
+        DataLoader2,
+        MultiProcessingReadingService,
+        )
+from torchdata.datapipes.iter import IterableWrapper
+from torchdata.datapipes.map import MapDataPipe
+
+
+class Test_iter2map_converter(unittest.TestCase):
```
ejguan, line 21: Could you please add this test to `test/test_dataloader2.py` as a separate class called `DataLoader2IntegrationTest`?

test/test_iter2map_converter.py
```diff
@@ -0,0 +1,41 @@
...
+
+    def test_lazy_load(self) -> None:
+        source_dp = IterableWrapper([(i, i) for i in range(10)])
+        map_dp = source_dp.to_map_datapipe()
+        dl: DataLoader2 = DataLoader2(datapipe=map_dp, reading_service=self._get_multi_reading_service())
+        # Lazy loading
+        self.assertTrue(dl.datapipe._map is None)
+        
+        for _ in dl:
+            pass
```
ejguan, line 34: How about this
```suggestion
        it = iter(dl)
        self.assertTrue(dl.datapipe._map is None)  # The datapipe in main process should be 
        self.assertEqual(list(it), list(range(10)))
```

ejguan (changes requested):
Thank you for adding the patch. I just left a few comments below regarding test.

test/test_dataloader2.py
```diff
@@ -219,5 +219,25 @@ def test_dataloader2_shuffle(self) -> None:
...
+        map_dp = source_dp.to_map_datapipe()
+        dl: DataLoader2 = DataLoader2(datapipe=map_dp, reading_service=self._get_multi_reading_service())
+        # Lazy loading
+        self.assertTrue(dl.datapipe._map is None)
+        
+        for _ in dl:
+            pass
+
+        # Lazy loading in multprocessing
+        self.assertTrue(dl.datapipe.__dict__['iterable'].dataset.__dict__['_datapipe']._map is None)
```
NivekT, line 239: Question: What is this line intending to do? Also I think it will not work when `MultiProcessingReadingService` gets replaced with the `Prototype` one.

DongyuXu77: This line intends to get the `dl.datapipe._map` after multiprocessing.  But `dl.datapipe` changed after the `iter` function was called, so I can't use `dl.datapipe._map` to get the map attribute directly.

Yes, It can't work in `PrototyeMultiprocessingReadingService`. If it's needed, I will try to fix it.

DongyuXu77: I think I know what you mean. 
In main process, the datapipe._map attribute should be None as you mentioned before. 
But I can't run `test_dataloader2.py`, it will get an error. Is it OK?

Besides, when I set `reading_service` to `PrototypeMultiprocessingReadingService`, It will get an exception saying Unrecognized type of request received. I think that's because some function doesn't realize in `Prototype` one, but I'm not sure.

ejguan: > But I can't run `test_dataloader2.py`, it will get an error. Is it OK?

Emm. It's not Okay since the Error apparently comes from the test. And, I am sorry that I forget to consider that `ReadingService` will adapt the DataPipe. So, here is my suggestion to the test.
```py
            # Lazy loading
            self.assertTrue(dl.datapipe._map is None)
            it = iter(dl)
            # Lazy loading in multiprocessing
            self.assertTrue(map_dp._map is None)
```

We can use `MultiprocessingReadingService` for now.

ejguan:
And for lint Error, can you please follow the instruction [here](https://github.com/pytorch/data/blob/main/CONTRIBUTING.md#code-style) to fix it?

test/test_dataloader2.py
DongyuXu77: OK, I will try it now

DongyuXu77:
> And for lint Error, can you please follow the instruction [here](https://github.com/pytorch/data/blob/main/CONTRIBUTING.md#code-style) to fix it?

Ok, thanks to your patient guidance. I will do it now

test/test_dataloader2.py
DongyuXu77: I find it needs to loading data after `iter` because `iter` is just an init function and until loading data the multiprocessing will work( I guess).

If remove the loading data operation `self.assertEqual(list(it), list(range(10)))`, the test function will not work properly since I restore the deleted line in `converter.py` it still could pass.

ejguan (approved):
LGTM, thank you

ejguan:
The failing tests are not related to your PR. Will merge the PR shortly.

DongyuXu77:
> LGTM, thank you

Thanks to your and NivekT guidance, I can't use words to express my gratitude.

NivekT (approved):

PR #774, opened 2022-09-15
Fix TorchVision tests caused by file renaming

ejguan:
This PR is required since https://github.com/pytorch/vision/pull/6587 has been landed and shipped to TorchVision nightly.

### DIFF
--- a/.github/workflows/domain_ci.yml
+++ b/.github/workflows/domain_ci.yml
@@ -56,7 +56,7 @@ jobs:
           path: vision
 
       - name: Run torchvision builtin datasets tests
-        run: pytest --no-header -v vision/test/test_prototype_builtin_datasets.py
+        run: pytest --no-header -v vision/test/test_prototype_datasets_builtin.py
 
   torchtext:
     if: ${{ github.repository_owner == 'pytorch' }}


.github/workflows/domain_ci.yml
```diff
@@ -56,7 +56,7 @@ jobs:
           path: vision
 
       - name: Run torchvision builtin datasets tests
-        run: pytest --no-header -v vision/test/test_prototype_builtin_datasets.py
+        run: pytest --no-header -v vision/test/test_prototype_datasets*.py
```
pmeier, line 59: You probably want

```suggestion
        run: pytest --no-header -v vision/test/test_prototype_datasets_builtin.py
```

Otherwise you will also run our utility tests and the like that are unrelated to `torchdata`.

pmeier (approved):
Thanks, I totally forgot!

.github/workflows/domain_ci.yml
pmeier: Or if you actually want to run all tests, you need to use 

```suggestion
        shell: bash
        run: pytest --no-header -v vision/test/test_prototype_datasets*.py
```

to be able to use the `*.py` globbing

ejguan: Sounds great. Thank you for the suggestion.

NivekT (approved):

PR #777, opened 2022-09-16
Enable 3.10 tests & fix bad protobuf version for MacOS Python 3.10

### LINKED ISSUE #775: Failing CI tests on MacOS with Python 3.10 caused by protobuf
### 🐛 Describe the bug

The nightly builds were failing during the last two days caused by an update from protobuf. 
See:
https://github.com/pytorch/data/actions/runs/3059924230/jobs/4937858576#step:13:272

This is caused by the issue from https://github.com/protocolbuffers/protobuf/issues/10571

### Versions

main branch

ejguan:
Fixes #775

### Changes

- Enable CI for Python 3.10
- Fix the version of protobuf in test to 3.20.1 when it's MacOS and Python 3.10

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -30,6 +30,7 @@ jobs:
           - 3.7
           - 3.8
           - 3.9
+          - "3.10"
     steps:
       - name: Get PyTorch Channel
         shell: bash
--- a/test/requirements.txt
+++ b/test/requirements.txt
@@ -5,5 +5,8 @@ s3fs
 iopath == 0.1.9
 numpy
 rarfile
-protobuf < 4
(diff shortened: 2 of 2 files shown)


.github/workflows/ci.yml
```diff
@@ -30,6 +30,7 @@ jobs:
           - 3.7
           - 3.8
           - 3.9
+          - "3.10"
```
NivekT, line 33: n00b question: is there a reason why it needs the quotation marks around 3.10 but not the others?

NivekT (approved):
LGTM!

.github/workflows/ci.yml
ejguan: If not, Github actions will treat `3.10` as `3.1` based on my prior experience.

PR #780, opened 2022-09-19
Add type hint stub

### LINKED ISSUE #779: Include `py.typed` marker file for mypy type hint compatibility
### 🚀 The feature

From what I can tell torchdata actually had type stubs but mypy won't pick it up because it is missing the `py.typed` marker file (see: https://mypy.readthedocs.io/en/stable/running_mypy.html#missing-library-stubs-or-py-typed-marker)

The ask is to add `py.typed` to the package data similar to how [torch does it](https://github.com/pytorch/pytorch/blob/d05a11337c8aafb663ca3b29722c5219d1589fec/setup.py#L991) so that projects using torchdata can have mypy recognize torchdata modules type hints.

### Motivation, pitch

My project uses torchdata and also uses mypy for static type checking. Since mypy does not pick up torchdata's type stubs I get an error as:

```
mfive/datapipes/amzn/datakit.py:9:1: error: Skipping analyzing "torchdata.datapipes": module is installed, but missing library stubs or py.typed marker  [import]
```

### Alternatives

I could add a `ignore-missing-imports` in my `mypy.ini` file:

```
[mypy-torchdata.*]
ignore_missing_imports = True
```

But that means that all the torchdata related types will be typed as `Any`, hence not letting me enable checks like `disallow_subclassing_any=True` if I'm writing my own custom datapipe that subclasses `torchdata.datapipes.iter.IterDataPipe`.

### Additional context

_No response_

ejguan:
Fixes #779

### Changes

- Add `py.typed` file stub
- Include the file into package data

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -182,6 +182,7 @@ if __name__ == "__main__":
         ],
         package_data={
             "torchdata": [
+                "py.typed",
                 "datapipes/iter/*.pyi",
                 "datapipes/map/*.pyi",
             ],


VitalyFedyunin (approved):

NivekT (approved):

pmeier:
This breaks `torchvision` pretty hard: https://app.circleci.com/pipelines/github/pytorch/vision/20354/workflows/861e7e2c-4e0b-4594-9d81-33f0b2d27f43/jobs/1655543

```
Found 299 errors in 26 files (checked 219 source files)
```

I haven't looked in detail, but some of the errors don't seem to be a usage error. I'll keep you posted.

pmeier:
An example where I don't think we are responsible

```py
from typing import BinaryIO, Tuple
import pathlib

from torchdata.datapipes.iter import FileLister, FileOpener, IterDataPipe, IterableWrapper


def load(path: pathlib.Path) -> IterDataPipe[Tuple[str, BinaryIO]]:
    if path.is_dir():
        dp = FileLister(str(path), recursive=True)
    else:
        dp = IterableWrapper([str(path)])
    return FileOpener(dp, mode="rb")
```

Running `mypy` on this yields

```
main.py:11: error: Incompatible types in assignment (expression has type "IterableWrapperIterDataPipe",
variable has type "FileListerIterDataPipe")  [assignment]
            dp = IterableWrapper([str(path)])
                 ^
main.py:12: error: Incompatible return value type (got "FileOpenerIterDataPipe", expected
"IterDataPipe[Tuple[str, BinaryIO]]")  [return-value]
```

The actual function is a little more complicated and we can't use `FileOpener` as return type, whereas `IterDataPipe[Tuple[str, BinaryIO]]` is true for all cases.

Could you advise how you want us to annotate? Can we maybe revert this PR and resolve all issues that this brings to torchvision before we merge a second time?

ejguan:
@pmeier 
At list for the first Error, the proper typing should be:
```py
def load(path: pathlib.Path) -> IterDataPipe[Tuple[str, BinaryIO]]:
    if path.is_dir():
        dp: IterDataPipe = FileLister(str(path), recursive=True)
    else:
        dp = IterableWrapper([str(path)])
    return FileOpener(dp, mode="rb")
```

However, even with the proper typing shown above, the Error is changed to `Incompatible types in assignment (expression has type "FileListerIterDataPipe", variable has type "IterDataPipe[Any]")`. And, it doesn't explain what causes the second Error.
So, I spent a few hours figuring out what is the root cause of the `mypy` Error. In the generated `datapipe.pyi` file, a new `IterDataPipe` class is defined, which overrides the original `IterDataPipe` from the inheritance graph for all other `DataPipe`.

All Errors are eliminated when I remove new `IterDataPipe` definition from `datapipe.pyi`  and import `IterDataPipe` directly from `torch.utils.data.datapipe`. And, the reason we define new `IterDataPipe` in `pyi` file is attaching all functional APIs to it. We need to do it in a different way by keeping the original `IterDataPipe` and extending the class with all functional APIs.

cc: @NivekT for python interface file

For this PR, I will revert it because our typing system needs to be fixed generally.

PR #789, opened 2022-09-24
add `padden_tokens` so max_token_count is not exceeded

### LINKED ISSUE #788: MaxTokenBucketizer does not guarentee `max_token_count` holds if padding is used
### 🐛 Describe the bug

The  `MaxTokenBucketizer` data pipe should probably be aware whether users will pad the resulting list of data to equal length.  As the description gives an example in the audio domain, where padding is frequently used to create a batch of equal length sequences, users will probably run into edge cases where the current behavior will create OOM errors. 

For example, assume a dataset of 5 sequences, with length  [100, 10, 10, 10, 100], and the requirement that no batch has more than 150 tokens. 

The desired behavior would be be 3 batches: [100], [10, 10, 10], and [100]. The current behavior leads to the following:

```python3
import torch as t

from torchdata.datapipes.iter import (
    MaxTokenBucketizer,
    IterableWrapper,
    Mapper,
)


class Sample:
    def __init__(self, length: int):
        self.seq = t.ones((length,))

    def __len__(self):
        return self.seq.shape[0]

    def __lt__(self, other):
        len(self).__lt__(len(other))


def sample_to_tensor(bucket):
    return [x.seq for x in bucket]


def pad_bucket(bucket):
    return t.nn.utils.rnn.pad_sequence(bucket, batch_first=True)


data_lengths = [100, 10, 10, 10, 100]
data = [Sample(x) for x in data_lengths]

dp = IterableWrapper(data)
dp = MaxTokenBucketizer(dp, max_token_count=150)
dp = Mapper(dp, sample_to_tensor)
dp = Mapper(dp, pad_bucket)

for x in dp:
    print(x.shape, f"num_tokens:{x.shape[0] * x.shape[1]}")
```
prints:
```
torch.Size([4, 100]) num_tokens:400
torch.Size([1, 100]) num_tokens:100
```

### Versions

PyTorch version: 1.12.1+cu113
Is debug build: False
CUDA used to build PyTorch: 11.3

OS: Arch Linux (x86_64)
GCC version: (GCC) 12.2.0
Clang version: Could not collect
CMake version: Could not collect
Libc version: glibc-2.36

Python version: 3.10.0 (default, Sep 21 2022, 22:36:10) [GCC 12.2.0] (64-bit runtime)
Python platform: Linux-5.19.10-arch1-1-x86_64-with-glibc2.36
Is CUDA available: True
CUDA runtime version: Could not collect
GPU models and configuration: GPU 0: NVIDIA GeForce RTX 3070
Nvidia driver version: 515.76
cuDNN version: Could not collect
HIP runtime version: N/A
MIOpen runtime version: N/A
Is XNNPACK available: True

[pip3] mypy-extensions==0.4.3
[pip3] numpy==1.23.3
[pip3] pytorch-lightning==1.7.7
[pip3] torch==1.12.1+cu113
[pip3] torchaudio==0.12.1+cu113
[pip3] torchdata==0.4.1
[pip3] torchmetrics==0.9.3
[pip3] torchvision==0.13.1+cu113
[conda] Could not collect

nikvaessen:
Fixes #788 

### Changes

- Add `include_padding` argument to `MaxTokenBucketizer`

### DIFF
--- a/test/test_iterdatapipe.py
+++ b/test/test_iterdatapipe.py
@@ -714,7 +714,7 @@ class TestIterDataPipe(expecttest.TestCase):
         # Functional test: Padded tokens not exceeding max_token_count
         source_data = ["111", "1111", "11111"]  # 3, 4, 5
         source_dp = IterableWrapper(source_data)
-        batch_dp = source_dp.max_token_bucketize(max_token_count=7, padded_tokens=True)
+        batch_dp = source_dp.max_token_bucketize(max_token_count=7, count_padding=True)
         exp_batch = [["111"], ["1111"], ["11111"]]
         self.assertEqual(list(batch_dp), exp_batch)
 
--- a/test/test_iterdatapipe.py
+++ b/test/test_iterdatapipe.py
@@ -712,7 +712,7 @@ class TestIterDataPipe(expecttest.TestCase):
         self.assertEqual(list(batch_dp), exp_batch)
 
         # Functional test: Padded tokens not exceeding max_token_count
-        source_data = ["111", "111", "111", "1111", ]  # 3, 3, 3, 4
+        source_data = ["111", "111", "111", "1111"]  # 3, 3, 3, 4
         source_dp = IterableWrapper(source_data)
         batch_dp = source_dp.max_token_bucketize(max_token_count=7, include_padding=True)
         exp_batch = [["111", "111"], ["111"], ["1111"]]
--- a/test/test_iterdatapipe.py
+++ b/test/test_iterdatapipe.py
@@ -712,10 +712,10 @@ class TestIterDataPipe(expecttest.TestCase):
         self.assertEqual(list(batch_dp), exp_batch)
 
         # Functional test: Padded tokens not exceeding max_token_count
-        source_data = ["111", "1111", "11111"]  # 3, 4, 5
+        source_data = ["111", "111", "111", "1111", ]  # 3, 3, 3, 4
         source_dp = IterableWrapper(source_data)
-        batch_dp = source_dp.max_token_bucketize(max_token_count=7, count_padding=True)
-        exp_batch = [["111"], ["1111"], ["11111"]]
+        batch_dp = source_dp.max_token_bucketize(max_token_count=7, include_padding=True)
+        exp_batch = [["111", "111"], ["111"], ["1111"]]
         self.assertEqual(list(batch_dp), exp_batch)
 
         # Functional test: sample length exceeding max_token_count
--- a/test/test_iterdatapipe.py
+++ b/test/test_iterdatapipe.py
@@ -703,6 +703,20 @@ class TestIterDataPipe(expecttest.TestCase):
         self.assertEqual(res_before_reset, exp_before_reset)
         self.assertEqual(res_after_reset, exp_after_reset)
 
+        # Functional test: Padded tokens exceeding max_token_count
+        source_data = ["111", "1111", "11111"]  # 3, 4, 5
+        source_dp = IterableWrapper(source_data)
+        batch_dp = source_dp.max_token_bucketize(max_token_count=6)
+        exp_batch = [['111', '1111'], ['11111']]
+        self.assertEqual(list(batch_dp), exp_batch)
+
+        # Functional test: Padded tokens not exceeding max_token_count
+        source_data = ["111", "1111", "11111"]  # 3, 4, 5
+        source_dp = IterableWrapper(source_data)
+        batch_dp = source_dp.max_token_bucketize(max_token_count=6, padded_tokens=True)
+        exp_batch = [['111'], ['1111'], ['11111']]
+        self.assertEqual(list(batch_dp), exp_batch)
+
         # __len__ Test: returns the number of batches
         with self.assertRaises(TypeError):
             len(batch_dp)
--- a/test/test_iterdatapipe.py
+++ b/test/test_iterdatapipe.py
@@ -706,15 +706,16 @@ class TestIterDataPipe(expecttest.TestCase):
         # Functional test: Padded tokens exceeding max_token_count
         source_data = ["111", "1111", "11111"]  # 3, 4, 5
         source_dp = IterableWrapper(source_data)
-        batch_dp = source_dp.max_token_bucketize(max_token_count=6)
-        exp_batch = [['111', '1111'], ['11111']]
+        batch_dp = source_dp.max_token_bucketize(max_token_count=7)
+        exp_batch = [["111", "1111"], ["11111"]]
+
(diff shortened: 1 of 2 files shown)


nikvaessen:
I'm unsure what the best name for the proposed argument is.

torchdata/datapipes/iter/transform/bucketbatcher.py
```diff
@@ -204,13 +204,25 @@ class MaxTokenBucketizerIterDataPipe(IterDataPipe[DataChunk[T_co]]):
...
+
+
     Args:
         datapipe: Iterable DataPipe being batched
         max_token_count: Maximum length of total length of data in each batch
         len_fn: Function to be applied to each element to get lengths. ``len(data)`` is used by default.
         min_len: Optional minimum length to be included into each batch
         max_len: Optional maximum length to be included into each batch.
         buffer_size: This restricts how many tokens are taken from prior DataPipe to bucketize
+        count_padding: If true, assume data will be padded to the largest length in batch in succeeding DataPipes.
```
ejguan, line 224: Maybe `include_padding`?

torchdata/datapipes/iter/transform/bucketbatcher.py
```diff
@@ -253,28 +266,44 @@ def __init__(
...
         buffer: List = []
         batch: List = []
         batch_size: int = 0
+        max_length: int = 0
         for d in self.datapipe:
             heapq.heappush(buffer, d)
             if len(buffer) == self.buffer_size:
                 length, token = heapq.heappop(buffer)
-                if batch_size + length > self.max_token_count:
+                max_length = max(length, max_length)
```
ejguan, line 280: This can be moved under `if self.count_padding:`

torchdata/datapipes/iter/transform/bucketbatcher.py
```diff
@@ -253,28 +266,44 @@ def __init__(
...
+                    batch = [token]
+                    batch_size = length
+                    max_length = length
+                else:
+                    batch.append(token)
+                    batch_size = new_batch_size
         while buffer:
             length, token = heapq.heappop(buffer)
-            if batch_size + length > self.max_token_count:
+            max_length = max(length, max_length)
```
ejguan, line 295: Ditto

torchdata/datapipes/iter/transform/bucketbatcher.py
```diff
@@ -253,28 +266,44 @@ def __init__(
...
-                    batch = []
-                    batch_size = 0
-                batch.append(token)
-                batch_size += length
+                    batch = [token]
+                    batch_size = length
+                    max_length = length
+                else:
+                    batch.append(token)
+                    batch_size = new_batch_size
```
ejguan, line 292: I feel we can make it as a function to reduce duplicated code.

test/test_iterdatapipe.py
```diff
@@ -703,6 +703,28 @@ def _custom_len_fn(token):
...
+        exp_batch = [["111", "1111"], ["11111"]]
+
+        self.assertEqual(list(batch_dp), exp_batch)
+
+        # Functional test: Padded tokens not exceeding max_token_count
+        source_data = ["111", "1111", "11111"]  # 3, 4, 5
+        source_dp = IterableWrapper(source_data)
+        batch_dp = source_dp.max_token_bucketize(max_token_count=7, count_padding=True)
+        exp_batch = [["111"], ["1111"], ["11111"]]
+        self.assertEqual(list(batch_dp), exp_batch)
```
ejguan, line 719: Could you please change this test to make sure part of input can be aggregated into a batch but part can not?

torchdata/datapipes/iter/transform/bucketbatcher.py
```diff
@@ -204,13 +204,25 @@ class MaxTokenBucketizerIterDataPipe(IterDataPipe[DataChunk[T_co]]):
...
     ``max_token_count``, each batch may be concatenated to a Tensor with the same size and minimum padding.
 
+    If ``include_padding`` is set to ``True``, the token count of each batch includes the padding a succeeding
+    DataPipe could add. This guarentees that even after the batch is padded, ``max_token_count`` will not be exceeded.
+    This can prevent out-of-memory issues for data with large variations in length.
+
+    Note that batches are bucketized starting from the smallest size in a buffer.
+    This can limit the variablity of batches if ``buffer_size`` is large.
+    To increase variablity, apply ``torchdata.datapipes.iter.Shuffler`` before this DataPipe,
+    and keep ``buffer_size`` small.
```
ejguan, line 214: apply `shuffler` before **and after** this DataPipe.

torchdata/datapipes/iter/transform/bucketbatcher.py
```diff
@@ -253,28 +266,45 @@ def __init__(
...
+            new_batch_size = batch_size + length
+
+        if new_batch_size > self.max_token_count:
+            data_chunk_to_yield = DataChunk(batch)
+            batch = [token]
+            batch_size = length
+            max_length = length
+        else:
+            batch.append(token)
+            batch_size = new_batch_size
```
ejguan, line 308: `max_length` needs to be reset to `length` here.

torchdata/datapipes/iter/transform/bucketbatcher.py
```diff
@@ -204,13 +204,25 @@ class MaxTokenBucketizerIterDataPipe(IterDataPipe[DataChunk[T_co]]):
...
+
+
     Args:
         datapipe: Iterable DataPipe being batched
         max_token_count: Maximum length of total length of data in each batch
         len_fn: Function to be applied to each element to get lengths. ``len(data)`` is used by default.
         min_len: Optional minimum length to be included into each batch
         max_len: Optional maximum length to be included into each batch.
         buffer_size: This restricts how many tokens are taken from prior DataPipe to bucketize
+        include_padding: If true, assume data will be padded to the largest length in batch in succeeding DataPipes.
```
ejguan, line 224: This doesn't reflect the fact that this function will compare the padded length with the `max_token_count`.

ejguan (commented):
Overall good. A few more comments below. Thank you

torchdata/datapipes/iter/transform/bucketbatcher.py
nikvaessen: `max_length` should be `max(len(x) for x in batch)` at this point. I think that holds with the code as is. If `self.include_padding` is `False`, it will always stay 0 and unused. If `self.include_padding=True`, `max_length` will be contain the value of `max(length, previous_max_length)` (line 296) when we **do not** yield a batch. Or `max_length` will be equal to `length` if we **do** yield a batch.

ejguan (approved):

NivekT (approved):

PR #830, opened 2022-10-13
Update gha version to utilize Node16

ejguan:
This PR would remove all the annoying warnings from all GHA workflows.
See more detail in https://github.blog/changelog/2022-09-22-github-actions-all-actions-will-begin-running-on-node16-instead-of-node12/


NivekT (approved):
LGTM!

PR #833, opened 2022-10-14
[DataLoader2] Skip wrapping if serialization wrapper attached & deepcopy DataPipe in load_state_dict

ejguan:
Per title
- We don't need to attach serialization wrapper if the last DataPipe has been `_DataPipeSerializationWrapper`
- When we `load_state_dict`, we still need a copy of `self.datapipe` as it has been done in `__init__` function. We should align the behavior of `load_state_dict` and `__init__`
  - Either deepcoy to `self._datapipe_before_reading_service_adapt`
  - Or skip deepcopy to `self._datapipe_before_reading_service_adapt` at those two places.

### DIFF
--- a/torchdata/dataloader2/dataloader2.py
+++ b/torchdata/dataloader2/dataloader2.py
@@ -9,7 +9,11 @@ import pickle
 from dataclasses import dataclass
 from typing import Any, Dict, Generic, Iterable, Iterator, Optional, TypeVar, Union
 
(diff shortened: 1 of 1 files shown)


ejguan:
> LGTM! Question: do we expect any case where the `datapipe` being passed in already has a wrapper?

It's more like a future proof. (Technically, if users manually `deserialize` the DataPipe state, the returned DataPipe is wrapped by the wrapper.

PR #812, opened 2022-10-06
Adding Cloud Storage Provider tutorial section

NivekT:
Stack from [ghstack](https://github.com/ezyang/ghstack):
* (to be filled)

PR #836, opened 2022-10-18
Add docs on accessing Azure blob storage through fsspec

sgrigory:
### Changes

Adding an example of DataPipe usage with Azure Blob storage via `fsspec`, similar to https://github.com/pytorch/data/pull/812. The example is placed into a new section in `docs/source/tutorial.rst`

Here is the screenshot showing that code snippets in the tutorial work as expected:

Screenshot 2022-10-18 at 19 33 49

####  Minor note

Technically, `fsspec` [allows both path prefixes `abfs://` or `az://`](https://github.com/fsspec/adlfs/blob/f15c37a43afd87a04f01b61cd90294dd57181e1d/README.md?plain=1#L33) for Azure Blob storage Gen2 as synonyms. However, only `abfs://` works for us for the following reason:
- If a path starts with `az`, the variable `fs.protocol` [here](https://github.com/pytorch/data/blob/768ecdae8b56af640a78e29f82864dc4f65df371/torchdata/datapipes/iter/load/fsspec.py#L82) is still `abfs`
- So the condition `root.startswith(protocol)` is false, and `is_local` is true
- As a result the path "doubles" in [this line](https://github.com/pytorch/data/blob/768ecdae8b56af640a78e29f82864dc4f65df371/torchdata/datapipes/iter/load/fsspec.py#L95), like on this screenshot:
Screenshot 2022-10-18 at 19 50 56

This won't have any effect for the users, however, as long as they use `abfs://` prefix recommended in the tutorial

### DIFF
--- a/docs/source/tutorial.rst
+++ b/docs/source/tutorial.rst
@@ -422,7 +422,5 @@ directory ``curated/covid-19/ecdc_cases/latest``, belonging to account ``pandemi
     # [['date_rep', 'day', ..., 'iso_country', 'daterep'], 
     # ['2020-12-14', '14', ..., 'AF', '2020-12-14'],
     # ['2020-12-13', '13', ..., 'AF', '2020-12-13']]
-
-
-If necessary, you can also access data in Azure Data Lake Storage Gen1 by using URIs staring with
- ``adl://`` and ``abfs://``, as described in `README of adlfs repo <https://github.com/fsspec/adlfs/blob/main/README.md>`_
+If necessary, you can also access data in Azure Data Lake Storage Gen1 by using URIs staring with 
+``adl://`` and ``abfs://``, as described in `README of adlfs repo <https://github.com/fsspec/adlfs/blob/main/README.md>`_
--- a/docs/source/tutorial.rst
+++ b/docs/source/tutorial.rst
@@ -401,7 +401,7 @@ can be used to list files in a directory in a container:
     from torchdata.datapipes.iter import IterableWrapper
 
     storage_options={'account_name': ACCOUNT_NAME, 'account_key': ACCOUNT_KEY}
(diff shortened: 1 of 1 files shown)


ejguan (approved):
This is great! Thank you. And, do you mind opening an issue about your minor note? I think it's a bug.

And, adding original author @NivekT of this tutorial to review.

docs/source/tutorial.rst
```diff
@@ -298,7 +298,7 @@ recommend using the functional form of DataPipes.
 Working with Cloud Storage Providers
 ---------------------------------------------
 
-In this section, we show examples accessing AWS S3 and Google Cloud Storage with built-in``fsspec`` DataPipes.
+In this section, we show examples accessing AWS S3 and Google Cloud Storage with built-in ``fsspec`` DataPipes.
```
NivekT, line 301:
```suggestion
In this section, we show examples accessing AWS S3, Google Cloud Storage, and Azure Cloud Storage with built-in ``fsspec`` DataPipes.
```

NivekT (approved):
One nit change. LGTM! Thank you so much for submitting this!

cc: @msaroufim

msaroufim:
Thank you for the speed @sgrigory! Let me know if you want more stuff to do!

msaroufim (approved):

ejguan:
@sgrigory Ops, I thought this PR came from OSS. Feel free to take over the imported Diff and land it. I don't want to steal credit from your valuable work! Thanks!

sgrigory:
> This is great! Thank you. And, do you mind opening an issue about your minor note? I think it's a bug.
> 
> And, adding original author @NivekT of this tutorial to review.

Thanks for the review - opened an issue #840

PR #846, opened 2022-10-20
[DataLoader2] Ensure finalize and finalize_iteration are called during shutdown/exception

### LINKED ISSUE #821: `DataLoader2.shutdown` should guarantee `finalize_iteration` is called for `ReadingService`
### 🐛 Describe the bug

Per title:
https://github.com/pytorch/data/blob/c42587a828d05f24f6f0586d17d3e9d55e1433ed/torchdata/dataloader2/dataloader2.py#L176-L183

Even though the `ReadingService` should guarantee the `finalize_iteration` called in `finalize`, it's better to guard this behavior in `DataLoader2` as well.

### Versions

main branch

NivekT:
* __->__ #846

Fixes #821

1. Ensure `finalize_iteration` is called during `DataLoader2.shutdown()`
2. If an uncaught exception is raised by `DataLoader2Iterator.next()`, ensure finalization calls are made

Differential Revision: [D40569566](https://our.internmc.facebook.com/intern/diff/D40569566)

### DIFF
--- a/torchdata/dataloader2/dataloader2.py
+++ b/torchdata/dataloader2/dataloader2.py
@@ -68,9 +68,8 @@ class DataLoader2Iterator(Iterator[T_co]):
                     self.dataloader.reading_service.finalize_iteration()
                 raise
             except Exception:
-                if self.dataloader.reading_service is not None:
-                    self.dataloader.reading_service.finalize_iteration()
-                    self.dataloader.reading_service.finalize()
+                if self.dataloader:
+                    self.dataloader.shutdown()
                 raise
         else:
             if self.dataloader.reading_service is not None:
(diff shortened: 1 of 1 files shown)


torchdata/dataloader2/dataloader2.py
```diff
@@ -67,6 +67,11 @@ def __next__(self) -> T_co:
                 if self.dataloader.reading_service is not None:
                     self.dataloader.reading_service.finalize_iteration()
                 raise
+            except Exception:
```
NivekT, line 70: 1. Is this potentially too broad? Should I limit this to specific exceptions? I wonder if users would want to intentionally catch exceptions that are raised here (such that finalization calls aren't needed).
2. We can also call `DataLoader2.shutdown()` here instead.

Thoughts?

VitalyFedyunin (approved):

torchdata/dataloader2/dataloader2.py
ejguan: I think calling dl2.shutdown should be good

NivekT:
@nivekt has imported this pull request.  If you are a Meta employee, you can view this diff [on Phabricator](https://www.internalfb.com/diff/D40569566).

Issue #802, opened 2022-10-03
Add support for `kwargs` in S3 DataPipes


NivekT:
### 🚀 The feature

`S3FileLister` and `S3FileLoader` currently doesn't support keyword argument beyond `request_timeout_ms`, `region`, `buffer_size`, and `multi_part_download`. Th

One example is [here](https://discuss.pytorch.org/t/bucket-versioning-using-s3fileloader/162699/2), where a user would like to read a specify version of the bucket. I imagine there are similar parameters that users may want to pass through. Some of which may be passed to the construction of the `S3Handler` and some others may be used in `handler.s3_read()`.

### Motivation, pitch

This will allow users to specify additional parameters to interact with S3 according to their needs.

### Alternatives

Add support in `fsspec` DataPipes instead and asks users to use those if necessary.

ejguan:
And, with your PR https://github.com/pytorch/data/pull/804, it should become doable for users.

kiukchung:
> I would say it's low pri until the perf issue has been solved for S3. And, in terms `fsspec`, it seems it's viable https://s3fs.readthedocs.io/en/latest/#bucket-version-awareness

Curious as to what the s3 perf issue is?

Issue #500, opened 2022-06-07
Performance Comparison between native AWSSDK and FSSpec


ejguan:
### 🐛 Describe the bug

After `AWSSDK` is integrated with TorchData, we now have two categories of `DataPipe`s to access and load data from AWS S3 Bucket:
1. [`DataPipe`](https://github.com/pytorch/data/blob/main/torchdata/datapipes/iter/load/fsspec.py) using `fsspec`: It relies on `s3fs` module to list/load data from S3 bucket. 
2. [`DataPipe`](https://github.com/pytorch/data/blob/main/torchdata/datapipes/iter/load/s3io.py) using `AWSSDK`: It relies on pybind from [`AWSSDK_CPP`](https://github.com/aws/aws-sdk-cpp) module.

And, I want to carry out a performance comparison of `Lister` and `Opener`/`Loader` between these two ways.
- For `Lister`s, I was using the same root path of `"s3://ai2-public-datasets/charades"` and validated that they returned the same values during iteration.
<details>
<summary>Testing script</summary>

```py
import numpy as np
import timeit

s3_path = "s3://ai2-public-datasets/charades"

def s3_fl_time():
    SETUP_CODE = """
from torchdata.datapipes.iter import IterableWrapper, S3FileLister
from __main__ import s3_path
dp = S3FileLister(IterableWrapper([s3_path]), region="us-west-2")
"""
    TEST_CODE = """
_ = list(dp)
"""
    times = timeit.repeat(setup = SETUP_CODE, stmt = TEST_CODE, repeat=10, number = 5)
    print(f"S3FileLister: Mean({np.average(times)}), STD({np.std(times)})")

def fsspec_fl_time():
    SETUP_CODE = """
from torchdata.datapipes.iter import IterableWrapper, FSSpecFileLister
from __main__ import s3_path
dp = FSSpecFileLister(IterableWrapper([s3_path]), anon=True)
"""
    TEST_CODE = """
_ = list(dp)
"""
    times = timeit.repeat(setup = SETUP_CODE, stmt = TEST_CODE, repeat=10, number = 5)
    print(f"FSSpecFileLister: Mean({np.average(times)}), STD({np.std(times)})")

if __name__ == "__main__":
    s3_fl_time()
    fsspec_fl_time()
```

</details>

And the result is:
```
S3FileLister: Mean(1.7595681754999994), STD(0.20364943594288445)
FSSpecFileLister: Mean(0.19180457339999962), STD(0.5630912985701465)
```
[...]

PR #847, opened 2022-10-20
[Benchmark] Adding minimal, reproducible AWS S3 benchmark

NivekT:
* __->__ #847


ejguan (approved):
LGTM. I think it's good!

kiukchung:
Bringing this back to life from an offline conversation I had with @msaroufim.

It would be interesting to see run this benchmark comparing fsspec-s3 vs the s3-plugin (the one that uses the pybinded s3 crt client).

My team at Amazon uses the s3-plugin (through torchdata) to stream-load json (text) training data from S3. We have millions of shards (file in s3) and each shard is between 1 ~ 10GB (single digit gigabytes) big. Using torchdata's builtin s3 datapipe (e.g. `load_files_from_s3()`) I'm getting a throughput of 1GiB/sec out of the box from a single dataloader worker. 

While the example benchmark here is great, it falls a bit short when it comes to real-life scenarios:

1. For serious use-cases, the size of the files in S3 are in the orders of single-digit GB rather than MB
2. For non-image (binary) examples, no prefetching is required since one can stream-read from S3 into StringIOBuffer. This is preferable to prefetching large S3 files which has drawbacks like needing to provision instances with larger instance volumes, and making sure to clean up the prefetches when the job finishes/fails.

While the fsspec-s3 reader (which relies on aioboto3 for asyncio reads) is great for composing a batch from multiple smaller shards, the s3 c++ pybind client would be much more performant when one cannot take advantage of asyncio parallelism when loading few large shards.

It would be a shame to ditch support for the s3-plugin in favor of a more pythonic fsspec-s3 as hinted in these issues:

1. https://github.com/pytorch/data/issues/871
2. https://github.com/pytorch/data/issues/802

I understand the portability/maintainability concern of shipping c-bindings with python packages. Happy to discuss solutions/collaborations from our end to ensure that torchdata users have options for both s3 datapipes.

ejguan:
Thanks for providing this valuable feedback. I appreciate that you have shown us different use cases loading from S3. We previously rely on Vision examples to run the benchmarking. 

> I understand the portability/maintainability concern of shipping c-bindings with python packages

This is generally not a problem as we have already shipped this packages with S3 plugin enabled. The main reason we currently are pivoting users to use `fsspec`-based DataPipe is a concern raised previously. See: https://github.com/pytorch/data/issues/500

Without clear understanding on what causes S3-plugin having worse performance in either small files or big archives, it's hard for us to recommend users to rely on S3-plugin. And, we are kind short-of-hand on investigating the root cause or best practice on loading from S3. (I have to admit that the dataset with large archive might still be smaller than GB-scale.)

NivekT:
Hi @kiukchung,

Thanks for looking into this!

> It would be interesting to see run this benchmark comparing fsspec-s3 vs the s3-plugin (the one that uses the pybinded s3 crt client).

In our Slack channel "aws_s3_plugin", I attached a file showing the benchmark results comparing in two. The setup is very similar to the one in this PR. Let us know your thoughts on that.

I agree that the two points that you mentioned would be interesting to look at, but as @ejguan mentioned, we are a bit short staffed at the moment. If you have publicly reproducible results/investigations, we can definitely look at it.

kiukchung:
Thanks @NivekT  and @ejguan. I spent the day running some benchmarks between s3io vs fsspec on our datasets in s3. Will comment on https://github.com/pytorch/data/issues/500.

PR #849, opened 2022-10-21
Fix is_local for paths starting with `az://`

### LINKED ISSUE #840: File paths get duplicated by `list_files_by_fsspec` pipeline if folder path starts with `az://`
## Context
For reading files from Azure Blob storage Gen2, `fsspec` [allows both path prefixes `abfs://` or `az://`](https://github.com/fsspec/adlfs/blob/f15c37a43afd87a04f01b61cd90294dd57181e1d/README.md?plain=1#L33) as synonyms. Paths starting `abfs://` work fine for us, but paths starting with `az://` result in duplicated output when passed to `list_files_by_fsspec` pipeline. 

This first showed up in #836  

## Example
```python

dp = IterableWrapper(["az://public/curated/covid-19/"]).list_files_by_fsspec(account_name="pandemicdatalake")
list(dp)[0]
```

Output:
```
'public/curated/covid-19/public/curated/covid-19/bing_covid-19_data'
```
instead of the correct one 
```
'public/curated/covid-19/bing_covid-19_data'

```

## Possible reason
This probably has to do with how `FSSpecFileListerIterDataPipe.__iter__` decides if the path is local:
- If a path starts with `az`, the variable `fs.protocol` [here](https://github.com/pytorch/data/blob/768ecdae8b56af640a78e29f82864dc4f65df371/torchdata/datapipes/iter/load/fsspec.py#L82) is still `abfs`
- So the condition `root.startswith(protocol)` is false, and `is_local` is true
- As a result the path "doubles" in [this line](https://github.com/pytorch/data/blob/768ecdae8b56af640a78e29f82864dc4f65df371/torchdata/datapipes/iter/load/fsspec.py#L95)

Perhaps we'd need to find a different way of checking if the path is local, not relying on matching the beginning of the path with `fs.protocol`

### Versions

Collecting environment information...
PyTorch version: 1.14.0.dev20221018
Is debug build: False

OS: macOS 12.6 (arm64)
GCC version: Could not collect
Clang version: 14.0.0 (clang-1400.0.29.102)
CMake version: version 3.22.1
Libc version: N/A

Python version: 3.9.13 (main, Oct 13 2022, 16:12:19)  [Clang 12.0.0 ] (64-bit runtime)
Python platform: macOS-12.6-arm64-arm-64bit
Is CUDA available: False
CUDA runtime version: No CUDA
CUDA_MODULE_LOADING set to: N/A
cuDNN version: No CUDA
HIP runtime version: N/A
MIOpen runtime version: N/A
Is XNNPACK available: True

[pip3] mypy==0.982
[pip3] mypy-extensions==0.4.3
[pip3] torch==1.14.0.dev20221018
[conda] pytorch                   1.14.0.dev20221018         py3.9_0    pytorch-nightly

mpfirrmann:
Signed-off-by: Marco Pfirrmann <marco.pfirrmann@tngtech.com>

Fixes #840. 
The module used to extract the protocol does not differentiate between `abfs://` and `az://` paths, always returning `abfs`. This fix adds `az` to the protocol list to be checked.

### DIFF
--- a/test/requirements.txt
+++ b/test/requirements.txt
@@ -12,3 +12,4 @@ rarfile
 protobuf >= 3.9.2, < 3.20
 datasets
 graphviz
+adlfs
--- a/test/test_remote_io.py
+++ b/test/test_remote_io.py
@@ -213,6 +213,21 @@ class TestDataPipeRemoteIO(expecttest.TestCase):
         res = list(fsspec_loader_dp)
         self.assertEqual(len(res), 18, f"{input} failed")
 
+    @skipIfNoFSSpecS3
+    def test_fsspec_azure_blob(self):
+        url = "public/curated/covid-19/ecdc_cases/latest/ecdc_cases.csv"
+        account_name = "pandemicdatalake"
+        azure_prefixes = ["abfs", "az"]
+        fsspec_loader_dp = {}
+
+        for prefix in azure_prefixes:
+            fsspec_lister_dp = FSSpecFileLister(f"{prefix}://{url}", account_name=account_name)
+            fsspec_loader_dp[prefix] = FSSpecFileOpener(fsspec_lister_dp, account_name=account_name).parse_csv()
+
+        res_abfs = list(fsspec_loader_dp["abfs"])[0]
+        res_az = list(fsspec_loader_dp["az"])[0]
+        self.assertEqual(res_abfs, res_az, f"{input} failed")
+
     @skipIfAWS
     def test_disabled_s3_io_iterdatapipe(self):
         file_urls = ["s3://ai2-public-datasets"]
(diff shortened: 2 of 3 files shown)


ejguan (commented):
Thank you. Changes look good. Have you verified it working properly with the script in the issue. Could you please add a test [here](https://github.com/pytorch/data/blob/main/test/test_remote_io.py)?

NivekT (commented):
Thanks for opening this PR. I agree testing this behavior will be helpful.

NivekT (approved):
The test LGTM!

NivekT (changes requested):
The test LGTM but we need to either install `adlfs` or skip it when `adlfs` is not available. The former is probably best.

mpfirrmann:
> The test LGTM but we need to either install `adlfs` or skip it when `adlfs` is not available. The former is probably best.

I added the `adlfs` package to `test/requirements.txt` and it should be installed by the testing workflows.

PR #850, opened 2022-10-21
RFC for using sphinx doctest

### LINKED ISSUE #848: [RFC] Verify that the docs contain working code and self-contained examples using doctest
### 🚀 The feature

Currently there does not seem to be an automatic way to verify that the examples in the documentation are actually working. This leads to issues like (https://github.com/pytorch/data/issues/433).

An example should also be complete enough so that developers can easily try out the code.

A solution could be to use the sphinx doctest extension to test the documentation before building it. Docstrings can be continuously migrated from standard reST doctests to test code that runs using the sphinx doctest extension.

### Motivation, pitch

Working examples that are up-to-date boost adoption of the library and make it easier for developers to become proficient in using the library.

Therefore one could consider using doctest in order to be forced to write self-contained examples that execute without error.

### Alternatives

Doctests can be executed in different ways:

-   Invoking plain python to execute the doctests as described [here](https://docs.python.org/3/library/doctest.html)
-   Using [pytest --doctest](https://docs.pytest.org/en/7.1.x/how-to/doctest.html) to execute the tests
-   Run within the documentation build process as `cd docs && make doctest`

I would recommend running the doctests while building the documentation using sphinx because it is easy to continuously
migrate the existing non-tested example code to code being tested.

### Additional context

A minimal example of the RFC can be found
[here](https://github.com/pytorch/data/pull/850). Please
note that the code is only meant as an example for discussion and might not (yet) meet the quality criteria of a PR.

The example implementation consists of the following parts:

-   An updated `docs/Makefile` with a `doctest` target
-   Enabling the sphinx extension `sphinx.ext.doctest` in `docs/source/conf.py`
-   A minimal example of an updated docstring in `torchdata/dataloader2/adapter.py`
-   Adding the `doctest` step to the CI in `.github/workflows/_build_test_upload.yml`

The tests can be executed like this: `cd docs && make doctest`.

mabu-github:
RFC for using sphinx doctest and continously migrating the doctest code to being tested

Fixes #848

### Changes

-   Updated `docs/Makefile` with a `doctest` target
-   Enabled the sphinx extension `sphinx.ext.doctest` in `docs/source/conf.py`
-   A minimal example of an updated docstring in `torchdata/dataloader2/adapter.py`
-   Adding the `doctest` step to the CI in `.github/workflows/_build_test_upload.yml`

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -77,6 +77,12 @@ jobs:
           BUILD_S3: 1
       - name: Install test requirements
         run: pip3 install -r test/requirements.txt
+      - name: Test documentation examples
+        run: |
+          cd ./docs
+          pip3 install -r requirements.txt
+          make doctest
+          cd ..
       - name: Run DataPipes tests with pytest
         if: ${{ ! contains(github.event.pull_request.labels.*.name, 'ciflow/slow') }}
         run:
--- a/torchdata/dataloader2/adapter.py
+++ b/torchdata/dataloader2/adapter.py
@@ -53,9 +53,21 @@ class Shuffle(Adapter):
             - None: No-op. Introduced for backward compatibility.
 
     Example:
-        >>>  dp = IterableWrapper(range(size)).shuffle()
-        >>>  dl = DataLoader2(dp, [Shuffle(False)])
-        >>>  self.assertEqual(list(range(size)), list(dl))
+
+    .. testsetup::
+
+        from torchdata.datapipes.iter import IterableWrapper
+        from torchdata.dataloader2 import DataLoader2
+        from torchdata.dataloader2.adapter import Shuffle
+
+        size = 12
+
+    .. testcode::
+
+        dp = IterableWrapper(range(size)).shuffle()
+        dl = DataLoader2(dp, [Shuffle(False)])
+        assert list(range(size)) == list(dl)
+
     """
 
     def __init__(self, enable=True):
--- a/docs/Makefile
+++ b/docs/Makefile
@@ -12,7 +12,12 @@ BUILDDIR      = build
 help:
 	@$(SPHINXBUILD) -M help "$(SOURCEDIR)" "$(BUILDDIR)" $(SPHINXOPTS) $(O)
 
-.PHONY: help Makefile
+doctest: html
+	$(SPHINXBUILD) -b doctest $(SPHINXOPTS) "$(SOURCEDIR)" "$(BUILDDIR)"/doctest
+	@echo "Testing of doctests in the sources finished, look at the " \
+	      "results in $(BUILDDIR)/doctest/output.txt."
+
+.PHONY: help doctest Makefile
 
 # Catch-all target: route all unknown targets to Sphinx using the new
 # "make mode" option.  $(O) is meant as a shortcut for $(SPHINXOPTS).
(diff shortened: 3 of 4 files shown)


torchdata/dataloader2/adapter.py
```diff
@@ -53,9 +53,21 @@ class Shuffle(Adapter):
...
+        from torchdata.dataloader2.adapter import Shuffle
+
+        size = 12
+
+    .. testcode::
+
+        dp = IterableWrapper(range(size)).shuffle()
+        dl = DataLoader2(dp, [Shuffle(False)])
+        assert list(range(size)) == list(dl)
+
```
ejguan, line 70: Could you please provide a screen shot of this change on our online doc?
It was like https://pytorch.org/data/beta/generated/torchdata.dataloader2.adapter.Shuffle.html#torchdata.dataloader2.adapter.Shuffle

.github/workflows/_build_test_upload.yml
```diff
@@ -504,6 +504,7 @@ jobs:
         run: |
           cd ./docs
           pip3 install -r requirements.txt
+          make doctest
```
ejguan, line 507: I would rather put this test into `ci.yml`. This `_build_test_upload` is the one we upload compiled binaries.

All PR will execute GHA in `ci.yml`.

torchdata/dataloader2/adapter.py
ejguan: This is how to build doc: https://github.com/pytorch/data/tree/main/docs#building-the-documentation

ejguan: And here is the way to preview the change locally https://github.com/pytorch/pytorch/blob/master/CONTRIBUTING.md#previewing-changes-locally

mabu-github: Here is a side by side comparison of the docs before (left) and after the changes (right). Nothing has changed except that it is now executable with doctest as there is some setup code (sphinx hides it from the docs) for the imports and the size parameter.

shuffle_docs

.github/workflows/_build_test_upload.yml
mabu-github: I have put it into ci.yml now.

torchdata/dataloader2/adapter.py
ejguan: That's great! Thank you!

ejguan (commented):
I think this PR looks good. Thank you! Since we have enabled the ci for doctest, we might need to fix all docstring in this PR.

.github/workflows/ci.yml
```diff
@@ -77,6 +77,12 @@ jobs:
           BUILD_S3: 1
       - name: Install test requirements
         run: pip3 install -r test/requirements.txt
+      - name: Test documentation examples
+        run: |
+          cd ./docs
+          pip3 install -r requirements.txt
+          make doctest
```
ejguan, line 84: Looks like the test is not carried out.
image

mabu-github: Doctest seems to depend on the docs having been built previously. So I added a dependency on the html target into the makefile.

mabu-github:
> I think this PR looks good. Thank you! Since we have enabled the ci for doctest, we might need to fix all docstring in this PR.

If you say that this kind of setup is the way to move forward, I am happy to committing to do further migration of the docstrings. Maybe we can go package by package to keep the PRs manageable?

The approach allows continuous migration and as a finalization we could then remove `doctest_test_doctest_blocks` config setting in the final PR to make sure that no untested doctest blocks starting with `>>> ...` can get into new code.

ejguan:
> Maybe we can go package by package to keep the PRs manageable?

Agree. Thank you!

ejguan (approved):
Thank you for this effort. LGTM!

NivekT (approved):

PR #861, opened 2022-10-25
Deprecate set-output based on github instruction

ejguan:
Follow https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/

### DIFF
--- a/.github/workflows/domain_ci.yml
+++ b/.github/workflows/domain_ci.yml
@@ -46,7 +46,7 @@ jobs:
 
       - name: Extract torchvision ref
         id: torchvision
-        run: echo "::set-output name=ref::$(python -c 'import torchvision; print(torchvision.version.git_version)')"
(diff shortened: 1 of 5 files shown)


.github/workflows/domain_ci.yml
```diff
@@ -46,7 +46,8 @@ jobs:
 
       - name: Extract torchvision ref
         id: torchvision
-        run: echo "::set-output name=ref::$(python -c 'import torchvision; print(torchvision.version.git_version)')"
+        run:
+          echo "name=ref::$(python -c 'import torchvision; print(torchvision.version.git_version)')" >> $GITHUB_OUTPUT
```
pmeier, line 50:
```suggestion
          echo "ref=$(python -c 'import torchvision; print(torchvision.version.git_version)')" >> $GITHUB_OUTPUT
```

ejguan: Thanks! Somehow I didn't clean this up thoroughly.

NivekT (approved):

PR #866, opened 2022-10-28
Support for recursive datatypes

abhi-glitchhg:
since release of mypy 0.981 recursive types are supported; i have just removed the `#` as per suggestion in the TODO comment and have changed the mypy version in the ci jobs. 

This PR is to check if ci is breaking with the changes are not.

Let me know if any further changes need to be made
thanks..

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -53,7 +53,7 @@ jobs:
           pip3 install --pre torch -f "${{ steps.pytorch_channel.outputs.value }}"
       - name: Install dependencies
         run: |
-          pip3 install requests mypy==0.982 graphviz numpy
+          pip3 install requests mypy==0.981 graphviz numpy
       - name: Build TorchData
         run: |
           python setup.py develop
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -2,6 +2,10 @@ name: Lint
 
 on:
   push:
+    branches:
+      - main
+      - release/*
+      -abhi-glitchhg
   pull_request:
 
 jobs:
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -2,9 +2,6 @@ name: Lint
 
 on:
   push:
-    branches:
-      - main
-      - release/*
   pull_request:
 
 jobs:
@@ -53,7 +50,7 @@ jobs:
           pip3 install --pre torch -f "${{ steps.pytorch_channel.outputs.value }}"
       - name: Install dependencies
         run: |
-          pip3 install requests mypy==0.812 graphviz numpy
+          pip3 install requests mypy==0.982 graphviz numpy
       - name: Build TorchData
         run: |
           python setup.py develop
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -5,7 +5,6 @@ on:
     branches:
       - main
       - release/*
-      -abhi-glitchhg
   pull_request:
 
 jobs:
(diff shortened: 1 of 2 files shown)


NivekT (commented):
@ejguan  Any issue form upgrading `mypy` to a newer version? Aside from fixing the lints?

abhi-glitchhg:
Hmm, mypy is raising one error on this line 

```
returning TypeVar should receive at least one argument containing the same
Typevar  [type-var]
Found 1 error in 1 file (checked 76 source files)
        def draw(self) -> T:
```

https://github.com/pytorch/data/blob/b1b3406389fa6665ec3422ae5e1ebf65ae7e2938/torchdata/datapipes/iter/util/randomsplitter.py#L103-L105 


How can i fix this? can anyone help me out here?

NivekT:
Based on this [issue](https://github.com/python/mypy/issues/13765), it seems to be warning that a single TypeVar will not do anything. I think you can either remove `-> T` or suppress the warning.

ejguan (changes requested):
Thank you for your contribution. The PR looks good. However, I am sorry that I have to block this PR for now until PyTorch Core has updated their pinned [mypy version](https://github.com/pytorch/pytorch/blob/6bb7f4f29f0a36ce410ce53d824f531eaf74c76e/.circleci/docker/requirements-ci.txt#L83) from `0.960` to `0.981`.

The main problem is that torchdata is still partially in the tree of PyTorch. And PyTorch has strong opinion on the version of `mypy`. See: https://github.com/pytorch/pytorch/blob/6bb7f4f29f0a36ce410ce53d824f531eaf74c76e/mypy.ini#L5
If we make this PR landed, in order to develop torchdata without `mypy` complain on my local environment, I have to update `mypy` to 0.981. However, if I need to work on something in PyTorch Core, the updated `mypy` will prevent me running static type checking on PyTorch.

IMHO, to land this PR, we accomplish one of the following 2 options:
- Make TorchData fully out of PyTorch
- Update PyTorch's `mypy` requirement

LMK if it makes sense. And, please correct me if anything wrong or we have other options.

abhi-glitchhg:
hello @ejguan! Thanks for the feedback.

> Make TorchData fully out of PyTorch

Honestly, I do not know the consequences and the amount of work required for this but the second option seems easier than the first. 

Anyway, there is no hurry from my side; I just noticed the TODO issue and was aware of the recent mypy version release. 

We could try updating the mypy version. Are there any existing blockers in doing this that you are aware of?

Would love to hear your thoughts. 
Thanks!

.github/workflows/lint.yml
```diff
@@ -64,7 +64,7 @@ jobs:
         run: |
           set -eux
           STATUS=
-          if ! mypy --config=mypy.ini; then
+          if ! mypy --config=mypy.ini --enable-recursive-aliases --install-types --non-interactive; then
```
abhi-glitchhg, line 67: I would also like to point out that, `--install-types --non-interactive` will install the missing package stubs hence [slower](https://mypy.readthedocs.io/en/stable/running_mypy.html#library-stubs-not-installed). 

Quoting relevant part from the above documentation link :- `This is somewhat slower than explicitly installing stubs before running mypy, since it may type check your code twice – the first time to find the missing stubs, and the second time to type check your code properly after mypy has installed the stubs.`

 I don't think downloading the package stubs would be a bottleneck; but if it is, then we could think of using cached installations. (just a suggestion)

ejguan:
BTW, regarding the recursive type hints, `0.981` is not the best version as it's an experimental feature. `0.990` is the version that officially supports recursive type hints.

ejguan:
Let's wait until PyTorch has officially updated mypy

abhi-glitchhg:
> 

I was just going to mention this.

.github/workflows/lint.yml
```diff
@@ -53,7 +53,7 @@ jobs:
           pip3 install --pre torch -f "${{ steps.pytorch_channel.outputs.value }}"
       - name: Install dependencies
         run: |
-          pip3 install requests mypy==0.812 graphviz numpy
+          pip3 install requests mypy==0.981 graphviz numpy
```
abhi-glitchhg, line 56:
```suggestion
          pip3 install requests mypy==0.990 graphviz numpy
```

PR #870, opened 2022-11-01
[DLv2] Forward fix adapters applied to None DataPipe

ejguan:
Per title. This is a forward fix of https://github.com/pytorch/data/pull/867


NivekT (approved):
LGTM!

PR #879, opened 2022-11-03
[PrototypeRS] Adding 'pause' and 'resume' operations to halt DataPipes

NivekT:
* #915
* #998
* #1011
* __->__ #879

The goal of this PR is fully stop DataPipe activities in preparation of snapshotting (which requires a halted state), so buffers will not be changing while we record the snapshot.

This PR doesn't handle pausing or resuming with `sharding_round_robin_dispatch`. This will be fixed in an upcoming PR and is tracked by #991.

---
Implementation note (feel free to discuss):
`_pause` exists for both `DataLoader2Iterator` and `DataLoader2` as private methods.
`resume` exists for both `DataLoader2Iterator` as public but it is private for `DataLoader2`.

`limit` and `clear_limit` exist only for `DataLoader2Iterator` as a public methods.
- Limit persists through iterations (with the iterator) until it is manually cleared via `clear_limit`.
- Creating a new iterator (`iter(dl2)`) will not retain the previous limit, because the limit is attached to the iterator.

Differential Revision: [D41744759](https://our.internmc.facebook.com/intern/diff/D41744759)


NivekT:
Another thing to discuss: should we do double underscore or something to protect `__reset`` and `__full_stop` from users?

torchdata/datapipes/iter/util/prefetcher.py
```diff
@@ -123,3 +123,18 @@ def reset(self):
...
+            self.prefetch_data.run_prefetcher = False
+            self.thread.join()
+            self.thread = None
+
+    def resume(self):
+        self.thread = threading.Thread(
+            target=PrefetcherIterDataPipe.thread_worker, args=(self.prefetch_data,), daemon=True
+        )
+        self.prefetch_data.run_prefetcher = True
+        self.thread.start()
```
NivekT, line 140: @ejguan What will be the proper ways to stop and resume this DataPipe? The goal here is to halt any thread that can change a buffer or the state of the DataPipe graph.

Will I have to set process group to `None` or change `_cv` as well?

NivekT (commented):
Updated the implementation but I still need some test cases. Suggestions are welcomed.

torchdata/datapipes/iter/util/prefetcher.py
```diff
@@ -123,3 +123,18 @@ def reset(self):
...
             self.prefetch_data.run_prefetcher = False
             self.thread.join()
+            self.thread = None
+
+    def full_stop(self):
+        if self.thread is not None:
+            # Note: the content of the buffer still exists in `prefetch_data.prefetch_buffer`
+            self.prefetch_data.run_prefetcher = False
+            self.thread.join()
+            self.thread = None
```
ejguan, line 133: It should be fine as long as `buffer` is filled before thread exits. You might want to validate the result when we do pause at the end of epoch and resume.

torchdata/datapipes/iter/util/prefetcher.py
ejguan: No need to change pg or cv.

ejguan (commented):
Can we change the name to `pause`? `full stop` seems like an operation ran at the end epoch

ejguan:
I forget to mention about one scenario. If I iterate datapipe after `pause` but before `resume`, I expect a `StopIteration` is raised. Am I right?
```py
>>> dp.pause()
>>> for d in dp:
...     print(d)  # Nothing should be printed
>>> dp.resume()
>>> for d in dp:
...     print(d)
# Values
```

NivekT:
> I forget to mention about one scenario. If I iterate datapipe after `pause` but before `resume`, I expect a `StopIteration` is raised. Am I right?
> 
> ```python
> >>> dp.pause()
> >>> for d in dp:
> ...     print(d)  # Nothing should be printed
> >>> dp.resume()
> >>> for d in dp:
> ...     print(d)
> # Values
> ```

If you try to call `next` after `pause`, this implementation raises `RuntimeError`.

ejguan:
> If you try to call `next` after `pause`, this implementation raises `RuntimeError`.

If we want to achieve mini-epoch for OSS as well, we might need it to be `StopIteration`.

NivekT:
I see. I think what we can do is:

Inside the internal of PrototypeRS, if it cannot submit a `GetNextRequest` after `pause` without calling `resume` first. This is implemented in this PR.

Publicly, DataLoader2 can behave different (maybe like you said return `StopIteration`), but it needs to do so without triggering reset when a new iterator is made? That might get a bit tricky and I need to think more about it. This part is not covered by this PR (since I'm not exposing public API of `pause` or `resume` here).

torchdata/dataloader2/communication/iter.py
```diff
@@ -143,12 +144,38 @@ def DataPipeBehindQueues(source_datapipe, protocol, blocking_request_get=False,
...
         elif isinstance(request, communication.messages.TerminateRequest):
             forever = False
             protocol.response_terminate()
 
         elif isinstance(request, communication.messages.GetNextRequest):
             while forever:
+                if protocol._pause:
+                    raise RuntimeError(
+                        "Cannot `GetNext` after `Pause` has been called. " "`Resume` must be called first."
+                    )
```
ejguan, line 178: I might think this problem over simplified. Could we do `protocol.response_stop_iteration()` here to achieve that?

torchdata/dataloader2/dataloader2.py
```diff
@@ -49,7 +49,10 @@ def __next__(self) -> T_co:
         if self.iterator_id == self.dataloader.valid_iterator_id:
             self.dataloader._reset_iter = True
             try:
-                return next(self.dataloader._datapipe_iter)  # type: ignore[arg-type]
+                if self.dataloader._paused:
+                    raise StopIteration
+                else:
+                    return next(self.dataloader._datapipe_iter)  # type: ignore[arg-type]
```
NivekT, line 55: Thoughts on having DL2 directly check if it is paused or not (versus at the RS level)?

This means providing the (potentially) public API `pause`/`resume`. We can also make them private.

NivekT (commented):
@ejguan I think this is ready for review, some notable things to look at:

1. I am adding an API `pause` `resume` at the DL2 level? Is this fine? Will it conflict with internal use cases? We can also make these methods private if necessary (with no impact to functionality).
2. I added a timeout while joining the thread for `Prefetcher`. For test case with `rs7`, we have a prefetcher in the main loop (not with the subprocess) and the thread doesn't properly join. Aside from that, the behavior is correct. Let me know if you have thoughts on this.

torchdata/dataloader2/dataloader2.py
ejguan: It makes sense to have DL2 having those APIs (we can add them as prototype feature raising warning when calling them). And, it also means we need to think about how to add the corresponding APIs to `ReadingService` as well so that DL2 can all them accordingly.

test/dataloader2/test_proto_multi_rs.py
```diff
@@ -0,0 +1,105 @@
...
+                    if i in {2, self.n_elements - 2}:
+                        dl.pause()
+                        dl.resume()
+
+                self.assertEqual(list(range(self.n_elements)), sorted(res))
+
+    def test_reading_service_pause_stop_yield(self) -> None:
+
+        # Functional Test: Confirms that `dl` will stop yielding elements after `_pause` is called
+        rs7 = PrototypeMultiProcessingReadingService(num_workers=2, worker_prefetch_cnt=0, main_prefetch_cnt=1)
```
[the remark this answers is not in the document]
ejguan: But, `rs2` works properly. right? What's the difference between those two tests?

torchdata/datapipes/iter/util/prefetcher.py
```diff
@@ -59,25 +62,27 @@ def __init__(self, source_datapipe, buffer_size: int = 10):
...
-            else:  # Buffer is full, waiting for main thread to consume items
-                # TODO: Calculate sleep interval based on previous consumption speed
-                time.sleep(PRODUCER_SLEEP_INTERVAL)
+                else:  # Buffer is full, waiting for main thread to consume items
+                    # TODO: Calculate sleep interval based on previous consumption speed
+                    time.sleep(PRODUCER_SLEEP_INTERVAL)
+            time.sleep(PRODUCER_SLEEP_INTERVAL)
 
     def __iter__(self):
         if self.buffer_size < 1:
```
ejguan, line 88: For convenience, let's remove the case of `self.buffer_size < 1` from `prefetch`.

torchdata/datapipes/iter/util/prefetcher.py
```diff
@@ -90,16 +95,20 @@ def __iter__(self):
...
-                        time.sleep(CONSUMER_SLEEP_INTERVAL)
+                        if not prefetch_data.stop_iteration:
+                            time.sleep(CONSUMER_SLEEP_INTERVAL)
+                        else:
+                            prefetch_data.run_prefetcher = False
             finally:
                 prefetch_data.run_prefetcher = False
                 if self.thread is not None:
-                    self.thread.join()
+                    self.thread.join(3)  # TODO: Is it fine for us to have a timeout here?
```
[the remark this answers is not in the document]
ejguan: I feel this is not the right way to do. Can you pls check if this line has been reached or not?
This is a chance that the dataloader iterator is holding the reference of datapipe's iterator, which would prevent this `finally` clause being invoked.

test/dataloader2/test_proto_multi_rs.py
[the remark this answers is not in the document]
NivekT: `rs2` resume after, `rs7` doesn't

torchdata/datapipes/iter/util/prefetcher.py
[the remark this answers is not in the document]
NivekT: Yes, that line `self.thread.join()` is reached

torchdata/datapipes/iter/util/prefetcher.py
NivekT: `self.buffer_size < 1` is currently not allowed, there is a check in `__init__` for that purpose.

The lines of code here are doing something else.

torchdata/datapipes/iter/util/prefetcher.py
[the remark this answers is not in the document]
ejguan: When is it reached? At the end of the the test case (during garbage collecting) or at the end of loop?
If we do `pause` and never resume, I think it's better to call `dl2.shutdown()` to indicate that the iteration has been ended.

torchdata/datapipes/iter/util/prefetcher.py
ejguan: What do you mean by doing something else?
I think it's fine to remove if `self.buffer_size < 1` is currently not allowed. It's basically not prefetching, right?

NivekT: Oh I see which lines you are referring to. Removed.

test/dataloader2/test_proto_multi_rs.py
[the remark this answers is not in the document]
ejguan: It would be easier to identify the issue starting from running a single dataloader2 per test. After we make sure it works properly within single dataloader2, we can expand the test case to multiple dataloader2.
With multiple dataloader2, we need to find a way to properly finalize reading service when needed.

torchdata/datapipes/iter/util/prefetcher.py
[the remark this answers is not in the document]
NivekT: It was reached at the end of test case. I think I identified the issue, `prefetch_data.stop_iteration` was not set to `True` previously, and that keeps the thread sleeping instead of exiting.

Setting that to `True` seems to do the trick without needing a timeout

test/dataloader2/test_proto_multi_rs.py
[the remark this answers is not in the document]
NivekT: I do test one DataLoader2 at a time when testing locally. Let me know if I should separate them into one DataLoader2 per `def` as well

NivekT:
@nivekt has imported this pull request.  If you are a Meta employee, you can view this diff [on Phabricator](https://www.internalfb.com/diff/D41744759).

torchdata/dataloader2/dataloader2.py
ejguan: Yeah. I am thinking to keep a noop `pause` and `resume` (with warning) to base `ReadingServiceInterface` should be good enough.

test/dataloader2/test_proto_multi_rs.py
```diff
@@ -0,0 +1,92 @@
...
+            #                  pauses and resumes
+            for rs in test_rss:
+                dl: DataLoader2 = DataLoader2(dp, reading_service=rs)
+                res = []
+                for i, x in enumerate(dl):
+                    res.append(x)
+                    if i in {2, self.n_elements - 2}:
+                        dl.pause()
+                        dl.resume()
+
```
ejguan, line 63: Let's do `dl.shutdown()` at the end of each `DataLoader2` to show the preferred way finalizing dataloader2

test/dataloader2/test_proto_multi_rs.py
```diff
@@ -0,0 +1,92 @@
...
+        test_rss2 = [rs7, rs8, rs9]
+        for rs in test_rss2:
+            dl: DataLoader2 = DataLoader2(self.double_pause_dp, reading_service=rs)
+            res = []
+            for i, x in enumerate(dl):
+                res.append(x)
+                if i in {2}:
+                    dl.pause()
+            self.assertEqual(3, len(res))
```
ejguan, line 81: Ditto

test/dataloader2/test_proto_multi_rs.py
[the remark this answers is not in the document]
ejguan: No need as long as the test is working properly.

test/dataloader2/test_proto_multi_rs.py
ejguan, line 81: BTW, could you please also provide `msg` to self.assertEqual, so we will be able to know which `ReadingService` is broken.

NivekT:
Updated the PR based on the comments. I am expecting all tests to pass (except for unrelated domain test failures).

NivekT:
@nivekt has imported this pull request.  If you are a Meta employee, you can view this diff [on Phabricator](https://www.internalfb.com/diff/D41744759).

NivekT:
@nivekt has imported this pull request.  If you are a Meta employee, you can view this diff [on Phabricator](https://www.internalfb.com/diff/D41744759).

NivekT:
@nivekt has imported this pull request.  If you are a Meta employee, you can view this diff [on Phabricator](https://www.internalfb.com/diff/D41744759).

NivekT:
@nivekt has imported this pull request.  If you are a Meta employee, you can view this diff [on Phabricator](https://www.internalfb.com/diff/D41744759).

NivekT:
@nivekt has imported this pull request.  If you are a Meta employee, you can view this diff [on Phabricator](https://www.internalfb.com/diff/D41744759).

NivekT:
Seems like the tests are not terminating, will need to debug this tomorrow.

torchdata/dataloader2/dataloader2.py
```diff
@@ -49,7 +51,10 @@ def __next__(self) -> T_co:
...
             except PauseIteration:
```
Miiira, line 58: Feeling we can consolidate dataloader._pause and PauseIteration here. DL Iter PauseIteration, RS pause, DL pause should be consistent

ejguan (commented):
The tests introduced by this PR is not hanging. It's the test `test_dataloader2_batch_collate` stuck during the tests. I haven't gotten a chance to find root cause though.
During my debug session, I did find a few trivial bugs within the ProtoMPRS. Will send a PR to fix them later.

test/dataloader2/test_proto_multi_rs.py
```diff
@@ -0,0 +1,111 @@
...
+
+import unittest
+from unittest import TestCase
+
+from torchdata.dataloader2 import DataLoader2, PrototypeMultiProcessingReadingService
+
+from torchdata.datapipes.iter import IterableWrapper
+
+
+class TestPrototypeMultiProcessingReadingService(TestCase):
```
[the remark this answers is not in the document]
NivekT: I parametrized those and pass them to readingservice as arguments. All tests still pass locally.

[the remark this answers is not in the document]
ejguan: Did you run the test `test_dataloader2_batch_collate` in `test_dataloader2.py`?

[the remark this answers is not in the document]
NivekT: Thanks for pointing that out. It looks like changes to `Prefetcher` is causing that to hang. I will take a closer look tomorrow.

torchdata/dataloader2/dataloader2.py
```diff
@@ -257,3 +314,17 @@ def load_state_dict(self, state_dict: Dict[str, Any]) -> None:
...
         self._datapipe_before_reading_service_adapt = clone(self.datapipe)
+
+    def _pause(self):
+        self._is_paused = True
+        if hasattr(self.reading_service, "_pause"):
+            self.reading_service._pause()
+        else:
+            warnings.warn("ReadingService doesn't support pause.")
+
+    def resume(self):
```
NivekT, line 325: Should we make this private and only expose the API through `DataLoader2Iterator`?

ejguan: yeah, let's make it private for now then if you have any concern.

torchdata/dataloader2/dataloader2.py
```diff
@@ -281,3 +338,17 @@ def load_state_dict(self, state_dict: Dict[str, Any]) -> None:
...
         self._datapipe_before_reading_service_adapt = clone(self.datapipe)
+
+    def _pause(self):
+        self._is_paused = True
+        if hasattr(self.reading_service, "_pause"):
+            self.reading_service._pause()
+        else:
+            warnings.warn("ReadingService doesn't support pause.")
+
+    def _resume(self):
```
NivekT, line 349: Add a warning here when `_resume` is called when it is not paused, and then do no-op

test/dataloader2/test_proto_multi_rs.py
```diff
@@ -0,0 +1,192 @@
...
+                    f"num_workers = {rs.num_workers}, "
+                    f"worker_prefetch_cnt = {rs.worker_prefetch_cnt}, main_prefetch_cnt = {rs.main_prefetch_cnt}",
+                )
+
+                # Functional Test: Calling `next` after `limit` will trigger `StopIteration`
+                with self.assertRaisesRegex(StopIteration, "pause"):
+                    next(it)
+
+                # Functional Test: Verify that `limit` persists without the need to set it again
+                it.resume()
```
NivekT, line 157: Add a test example where `limit` is set to something else that is `None`, show that it resets the counter to 0.

test/dataloader2/test_proto_multi_rs.py
```diff
@@ -0,0 +1,220 @@
...
+                        dl._pause()
+                self.assertEqual(
+                    3,
+                    len(res),
+                    msg=f"The test is failing for rs{n + 7} with '{mp_method}', num_workers = {rs.num_workers}, "
+                    f"worker_prefetch_cnt = {rs.worker_prefetch_cnt}, main_prefetch_cnt = {rs.main_prefetch_cnt}",
+                )
+                dl.shutdown()
+
+    def test_reading_service_limit(self) -> None:
```
ejguan, line 125: Can you add another test for the case of dispatching process?
As, it's going to be used by DPP case.
You can do:
```py
dp = source_dp.shuffle().sharding_round_robin_dispatch(SHARDING_PRIORITIES.MULTIPROCESSING)
dp = dp.map(fn).batch(n)

rs = ProtoRS()
dl = DataLoader2(dp, reading_service=rs)
```

NivekT: I checked, I don't think it works with `.sharding_round_robin_dispatch` for `num_workers>1`, this is because my implementation tries to get a list of DataPipes that may need to be paused with `traverse_dps`:
`dp_list = list_dps(traverse_dps(source_datapipe))`

For a pipeline (passing through `ProtoRS` in DL2):
`dp1 = IterableWrapper().shuffle().sharding_filter().map()` or
`dp2 = IterableWrapper().shuffle().sharding_round_robin_dispatch().map()`
I get this:
`dp_list = [_IterDataPipeSerializationWrapper, Mapper, ShardingFilter, Shuffler, IterableWrapper]`

But for the latter, I get:
`dp_list = [_IterDataPipeSerializationWrapper, Mapper, _IterateQueueDP, QueueWrapper]`.

This leads to two problems:
1. `QueueWrapper` has `pause` that trigger `self.protocol.request_pause()`, so I think this leads to infinite recursive call of `pause`. I think this is fixable but I need to take a closer look.
2. The DataPipes prior to `sharding_round_robin_dispatch` are missing from `dp_list`, if there is a `.prefetch` before sharding, I don't think it will be paused.

ejguan: For the case of `dp2 = IterableWrapper().shuffle().sharding_round_robin_dispatch().map()`, the graph you captured was the graph in worker process. And the `_IterateQueueDP` and `QueueWrapper` are the connection DataPipe between worker process and dispatching process.
So, in your current flow, you need an extra step to do:
5. When there is `_IterateQueueDP` in the worker graph, it means there is a dispatching process. You need to call `pause` on it to pass the Pause request to the dispatching process.

NivekT:
@nivekt has imported this pull request.  If you are a Meta employee, you can view this diff [on Phabricator](https://www.internalfb.com/diff/D41744759).

NivekT:
@ejguan The hanging issue should be fixed and this should be good to go. I will enable these functionalities with round robin dispatch in a later PR.

test/dataloader2/test_dataloader2.py
```diff
@@ -235,6 +235,7 @@ def _collect_data(self, datapipe, reading_service_gen):
             result.append(row)
         for row in dl:
             result.append(row)
+        dl.shutdown()
```
NivekT, line 238: This doesn't fail if this is removed, but I think we should add it anyway.

ejguan: I am going to send a fix to it as it seems also breaking the tests for sequentialRS

NivekT: Makes sense to me. I don't think it is breaking this PR though? Or did I miss some CI failure

torchdata/dataloader2/reading_service.py
```diff
@@ -374,11 +374,48 @@ def clean_me(process, req_queue, res_queue):
...
+                "please use more than 0 worker."
+            )
+
+    def _resume(self):
+        """
+        Resumes DataPipes' activities. This is required to be called after `_pause` before
+        the DataLoader can keep yielding elements.
+        """
+        if self.main_prefetch_cnt > 0:
+            end_datapipe: DataPipe = self._end_datapipe.source_datapipe  # type: ignore[union-attr]
```
ejguan, line 401: Please use `self._main_prefetch_datapipe` rather than `self._end_datapipe.source_datapipe` to access main prefetcher, because there can be non-replicable DataPipes attached to the end of pipeline.

torchdata/dataloader2/dataloader2.py
```diff
@@ -62,7 +72,15 @@ def __next__(self) -> T_co:
                 if self.dataloader:
                     self.dataloader.shutdown()
                 raise
-        else:
+            finally:
```
ejguan, line 75: I feel there is no need to do `finally`. You can move those lines to the line before `return next_val`.

torchdata/dataloader2/communication/protocol.py
```diff
@@ -81,18 +97,36 @@ def response_terminate(self):
         if not self.have_pending_request():
             raise Exception("Attempting to reply with pending request")
         if not isinstance(self._req_received, communication.messages.TerminateRequest):
-            raise Exception("Replaying with terminate status to other type of message")
+            raise Exception("Replaying with `terminate` status to other type of message")
```
ejguan, line 100: Not related to this PR. I think we might want to change the `Exception` to something more specific so that we can easily handle those `Exception` easily.

torchdata/dataloader2/communication/iter.py
```diff
@@ -148,12 +150,43 @@ def DataPipeBehindQueues(source_datapipe, protocol, blocking_request_get=False):
             source_datapipe.reset_iterator()
             protocol.response_reset_iterator()
 
+        elif isinstance(request, communication.messages.PauseRequest):
+            dp_list = list_dps(traverse_dps(source_datapipe))
+            for dp in dp_list:
+                # TODO: Remove this condition after there is `pause` support for round-robin sharding
+                if isinstance(dp, (QueueWrapper, _IterateQueueDataPipes)):
+                    warnings.warn("There is no support for `pause` with round-robin sharding at the moment.")
```
ejguan, line 158: When it is dispatching process, the graph has to contain both `QueueWrapper` and `_IterateQueueDataPipes`. Let's check only one of them to trim the warning to once.

torchdata/dataloader2/communication/iter.py
```diff
@@ -148,12 +150,43 @@ def DataPipeBehindQueues(source_datapipe, protocol, blocking_request_get=False):
...
+                    dp.pause()
+
+            protocol.response_pause()
+            yield True  # Return control
+
+        elif isinstance(request, communication.messages.ResumeRequest):
+            dp_list = list_dps(traverse_dps(source_datapipe))
+            for dp in reversed(dp_list):
+                # TODO: Remove this condition after there is `resume` support for round-robin sharding
+                if isinstance(dp, (QueueWrapper, _IterateQueueDataPipes)):
```
ejguan, line 169: Ditto

torchdata/dataloader2/communication/iter.py
```diff
@@ -148,12 +150,43 @@ def DataPipeBehindQueues(source_datapipe, protocol, blocking_request_get=False):
...
+            protocol.response_resume()
+            yield True  # Return control
+
         elif isinstance(request, communication.messages.TerminateRequest):
             forever = False
             protocol.response_terminate()
 
         elif isinstance(request, communication.messages.GetNextRequest):
             while forever:
+                if protocol._pause:
```
ejguan, line 182: Can we add a method to `ProtocolServer` like `is_paused`. Relying on a private attribute is not a good practice.

test/dataloader2/test_proto_multi_rs.py
```diff
@@ -0,0 +1,303 @@
...
+    """
+
+    # Test DataPipes
+    n_elements = 10
+    dp1 = IterableWrapper(range(n_elements)).shuffle().sharding_filter()
+    double_pause_dp = dp1.prefetch().prefetch()
+    test_dps = [dp1, double_pause_dp]
+    mp_start_methods = ["fork", "forkserver", "spawn"]
+
+    def test_reading_service_pause_resume_0_worker(self) -> None:
```
ejguan, line 34: nit: can you combine the following two tests by parametrizing the number of workers and prefetch cnts

ejguan (commented):
The PR looks good in terms of test cases. I left a few comments below

NivekT:
CI is expected to fail until PyTorch core nightly updates

torchdata/dataloader2/communication/protocol.py
NivekT: Opened an issue: https://github.com/pytorch/data/issues/996

torchdata/dataloader2/dataloader2.py
NivekT: Would it be better to returning the value to the model first before executing pause? Or you think it will make 0 difference at all?

ejguan: It might be my PTSD caused by the racing generator problem 😢 TBH I don't have a strong opinion on it

torchdata/dataloader2/dataloader2.py
```diff
@@ -46,12 +48,20 @@ class DataLoader2Iterator(Iterator[T_co]):
...
+        self.limit_counter: Optional[int] = None
+        self.limit_threshold: Optional[int] = None
 
     def __next__(self) -> T_co:
         if self.iterator_id == self.dataloader.valid_iterator_id:
             self.dataloader._reset_iter = True
             try:
-                return next(self.dataloader._datapipe_iter)  # type: ignore[arg-type]
+                if self.dataloader._is_paused:
+                    raise StopIteration("DataLoader2 has been paused. `resume` must be called before continuing.")
```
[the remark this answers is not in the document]
ejguan: If that's the case, I don't think we should call `rs.finalize_iteration`

torchdata/dataloader2/reading_service.py
```diff
@@ -374,11 +374,48 @@ def clean_me(process, req_queue, res_queue):
...
+            end_datapipe = self._end_datapipe  # type: ignore[assignment]
+        if self.num_workers > 0:
+            end_datapipe.request_resume()
+        else:
+            raise RuntimeError(
+                "If you would like to use `resume` with `PrototypeMultiProcessingReadingService`, "
+                "please use more than 0 worker."
+            )
+        if self.main_prefetch_cnt > 0 and self.num_workers > 0:
+            self._end_datapipe.resume()  # type: ignore[union-attr]
```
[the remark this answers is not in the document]
NivekT: We are resuming in reverse order compared to pause.

The first resume is applied onto `_worker_consumer_dp`. The second resume is applied onto `_main_prefetch_dp`.

torchdata/datapipes/iter/util/distributed.py
```diff
@@ -229,3 +229,11 @@ def __setstate__(self, state):
...
         self._sync_counter = torch.tensor([0], dtype=torch.int32)
         self._done_callback = False
+
+    def pause(self):
+        if self._executor is not None:
+            self._executor.shutdown()
+            self._executor = None
+
+    def resume(self):
+        self._executor = _PrefetchExecutor(iter(self.datapipe), 1, self._callback_fn, self.timeout)
```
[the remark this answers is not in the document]
NivekT: Is there a way for it not to read from the beginning again? Halting the executor?

For now, I will raise an Exception.

torchdata/dataloader2/dataloader2.py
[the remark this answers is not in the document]
NivekT: I see. We should use `PauseIteration`, updating.

test/dataloader2/test_proto_multi_rs.py
NivekT: You mean `test_reading_service_pause_resume_0_worker` and `test_reading_service_pause_resume`? I think the expected behaviors are a bit different, hence I separated them

NivekT: I think I can use parameterize to replace the for loop though. I will do that

NivekT:
@nivekt has imported this pull request.  If you are a Meta employee, you can view this diff [on Phabricator](https://www.internalfb.com/diff/D41744759).

NivekT:
Addressed and resolved most of the comments but 2. CI is being re-ran after nightly updated.

test/dataloader2/test_proto_multi_rs.py
```diff
@@ -0,0 +1,311 @@
...
+#
+# This source code is licensed under the BSD-style license found in the
+# LICENSE file in the root directory of this source tree.
+
+
+import multiprocessing as mp
+import unittest
+from unittest import TestCase
+
+from test_dataloader2 import parametrize
```
ejguan, line 12: Let's import `parametrize` from `torch.testing._internal.common_utils` as well.

test/dataloader2/test_proto_multi_rs.py
```diff
@@ -0,0 +1,311 @@
...
+            res0.append(x)
+            if i in {2}:
+                with self.assertRaisesRegex(RuntimeError, r"pause"):
+                    dl0._pause()
+                with self.assertRaisesRegex(RuntimeError, r"resume"):
+                    dl0._resume()
+        dl0.shutdown()
+
+    @mp_ctx_parametrize
+    @dp_parametrize
```
ejguan, line 59: We can do
```py
@parametrize(
    "num_workers,worker_prefetch_cnt,main_prefetch_cnt",
    [
        (1, 0, 0),
        (1, 0, 2),
        (2, 0, 0),
        ...
    ]
)
```

If this fixture can be reused, you can define it like `mp_ctx_parametrize` at the beginning of this file

test/dataloader2/test_proto_multi_rs.py
```diff
@@ -0,0 +1,311 @@
...
+                sorted(res),
+                msg=f"The test is failing for rs{n + 1} with '{ctx}', num_workers = {rs.num_workers}, "
+                f"worker_prefetch_cnt = {rs.worker_prefetch_cnt}, "
+                f"main_prefetch_cnt = {rs.main_prefetch_cnt}",
+            )
+            dl.shutdown()
+
+    @mp_ctx_parametrize
+    @dp_parametrize
+    def test_reading_service_pause_stop_yield(self, ctx, dp) -> None:
```
ejguan, line 104: Ditto

test/dataloader2/test_proto_multi_rs.py
```diff
@@ -0,0 +1,311 @@
...
+            self.assertEqual(
+                3,
+                len(res),
+                msg=f"The test is failing for rs{n + 7} with '{ctx}', num_workers = {rs.num_workers}, "
+                f"worker_prefetch_cnt = {rs.worker_prefetch_cnt}, main_prefetch_cnt = {rs.main_prefetch_cnt}",
+            )
+            dl.shutdown()
+
+    @dp_parametrize
+    def test_reading_service_limit(self, dp) -> None:
```
ejguan, line 135: Ditto

torchdata/dataloader2/dataloader2.py
```diff
@@ -73,6 +91,46 @@ def __next__(self) -> T_co:
...
+            n_batches: Number of batches after which the DataLoader2 will pause
+        """
+        self.limit_counter = 0
+        self.limit_threshold = n_batches
+
+    def clear_limit(self) -> None:
+        """
+        Set the ``limit_threshold`` to ``None`` such that the iterator will not automatically call ``pause``
+        """
+        self.limit_threshold = None
```
ejguan, line 132: cc: @Miiira Do you think internal users prefer `clear_limit` or `limit(None)`/`limit()` to clean up the limit threshold?

ejguan (approved):
Overall LGTM. Thank you. I left a few comments to help simplify the test and a comment for internal users

torchdata/dataloader2/dataloader2.py
NivekT: Just to add more context, under this implementation both `clear_limit` and `limit(None)` works, but not `limit()`.

Let us know what would internal users prefer.

NivekT:
@nivekt has imported this pull request.  If you are a Meta employee, you can view this diff [on Phabricator](https://www.internalfb.com/diff/D41744759).

torchdata/dataloader2/dataloader2.py
Miiira: They are using `limit(None)` now so I'd assume same preference. Curious if `limit(None)` already works do we still need to expose `clear_limit()`?

NivekT: I personally found `clear_limit()` more intuitive but we can totally remove it

ejguan: If so, how about keeping APIs minimum. It's way easier to add one more method than removing one.

NivekT: Okay, let me take it out

NivekT:
@nivekt has imported this pull request.  If you are a Meta employee, you can view this diff [on Phabricator](https://www.internalfb.com/diff/D41744759).

NivekT:
@nivekt has imported this pull request.  If you are a Meta employee, you can view this diff [on Phabricator](https://www.internalfb.com/diff/D41744759).

NivekT:
@nivekt has imported this pull request.  If you are a Meta employee, you can view this diff [on Phabricator](https://www.internalfb.com/diff/D41744759).

torchdata/datapipes/iter/util/prefetcher.py
```diff
@@ -57,54 +58,56 @@ def __init__(self, source_datapipe, buffer_size: int = 10):
...
-                    else:
-                        # TODO: Calculate sleep interval based on previous availability speed
+        try:
+            prefetch_data = _PrefetchData(self.source_datapipe, self.buffer_size)
+            self.prefetch_data = prefetch_data
+            thread = threading.Thread(target=PrefetcherIterDataPipe.thread_worker, args=(prefetch_data,), daemon=True)
+            self.thread = thread
+            self.thread.start()
+
+            while prefetch_data.run_prefetcher:
```
ejguan, line 94: One more thing: Should this line be `while not prefetch_data.stop_iteration:`?

It's a little bit fuzzy to me if we are allowed to do `p

ejguan: I guess my main question is if `pause` of `DataPipe` is exposed to users.

torchdata/datapipes/iter/util/prefetcher.py
```diff
@@ -57,54 +58,56 @@ def __init__(self, source_datapipe, buffer_size: int = 10):
...
+        while not prefetch_data.stop_iteration:
+            while prefetch_data.run_prefetcher:
+                if len(prefetch_data.prefetch_buffer) < prefetch_data.buffer_size and not prefetch_data.stop_iteration:
+                    try:
+                        item = next(itr)
+                        prefetch_data.prefetch_buffer.append(item)
+                    except StopIteration:
+                        prefetch_data.stop_iteration = True
+                    except communication.iter.InvalidStateResetRequired:
+                        prefetch_data.stop_iteration = True
```
ejguan, line 75: Related to my last comment. `stop_iteration` should dominate `run_prefetcher`. When `stop_iterator` is met, there is no reason to process with prefetching, right?

ejguan: IIIUC, the only case that `stop_iteration` remains `False` when `run_prefetcher` is changed to `False` is `pause` called.

NivekT: What you stated should be `True`, but it requires changing other parts as well.

torchdata/datapipes/iter/util/prefetcher.py
NivekT: I think you are right. Though from a glance changing that `while` condition will also require changing the exception conditions.

Particularly, it should become `except StopIteration: prefetch_data.run_prefetcher = False` and the same for `except communication.iter.InvalidStateResetRequired`. 

Overall, it should be:
`prefetch_data.run_prefetcher` - indicates whether or not we should be prefetching, it should become `False` when the source runs out of elements, invalid state, termination, and maybe other exception raised.
`prefetch_data.stop_iteration` - this should. indicate when we should stop yielding elements, it should mostly be the same as above except when the source is out of elements, it should keep yielding until the buffer is empty.

PR #880, opened 2022-11-04
Add a masks option to filter files in s3 datapipe

### LINKED ISSUE #737: Support file mask on list_files_by_s3 like list_files
### 🚀 The feature

Support a file mask on `list_files_by_s3` just like `list_files` does.

### Motivation, pitch

```python
dp.list_files_by_s3(masks="*.csv")
```

is nicer than

```python
dp.list_files_by_s3().filter(filter_fn=lambda filename: filename.endswith(".csv"))
```

### Alternatives

_No response_

### Additional context

_No response_

sebathomas:
Add a new option to the constructor of `S3FileListerIterDataPipe` that allows to filter the list of files with a pattern, using the existing filter function `match_masks`.

I added a unit test for the s3 datapipe and I tested it on my machine with a real S3 bucket.

Fixes #737.

### DIFF
--- a/test/test_local_io.py
+++ b/test/test_local_io.py
@@ -674,7 +674,7 @@ class TestDataPipeLocalIO(expecttest.TestCase):
 
     # TODO(120): this test currently only covers reading from local
     # filesystem. It needs to be modified once test data can be stored on
-    # gdrive/s3/onedrive
+    # gdrive/onedrive
     @skipIfNoIoPath
     def test_io_path_file_lister_iterdatapipe(self):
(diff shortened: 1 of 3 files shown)


NivekT (commented):
The implementation looks fine. Do we want to add an extra argument to this DataPipe? 

@ejguan WDYT?

ejguan (approved):
LGTM, thank you

NivekT:
We will need to update some things to make the test compatible internally. We will keep you posted.

PR #883, opened 2022-11-04
Fix max_token_bucketize to accept incomparable data

### LINKED ISSUE #881: MaxTokenBucktizer for dict type item
### MaxTokenBucktizer For Dict Type Item

I build a dataset which type of items is a python dictionary, and when I try to use MaxTokenBucktizer, I got an error
image
And I check the source code of torchdata, it caused by heappush method which not support python dict.
image

Could another data structures be used without affecting performance to store buffer data?  I think it's very common to use dictionary as item type..

### Versions

pytorch 1.12.1
torchdata 0.4.1

ejguan:
Fixes #881

- Add `PrioritizedItem` to store data and token_length for the sake of incomparable data type.
- Change the test to validate the length rather than exact data because the data itself won't be ordered in `heapq`.

### DIFF
--- a/torchdata/datapipes/iter/transform/bucketbatcher.py
+++ b/torchdata/datapipes/iter/transform/bucketbatcher.py
@@ -7,14 +7,17 @@
 import heapq
 import random
 
+from dataclasses import dataclass, field
 from functools import partial
-from typing import Callable, Iterator, List, Optional, TypeVar
+from typing import Callable, Generic, Iterator, List, Optional, TypeVar
 
 import torch
 
 from torchdata.datapipes import DataChunk, functional_datapipe
 from torchdata.datapipes.iter import IterDataPipe
 
+
+T = TypeVar("T")
 T_co = TypeVar("T_co", covariant=True)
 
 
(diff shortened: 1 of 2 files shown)


torchdata/datapipes/iter/transform/bucketbatcher.py
```diff
@@ -181,13 +184,18 @@ def _default_len_fn(token):
     return len(token)
 
 
-def _token_len_fn(token, len_fn):
-    return len_fn(token), token
+@dataclass(order=True)
```
NivekT, line 187:
```suggestion
@dataclass(order=True, frozen=True)
```
super nit: feel free to ignore, we can set `frozen` if we do not expect the need to modify it

NivekT (approved):
Not very related , non-blocking question:

We say:
> To increase variablity, apply ``torchdata.datapipes.iter.Shuffler`` before and after this DataPipe,

Do users want to use `in_batch_shuffle` instead of `shuffle` after this DataPipe to increase variability? Or maybe both?

torchdata/datapipes/iter/transform/bucketbatcher.py
ejguan: Makes sense.

Issue #885, opened 2022-11-04
Determinsim about Local shuffle/random_op after `sharding_filter`


ejguan:
## Current state of determinism

Using `DataLoader2` + `PrototypeMultiProcessingReadingService` as an example:
1. Before each iteration starts, a distributed shared seed will be generated ([link](https://github.com/pytorch/data/blob/b1b3406389fa6665ec3422ae5e1ebf65ae7e2938/torchdata/dataloader2/reading_service.py#L282-L285))
2. With multiprocessing, each subprocess will reset all of shuffle operations to the same random seeds at the beginning of each iteration based on the distributed shared seed in step 1. ([link](https://github.com/pytorch/data/blob/b1b3406389fa6665ec3422ae5e1ebf65ae7e2938/torchdata/dataloader2/reading_service.py#L212-L217))
3. And, `torch`, `numpy` and `python.random` will get a different process-local seeds for each subprocess ([link](https://github.com/pytorch/data/blob/b1b3406389fa6665ec3422ae5e1ebf65ae7e2938/torchdata/dataloader2/reading_service.py#L219-L233))

## Additional feature

For the step 2 in the last section, we set the same shuffle seed across distribtued/mp workers because we want to make sure the shuffled data can be sharded in a mutually exclusive and collectively exhaustive manner.
An additional feature is needed to make sure all random operations after `sharding_filter` having the different seeds across workers to preserve fully data randomization.

Let's say we have a pipeline as:
```
data_source.shuffle().sharding_filter().map(fn).batch(8).shuffle()
```
We will have the random state shared for the first `shuffle`, but different states for the second `shuffle`. And, those states should be generated in a deterministic manner so we will be able to reproduce it.

### Versions

main branch

PR #888, opened 2022-11-07
[1/n] Add graph function to list DataPipes from graph

ejguan:
Add a `list_dps` function to list `DataPipes` from the graph.
- It's similar to [`get_all_graph_pipes `](https://github.com/pytorch/pytorch/blob/896fa8c5c9b0191c9621e04ab5e20057614d48ad/torch/utils/data/graph_settings.py#L19) from pytorch core
- An extra argument of `exclude_dps` to exclude the `DataPipe` and its prior graph from the result.

Reason to add this function:
- It's required to set random states differently for DataPipe before/after `sharding_filter`
```py
graph = traverse_dps(datapipe)
sf_dps = find_dps(graph, ShardingFilter)

# DataPipes prior to `sharding_filter`
p_dps = []
for sf_dp in sf_dps:
    p_dps.extend(list_dps(traverse_dps(sf_dp)))

# DataPipes after `sharding_filter`
a_dps = list_dps(graph, exclude_dps=sf_dps)
```

Step 1 for https://github.com/pytorch/data/issues/885


torchdata/dataloader2/graph/utils.py
```diff
@@ -34,6 +30,33 @@ def helper(g) -> None:  # pyre-ignore
     return dps
 
 
+def list_dps(graph: DataPipeGraph, exclude_dps: Optional[Union[DataPipe, List[DataPipe]]] = None) -> List[DataPipe]:
+    r"""
+    Given the graph of DataPipe generated by ``traverse_dps`` function,
+    return DataPipe instances excluding the provided DataPipe and its prior DataPipe graph.
+    """
+    dps: List[DataPipe] = []
```
VitalyFedyunin, line 38: Are you ok to have 2 entries of the same datapipe here?
I'm talking about:
```
B, C = A.fork(2)
D = B.concat(C).map()
```
Your solution will end up listing  datapipes twice.

ejguan: I have a `cache` object to prevent the same `DataPipe` in the list for multiple times.

torchdata/dataloader2/graph/utils.py
```diff
@@ -34,6 +30,33 @@ def helper(g) -> None:  # pyre-ignore
     return dps
 
 
+def list_dps(graph: DataPipeGraph, exclude_dps: Optional[Union[DataPipe, List[DataPipe]]] = None) -> List[DataPipe]:
```
NivekT, line 33: If a DataPipe is in `exclude_dps`, its predecessors will also be ignored. It seems like that is intended, if so, can we mention that in the docstring?

torchdata/dataloader2/graph/utils.py
```diff
@@ -34,6 +30,33 @@ def helper(g) -> None:  # pyre-ignore
...
+            cache.add(id(exclude_dps))
+        else:
+            for dp in exclude_dps:  # type: ignore[union-attr]
+                cache.add(id(dp))
+
+    def helper(g) -> None:  # pyre-ignore
+        for dp_id, (dp, src_graph) in g.items():
+            if dp_id not in cache:
+                cache.add(dp_id)
+                dps.append(dp)
```
NivekT, line 52: Side note, this doesn't guarantee any order and returns a list. This should suffice for the current use case.

I wonder if people will want a `Dict` (basically just `traverse_dps` without the IDs). To be clear, this seems fine to me for now.

NivekT (approved):
LGTM!

torchdata/dataloader2/graph/utils.py
ejguan: If we treat the graph as a tree, there can be in different orders like how do we traverse a tree. Pre-order; post-order, etc. It highly depends on users' preference especially our graph can be more complicated than a tree. I would say let's wait until we have received any use case.

VitalyFedyunin (approved):

PR #894, opened 2022-11-14
[2/n] Add setting function to set seeds to the graph

ejguan:
### Changes

- Update `list_dps` ~with a new argument (`exclusive`)~ to make sure all `DataPipes` will be excluded even though they are the predecessors of the non-excluded `DataPipe`. 
- Update `list_dps` to use BFS to return the list of `DataPipes`.
- Add `_set_worker_seed_for_dp_graph ` to set seeds for `DataPipes` before `sharding_filter` with the same seed generator and set different seeds for `DataPipes` after `sharding_filter` with the worker-local seed generator.


torchdata/dataloader2/graph/utils.py
```diff
@@ -30,21 +30,31 @@ def helper(g) -> None:  # pyre-ignore
...
+            exclude_dps = [
+                exclude_dps,
+            ]
+        for exclude_dp in exclude_dps:  # type: ignore[union-attr]
+            assert isinstance(exclude_dp, (IterDataPipe, MapDataPipe))
+            if exclusive:
+                for dp in list_dps(traverse_dps(exclude_dp)):  # type: ignore[arg-type]
+                    cache.add(id(dp))
+            else:
+                cache.add(id(exclude_dp))
```
ejguan, line 57: Before this PR, `cache` would only store the id of `exclude_dps`. However, there are scenarios like:
```mermaid
graph TD;
DP1-->DP2;
DP2-->DP3;
DP1-->DP3;
DP3-->output;
```
If we exclude `DP2`, `DP1` will still be included in the result as `[DP1, DP3]`. After this PR:
- `exclusive=True`: Return `[DP3]`
- `exclusive=False`: Return `[DP1, DP3]` (previous behavior)

BTW, for the argument name, appreciate any recommendation based on the behavior.

torchdata/dataloader2/graph/settings.py
```diff
@@ -0,0 +1,69 @@
...
+    return int(torch.empty((), dtype=dtype).random_(generator=rng).item())
+
+
+def _seed_datapipe(datapipe: DataPipe, seed_generator: torch.Generator) -> DataPipe:
+    seed = _generate_random_seed(seed_generator)
+    datapipe.set_seed(seed)
+    return datapipe
+
+
+def _set_graph_seeds(datapipe: DataPipe, seed_generator: torch.Generator, worker_id: int = 0) -> DataPipe:
```
ejguan, line 34: I am using a private name because I don't like the name of this function. LMK if you have any better idea.

torchdata/dataloader2/graph/settings.py
```diff
@@ -0,0 +1,69 @@
...
+    r"""
+    Set seeds to the graph of ``DataPipes`` based on a Seed Generator. All random ``DataPipes`` prior to
+    ``ShardingFilter`` will be set seeds by the same Seed Generator to preserve the same random state
+    across distributed/non-distributed workers. And, the random ``DataPipes`` after ``ShardingFilter``
+    will be set seeds by the worker-local Seed Generator deterministically created based on ``worker_id``.
+    """
```
NivekT, line 40: Add a note about:
1. `worker_seed = base_seed + worker_id`, currently no option to override
2. If `ShardingFilter` is not used, shuffle seeds (and any other RNG) in workers will be different.
3. Any discrepancy between DL1 vs DL2 (might need randomness test??)

torchdata/dataloader2/graph/settings.py
NivekT: 1. `_set_worker_seed`
2. `_set_worker_seed_for_dp_graph`

torchdata/dataloader2/graph/utils.py
NivekT: Just change the behavior to `exclusive=True` without adding the arg and update the docstring/example.

NivekT (approved):
LGTM! A few comments:

torchdata/dataloader2/graph/settings.py
ejguan: I will add those comments to the ReadingService in the next PR. This function is not exposed to users for now and no online doc for it as well.

NivekT (approved):
The BFS looks good to me. Can we add a test that confirms the result's ordering? (It might already exist.)

PR #896, opened 2022-11-15
Fix tests to access S3

ejguan:
Fixes failed tests as the open-sourced Dataset has changed.


NivekT (approved):
Thanks for spotting and fixing this!

PR #897, opened 2022-11-15
[DLv2] Move serialization functions to graph and add clone function

ejguan:
### Changes

- Move `_wrap_and_copy_dp` and `_copy` from DataLoader2 to graph module -> clean up `dataloader2.py`
- Add `clone` function. (This is going to be used by `Prototyping2RS`)

### DIFF
--- a/torchdata/dataloader2/graph/__init__.py
+++ b/torchdata/dataloader2/graph/__init__.py
@@ -6,12 +6,6 @@
 
 
 from torch.utils.data.graph import DataPipe, DataPipeGraph, traverse_dps
-from torchdata.dataloader2.graph.serialization import (
-    clone,
-    deserialize_datapipe,
-    serialize_datapipe,
-    wrap_datapipe_for_serialization,
-)
 
 from torchdata.dataloader2.graph.utils import find_dps, list_dps, remove_dp, replace_dp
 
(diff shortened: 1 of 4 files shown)


torchdata/dataloader2/graph/__init__.py
```diff
@@ -6,18 +6,28 @@
...
     "DataPipeGraph",
+    "clone",
+    "deserialize_datapipe",
     "find_dps",
     "list_dps",
     "remove_dp",
     "replace_dp",
+    "serialize_datapipe",
     "traverse_dps",
+    "wrap_datapipe_for_serialization",
```
ejguan, line 30: I still hesitate if we want to expose those functions. I personally think those functions should not be used by users.
If you agree, I can remove them from `__all__` and make them staying at `torchdata.dataloader2.graph.serialization` with an underscore.
@NivekT WDYT?

NivekT: Agree. Unless there are use cases of these functions in a custom ReadingService, I agree that we probably don't want users to use these. Let's make them private since the only use case I can see is for `DataLoader2`.

NivekT (approved):
LGTM with one comment!

https://github.com/pytorch/data/pull/746 will need to rebase if this PR lands first. Unless #746 magically lands first today or tomorrow...

PR #903, opened 2022-11-17
[DataPipe] Add RoundRobinDemux

ejguan:
The only reason I added this DataPipe without directly using `enumereate().demux().drop_index()` is this `RoundRobinDemux` should provide a valid length. So, this PR needs https://github.com/pytorch/pytorch/pull/89216 landed

And, this DataPipe will be used for `Proto2RS`.

Side change: Move `Unzip` to `combining.py` as well.


ejguan:
Failed tests should be fixed the the PR in PyTorch Core landed.

ejguan:
Failed tests should be fixed by https://github.com/pytorch/data/pull/905

torchdata/datapipes/iter/util/combining.py
```diff
@@ -213,3 +216,139 @@ def __iter__(self) -> Iterator:
...
         return len(self.source_iterdatapipe)
+
+
+def _drop_index(idx_data):
+    idx, data = idx_data
+    return data
+
+
+@functional_datapipe("round_robin_demux")
+class RoundRobinDemultiplexerIterDataPipe(IterDataPipe):
```
wenleix, line 227: curious: why `demux` is PyTorch core (https://github.com/pytorch/pytorch/blob/7a2930b357a4e62bb0bab53bb0d23c607b6ede38/torch/utils/data/datapipes/iter/combining.py#L305-L306) but `round_robin_demux` is in TorchData?

torchdata/datapipes/iter/util/combining.py
```diff
@@ -213,3 +216,139 @@ def __iter__(self) -> Iterator:
...
+    Args:
+        datapipe: Iterable DataPipe being filtered
+        num_instances: number of instances of the DataPipe to create
+        buffer_size: this defines the maximum number of inputs that the buffer can hold across all child
+            DataPipes while waiting for their values to be yielded.
+            Defaults to ``1000``. Use ``-1`` for the unlimited buffer.
+
+    Examples:
+        >>> from torchdata.datapipes.iter import IterableWrapper
+        >>> def odd_or_even(n):
```
wenleix, line 241: this doesn't seem to be used in this example?

torchdata/datapipes/iter/util/combining.py
```diff
@@ -213,3 +216,139 @@ def __iter__(self) -> Iterator:
...
+            raise ValueError(f"Expected `num_instaces` larger than 0, but {num_instances} is found")
+        if num_instances == 1:
+            warnings.warn(
+                "The operation of `round_robin_demux` with `num_instances=1` is an no-op and returns the provided `datapipe` directly"
+            )
+            return datapipe
+
+        datapipe = datapipe.enumerate()
+        container = _RoundRobinDemultiplexerIterDataPipe(datapipe, num_instances, buffer_size=buffer_size)
+        return [_ChildDataPipe(container, i).map(_drop_index) for i in range(num_instances)]
```
wenleix, line 262: As a Python noob, TIL I learnt `X()` in Python may not return a instance of `X`

https://www.geeksforgeeks.org/__new__-in-python/

> ...  because other class constructors can be called by __new__ method or it can simply return other objects as an instance of this class.

Will `_ChildDataPipe.__init__` be called for each instance in the list? ~

torchdata/datapipes/iter/util/combining.py
```diff
@@ -213,3 +216,139 @@ def __iter__(self) -> Iterator:
 
     def __len__(self) -> int:
         return len(self.source_iterdatapipe)
+
+
+def _drop_index(idx_data):
+    idx, data = idx_data
```
wenleix, line 222: curious: why not 
```
_, data = idx_data
```

here?

torchdata/datapipes/iter/util/combining.py
```diff
@@ -213,3 +216,139 @@ def __iter__(self) -> Iterator:
...
+        if num_instances < 1:
+            raise ValueError(f"Expected `num_instaces` larger than 0, but {num_instances} is found")
+        if num_instances == 1:
+            warnings.warn(
+                "The operation of `round_robin_demux` with `num_instances=1` is an no-op and returns the provided `datapipe` directly"
+            )
+            return datapipe
+
+        datapipe = datapipe.enumerate()
+        container = _RoundRobinDemultiplexerIterDataPipe(datapipe, num_instances, buffer_size=buffer_size)
```
wenleix, line 261: curious: can we just do 
```python
_DemultiplexerIterDataPipe(
    datapipe, 
    num_instances, 
    partial(num_instances, _round_robin_fn), 
    False, 
    buffer_size
)
```

where `_round_robin_fn` is:
```python
def _round_robin_fn(num_instances: int, idx_data) -> int:
        idx, _ = idx_data
        return idx % num_instances
```


Is that because you want to override `_DemultiplexerIterDataPipe.get_length_by_instance` ?

wenleix (approved):
LGTM % some noob/minor comments

wenleix:
Why demux and unzip data pipes are in `combining.py` ? (shouldn't they be in `splitting.py`)?

torchdata/datapipes/iter/util/combining.py
ejguan: Historical reason. We put lots of DataPipe in core but we plan to move them to torchdata eventually. To reduce the amount of work, I think it's better to put `round_robin_demux` to torchdata directly.

torchdata/datapipes/iter/util/combining.py
ejguan: Good catch.

torchdata/datapipes/iter/util/combining.py
ejguan: Yeah. `demux` doesn't provide a valid length but this `DataPipe` should have a deterministic result. Let me know if you have better way to do so.

ejguan:
> Why demux and unzip data pipes are in `combining.py` ? (shouldn't they be in `splitting.py`)?

Aha. We categorized DataPipe into a few sectors. See: https://pytorch.org/data/beta/index.html We put them together as they are the counter party of `mux`, `zip`, etc., and we don't have such a mount of `DataPipes`.

When https://github.com/pytorch/data/issues/293 is done, we probably need to make them into `combining.py` and `splitting.py`.

torchdata/datapipes/iter/util/combining.py
wenleix: noob question : what's "deterministic result" here mean? 

Especially I was somehow under the impression that in general try to avoid query `len(datapipe)` for large datapipe , since it probably needs to read all data and it's expensive?

ejguan: If the prior DataPipe provides a length, we should be able to get the length ahead of time.
For `demux` which uses a `classify_fn`, there is no way for us to know the length and the length can vary across epoches.

> in general try to avoid query `len(datapipe)` for large datapipe , since it probably needs to read all data and it's expensive?

It's true. But, we do provide `DataPipe` that can manually assign length to the DataPipe https://github.com/pytorch/data/blob/67afe0c36f5f93d34fde7026e7e95835c1160bb3/torchdata/datapipes/iter/util/header.py#L62

Issue #904, opened 2022-11-18
Azure test is broken due to missing authentications


ejguan:
### 🐛 Describe the bug

See example CI test result:
https://github.com/pytorch/data/actions/runs/3496367745/jobs/5854203850#step:13:527

To resolve it, we need to properly set environment variables `AZURE_CLIENT_ID`, `AZURE_TENANT_ID` and `AZURE_CLIENT_SECRET` as shown in https://github.com/Azure/azure-sdk-for-python/blob/main/sdk/identity/azure-identity/TROUBLESHOOTING.md#troubleshoot-environmentcredential-authentication-issues

To temporarily mitigate the problem, I am going to skip the test

### Versions

main

NivekT:
I think this can be closed after #905 has landed?

ejguan:
I prefer keeping it open since the test failure hasn't been fixed. I just skipped it.

PR #905, opened 2022-11-18
Fix tests for Azure and DataLoader

ejguan:
### Changes

- For Azure test: Please see https://github.com/pytorch/data/issues/904
- For DataLoader test, https://github.com/pytorch/pytorch/pull/88972 introduces an Error when `prefetch_factor` is specified when `num_workers>0`.

### DIFF
--- a/torchdata/dataloader2/reading_service.py
+++ b/torchdata/dataloader2/reading_service.py
@@ -369,7 +369,7 @@ class MultiProcessingReadingService(ReadingServiceInterface):
     pin_memory: bool
     timeout: float
     worker_init_fn: Optional[Callable[[int], None]]
-    prefetch_factor: int
+    prefetch_factor: Optional[int]
     persistent_workers: bool
 
     def __init__(
@@ -379,7 +379,7 @@ class MultiProcessingReadingService(ReadingServiceInterface):
         timeout: float = 0,
         worker_init_fn: Optional[Callable[[int], None]] = None,
         multiprocessing_context=None,
-        prefetch_factor: int = 2,
+        prefetch_factor: Optional[int] = None,
         persistent_workers: bool = False,
     ) -> None:
         self.num_workers = num_workers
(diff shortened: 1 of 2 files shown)


torchdata/dataloader2/reading_service.py
```diff
@@ -369,7 +369,7 @@ class MultiProcessingReadingService(ReadingServiceInterface):
     pin_memory: bool
     timeout: float
     worker_init_fn: Optional[Callable[[int], None]]
-    prefetch_factor: int
+    prefetch_factor: Optional[int]
```
wenleix, line 372: noob question: why it's called "factor" ?

wenleix (approved):
LGTM

torchdata/dataloader2/reading_service.py
ejguan: Argument from DataLoader1
https://github.com/pytorch/pytorch/blob/837ca8f344380f2356b01662f215ff561b09401f/torch/utils/data/dataloader.py#L176

`prefetch_factor * num_workers`

dahsh: It doesn't have to be done in this PR; maybe add docsting for arguments in MultiProcessingReadingServicer? Thanks.

dahsh (approved):

torchdata/dataloader2/reading_service.py
wenleix: > ``2`` means there will be a total of 2 * num_workers batches prefetched across all workers.

noob question: it sounds like different data loader workers may have different numbers of prefetched batches? -- because otherwise, the doc should just say "2 means each worker will prefetech 2 batches"

ejguan: For `DataLoader1`, each worker always prefetches `prefetch_factor` batches.

PR #907, opened 2022-11-23
Add worker_init_fn and worker_reset_fn to ProtoMPRS

ejguan:
### Changes

- Add `worker_init_fn` and `worker_reset_fn` to `PrototypeMultiprocessingReadingService`. This should help `SequentialRS` to extend `worker_reset_fn` per epoch (still need to hack by assigning `worker_reset_fn` manually)
- Move `_process_init_fn` and `_process_reset_fn` and move random number/tensor generation to `torchdata.dataloader2.utils`
- Instead of using explicit arguments for `init_fn` or `reset_fn`, add `DistInfo` and `WorkerInfo` to pass relevant information to worker. This should help extensibility to add extra information to those `Info` class
-

### DIFF
--- a/torchdata/dataloader2/communication/eventloop.py
+++ b/torchdata/dataloader2/communication/eventloop.py
@@ -31,13 +31,11 @@ __all__ = [
 ]
 
 
-def DataPipeToQueuesLoop(source_datapipe, req_queue, res_queue, call_on_process_init=None, call_on_epoch_reset=None):
+def DataPipeToQueuesLoop(source_datapipe, req_queue, res_queue, call_on_process_init=None):
     r"""
     Initialize with the given init function, set the appropriate pipe and protocol server type, and
     create a loop with the protocol server.
     """
-    if call_on_process_init is not None:
-        call_on_process_init(source_datapipe)
     if isinstance(source_datapipe, IterDataPipe):
         pipe_type = communication.iter
         protocol_type = communication.protocol.IterDataPipeQueueProtocolServer
@@ -46,18 +44,19 @@ def DataPipeToQueuesLoop(source_datapipe, req_queue, res_queue, call_on_process_
         protocol_type = communication.protocol.MapDataPipeQueueProtocolServer  # type: ignore[assignment]
     else:
         raise Exception("Only supports IterDataPipe or MapDataPipe, got", source_datapipe)
+    if call_on_process_init is not None:
+        call_on_process_init(source_datapipe)
 
     torch.set_num_threads(1)
     for _ in pipe_type.DataPipeBehindQueues(
         source_datapipe,
         protocol_type(req_queue, res_queue),
         blocking_request_get=True,
-        reset_epoch_fn=call_on_epoch_reset,
     ):
         pass
 
 
-def SpawnProcessForDataPipeline(multiprocessing_ctx, datapipe, call_on_process_init=None, call_on_epoch_reset=None):
+def SpawnProcessForDataPipeline(multiprocessing_ctx, datapipe, call_on_process_init=None):
     r"""
     Given a DataPipe, starts a new process with ``DataPipeToQueuesLoop`` as target,
     and returns ``(process, req_queue, res_queue)``.
@@ -65,7 +64,7 @@ def SpawnProcessForDataPipeline(multiprocessing_ctx, datapipe, call_on_process_i
     req_queue = multiprocessing_ctx.Queue()
     res_queue = multiprocessing_ctx.Queue()
     process = multiprocessing_ctx.Process(
-        target=DataPipeToQueuesLoop, args=(datapipe, req_queue, res_queue, call_on_process_init, call_on_epoch_reset)
+        target=DataPipeToQueuesLoop, args=(datapipe, req_queue, res_queue, call_on_process_init)
     )
     return process, req_queue, res_queue
 
--- a/torchdata/dataloader2/reading_service.py
+++ b/torchdata/dataloader2/reading_service.py
@@ -5,13 +5,11 @@
 # LICENSE file in the root directory of this source tree.
 
 
-import functools
 import multiprocessing as mp
-import random
 
 from abc import ABC, abstractmethod
-
 from datetime import timedelta
+from functools import partial
 from typing import Callable, List, Optional
 
 import torch
@@ -22,14 +20,15 @@ from torch.utils.data import DataLoader
 from torchdata._constants import default_dl2_worker_join_timeout_in_s, default_timeout_in_s
 from torchdata.dataloader2 import communication
 from torchdata.dataloader2.graph import DataPipe
+from torchdata.dataloader2.utils import (
+    DistInfo,
+    generate_random_scalar_tensor,
+    process_init_fn,
+    process_reset_fn,
+    WorkerInfo,
+)
+from torchdata.dataloader2.utils.worker import _ExtraInfo
 from torchdata.datapipes.iter import FullSync, IterableWrapper, IterDataPipe
-
-try:
-    import numpy
-
-    HAS_NUMPY = True
-except ModuleNotFoundError:
-    HAS_NUMPY = False
 
 
 class ReadingServiceInterface(ABC):
@@ -115,10 +114,6 @@ def _collate_no_op(batch):
     return batch[0]
 
 
-def _generate_random_seed(rng: Optional[torch.Generator] = None, dtype: torch.dtype = torch.int64) -> torch.Tensor:
-    return torch.empty((), dtype=dtype).random_(generator=rng)
-
-
 class _IterateQueueDataPipes(IterDataPipe):
     r"""
     Takes in ``QueueWrapper``s and iterates through them in a round-robin manner to get batches one-by-one.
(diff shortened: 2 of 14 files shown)


torchdata/dataloader2/reading_service.py
```diff
@@ -184,67 +181,57 @@ class PrototypeMultiProcessingReadingService(ReadingServiceInterface):
...
-        torch_seed = _generate_random_seed(worker_seed_generator).item()
-        torch.manual_seed(torch_seed)
-
-        if HAS_NUMPY:
-            # Numpy only accepts uint32 as the seed
-            np_seed = _generate_random_seed(worker_seed_generator, torch.int32).item()
-            if np_seed < 0:
-                np_seed = 2 ** 32 + np_seed
-            numpy.random.seed(np_seed)
+        self._dist_info = DistInfo(1, 0)
```
wenleix, line 234: do we plan to support `world_size > 1` in the future? (understand it's a prototype now~)

torchdata/dataloader2/reading_service.py
```diff
@@ -184,67 +181,57 @@ class PrototypeMultiProcessingReadingService(ReadingServiceInterface):
...
     _pg: Optional[dist.ProcessGroup]
-    _world_size: int
-    _rank: int
+    _dist_info: DistInfo
 
     def __init__(
         self,
         num_workers: int = 0,
-        multiprocessing_context=None,
+        multiprocessing_context: Optional[str] = None,
```
wenleix, line 212: is this  `multiprocessing_context` expected to be one of the 'fork', 'spawn' and 'forkserver'  ?

torchdata/dataloader2/utils/worker.py
```diff
@@ -0,0 +1,121 @@
...
+    """
+    # This function will receive worker local copy of datapipe and reset function from ``initialize_iteration``
+    worker_seed_generator = torch.Generator()
+    worker_seed_generator.manual_seed(extra_info.shared_seed)
+    torch.utils.data.graph_settings.apply_random_seed(
+        datapipe,
+        worker_seed_generator,
+    )
+    # Set different seeds across distributed workers
+    global_worker_id = worker_info.worker_id * dist_info.world_size + dist_info.rank
```
wenleix, line 101: If I understand correctly, Say we have 2 distributed training process (world_size=2) and 3 data loading subprocess per training process (num_workers=3). the global_worker_id assignment is like following:

```
rank        0   1
------------------
worker0     0   1     
worker1     2   3
worker2     4   5
```

Essentially the global worker id with rank 0 are (0, 2, 4) while the global worker id with rank 1 are (1, 3, 5). Wondering why not make global worker id with the same rank continuous? ~

wenleix (approved):
Thanks!

torchdata/dataloader2/communication/eventloop.py
```diff
@@ -46,26 +44,27 @@ def DataPipeToQueuesLoop(source_datapipe, req_queue, res_queue, call_on_process_
         protocol_type = communication.protocol.MapDataPipeQueueProtocolServer  # type: ignore[assignment]
     else:
         raise Exception("Only supports IterDataPipe or MapDataPipe, got", source_datapipe)
+    if call_on_process_init is not None:
```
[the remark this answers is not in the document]
ejguan: I moved it to make sure the Error is raised earlier before calling `call_on_process_init`.

torchdata/dataloader2/reading_service.py
```diff
@@ -184,67 +181,57 @@ class PrototypeMultiProcessingReadingService(ReadingServiceInterface):
         multiprocessing_context (str, optional): Multiprocessing starting method.
             If method is None then the default context is returned.
             Otherwise, method should be 'fork', 'spawn'.
+        prefetch_worker: (int, 10 by default): Number of data will be prefetched at
```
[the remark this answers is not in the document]
ejguan: How about `worker_prefetch_factor`?

torchdata/dataloader2/reading_service.py
```diff
@@ -253,32 +240,31 @@ def initialize(self, datapipe: DataPipe) -> DataPipe:
         creates subprocesses.
         """
         if dist.is_available() and dist.is_initialized():
-            self._world_size = dist.get_world_size()
-            self._rank = dist.get_rank()
+            _world_size = dist.get_world_size()
+            _rank = dist.get_rank()
+            self._dist_info = DistInfo(_world_size, _rank)
```
ejguan, line 245: cc: @wenleix Here is the place fetching distributed information lazily.

torchdata/dataloader2/reading_service.py
ejguan: It's supporting `world_size > 1`. And, in some cases like `lightning`, they initialize distributed environment after `DataLoader` is constructed. So, I have to make sure `distributed` call moved to `__iter__`.

torchdata/dataloader2/reading_service.py
ejguan: Not always. For windows, only `spawn` can be supported. That's the reason I checked the validity of `mp_context` using `mp.get_all_start_methods`.
See: https://docs.python.org/3/library/multiprocessing.html#multiprocessing.get_all_start_methods

torchdata/dataloader2/utils/worker.py
ejguan: > Essentially the global worker id with rank 0 are (0, 2, 4) while the global worker id with rank 1 are (1, 3, 5). Wondering why not make global worker id with the same rank continuous? ~

Aha, you found a tricky problem we found earlier. We can't make the same rank continuous because we want to make sure data are sharded as evenly as possible. Using the same example with `world_size=2, num_workers=3`, if the total number of data is 8 and we do continuous global worker id per rank, the length of data in rank 0 will be 5 but rank 1 will be 3. And, if I make the `global_worker_id` as the current implementation, both ranks will be able to fetch 4 data

torchdata/dataloader2/utils/worker.py
```diff
@@ -0,0 +1,121 @@
...
+
+@dataclass(frozen=True)
+class _ExtraInfo:
+    r"""
+    Message class for extra arguments.
+    """
+    shared_seed: int
+
+
+def process_init_fn(
```
ejguan, line 61: We have to keep `process_init_fn` separately because `apply_sharding` can be called once per sharding-level cc: @NivekT

torchdata/dataloader2/reading_service.py
```diff
@@ -184,67 +181,57 @@ class PrototypeMultiProcessingReadingService(ReadingServiceInterface):
...
 
     def __init__(
         self,
         num_workers: int = 0,
-        multiprocessing_context=None,
-        prefetch_worker: int = 10,
-        prefetch_mainloop: int = 10,
+        multiprocessing_context: Optional[str] = None,
+        worker_prefetch_cnt: int = 10,
+        main_prefetch_cnt: int = 10,
```
ejguan, line 214: Let me know if you have different opinion on those names

NivekT (approved):
I was wondering if we will need something similar for `DistributedRS` but it seems like it may not be necessary. And we can do that when needed.

torchdata/dataloader2/reading_service.py
wenleix: Right, i mean it cannot be out of these three options~ (although not all of them are supported~)

wenleix:
@NivekT 

> I was wondering if we will need something similar for DistributedRS but it seems like it may not be necessary. And we can do that when needed.

Agree. Looks like if we just want to have "co-located data reading and training" (a.k.a "onbox"), this multi-process reading service should be good enough~

By `DistributedRS`, do you mean "disaggregated data reading and training" (or at least hybrid). One scenario I can think this is useful is when we cannot saturate trainer with just local CPUs.

torchdata/dataloader2/utils/worker.py
wenleix: @ejguan  I see. Because the sharding mechanism is something like `tuple_id % total_worker == global_worker_id`? So when it's not divisible, lower "global_worker_id" will get the "extra one row". Result in more rows in low rank worker?~

torchdata/dataloader2/reading_service.py
ejguan: You are right

torchdata/dataloader2/utils/worker.py
ejguan: I am not 100% sure that I understand.
We will try to fill in low workers for all nodes from low rank to high rank.

ejguan:
> By `DistributedRS`, do you mean "disaggregated data reading and training" (or at least hybrid). One scenario I can think this is useful is when we cannot saturate trainer with just local CPUs.

@wenleix 
DistributedRS mean this [one](https://github.com/pytorch/data/blob/45bc070c531865540c55b81859a49f5ee3ba7180/torchdata/dataloader2/reading_service.py#LL410C7-L410C32). It's just on-trainer distributed not disaggregated. Disaggregated requires more integration with customers' backend system that we have zero control (I am not sure if we can work with TorchX to enable this by TorchData directly)
We currently move all distributed code into `ProtoRS` as well to support users' request (vision). In the long term, we will rely on `SequentialRS` to combine both `ProtoMPRS` and `DistributedRS` together for users to choose either to enable MP + Dist or just MP or just Dist.

PR #908, opened 2022-11-25
Make HeaderIterDataPipe with limit=None a no-op

### LINKED ISSUE #762: Allow Header(limit=None) ?
Not urgent at all, just a minor suggestion:

In the benchmark scripts I'm currently running I want to limit the number of samples in a data-pipe according to an `args.limit` CLI parameter. I'd be nice to be able to just write:

```py
dp = Header(dp, limit=args.limit)
```

and let `Header` be a no-op when `limit=None`. This might be a bit niche, and the alternative is to just protect the call in a `if` block, so I would totally understand if this isn't in scope (and it's really not urgent in any case)

felix-newman:
Fixes #762

### Changes

- Passing limit=None makes the HeaderIterDataPipe a no-op

### DIFF
--- a/test/test_iterdatapipe.py
+++ b/test/test_iterdatapipe.py
@@ -344,6 +344,11 @@ class TestIterDataPipe(expecttest.TestCase):
         header_dp = source_dp.header(100)
         self.assertEqual(list(range(5)), list(header_dp))
 
+        # Functional Test: ensure the source is not modified if limit is set to None
+        source_dp = IterableWrapper(range(5))
+        header_dp = source_dp.header(None)
+        self.assertEqual(list(range(5)), list(header_dp))
+
         # Reset Test:
         source_dp = IterableWrapper(range(20))
         header_dp = Header(source_dp, 5)
@@ -360,6 +365,10 @@ class TestIterDataPipe(expecttest.TestCase):
         header_dp = source_dp.header(30)
         self.assertEqual(20, len(header_dp))
 
+        # __len__ Test: returns the length of source when limit is set to None
+        header_dp = source_dp.header(None)
+        self.assertEqual(20, len(header_dp))
+
         # __len__ Test: returns limit if source doesn't have length
         source_dp_NoLen = IDP_NoLen(list(range(20)))
         header_dp = source_dp_NoLen.header(30)
@@ -370,7 +379,13 @@ class TestIterDataPipe(expecttest.TestCase):
                 str(wa[0].message), r"length of this HeaderIterDataPipe is inferred to be equal to its limit"
             )
 
+        # __len__ Test: raises TypeError if source doesn't have length and limit is set to None
+        header_dp = source_dp_NoLen.header(None)
+        with self.assertRaises(TypeError):
+            len(header_dp)
+
         # __len__ Test: returns limit if source doesn't have length, even when it has been iterated through once
+        header_dp = source_dp_NoLen.header(30)
         for _ in header_dp:
             pass
         self.assertEqual(30, len(header_dp))
--- a/torchdata/datapipes/iter/util/header.py
+++ b/torchdata/datapipes/iter/util/header.py
@@ -4,7 +4,7 @@
 # This source code is licensed under the BSD-style license found in the
 # LICENSE file in the root directory of this source tree.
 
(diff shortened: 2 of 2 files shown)


test/test_iterdatapipe.py
```diff
@@ -344,6 +344,11 @@ def test_header_iterdatapipe(self) -> None:
         header_dp = source_dp.header(100)
         self.assertEqual(list(range(5)), list(header_dp))
 
+        # Functional Test: ensure the source is not modified if limit is set to None
```
mabu-github, line 347: To me it seems as if test_header_iterdatapipe() contains a whole test suite comprised of multiple tests. Imho there should be one test for each case and one (logical) assertion per test.

Maybe this can be improved in a separate PR so that we don't get scope creep here @ejguan?

test/test_iterdatapipe.py
```diff
@@ -370,7 +379,13 @@ def test_header_iterdatapipe(self) -> None:
                 str(wa[0].message), r"length of this HeaderIterDataPipe is inferred to be equal to its limit"
             )
 
+        # __len__ Test: raises TypeError if source doesn't have length and limit is set to None
+        header_dp = source_dp_NoLen.header(None)
+        with self.assertRaises(TypeError):
```
mabu-github, line 384: TypeError and ValueError are very general and theoretically if we didn't know the implementation, it could also happen because of different reasons. Therefore it is good practice to also check the error message to narrow down the error (e.g. using https://docs.python.org/3/library/unittest.html#unittest.TestCase.assertRaisesRegex).

torchdata/datapipes/iter/util/header.py
```diff
@@ -33,25 +33,29 @@ class HeaderIterDataPipe(IterDataPipe[T_co]):
...
         for value in self.source_datapipe:
             i += 1
-            if i <= self.limit:
+            if self.limit is None or i <= self.limit:
                 yield value
             else:
                 break
-        self.length = min(i, self.limit)  # We know length with certainty when we reach here
+        # We know length with certainty when we reach here
+        self.length = i if self.limit is None else min(i, self.limit)
```
mabu-github, line 49: This can probably be simplified to using just `i` without a min.  As incrementing is done before yielding or breaking, the length is `i-1`.

The following test case should be added: use a generator (has no len function) with some yields, drain it and assert self.length afterwards.

torchdata/datapipes/iter/util/header.py
```diff
@@ -33,25 +33,29 @@ class HeaderIterDataPipe(IterDataPipe[T_co]):
...
 
     def __len__(self) -> int:
         try:
             source_len = len(self.source_datapipe)
-            return min(source_len, self.limit)
+            return source_len if self.limit is None else min(source_len, self.limit)
         except TypeError:
+            if self.limit is None:
+                raise
```
mabu-github, line 57: Should we add another warning here @ejguan?

torchdata/datapipes/iter/util/header.py
ejguan: TBH, I think `self.length` should even be eliminated here for the reason that the `DataPipe` graph for each epoch might have different lengths as users can always alter the graph after one epoch ends.
We can simply rely on `limit` and `len(self.source_datapipe)` IMHO

test/test_iterdatapipe.py
ejguan: Sounds reasonable to me. And, I feel it's even better to add a directory and split this single test file into a few categories like `archive_datapipe.py`, `augmenting_datapipe.py`, etc.
DataPipe are roughly categorized in https://pytorch.org/data/beta/index.html

cc: @NivekT WDYT about the test suite

NivekT: Agrees with the concerns here. I think we should refactor all the DataPipe unit tests into a better format at some point... Open to suggestions but probably beyond the scope of this PR

torchdata/datapipes/iter/util/header.py
felix-newman: I think this would be to the user. Added it for the updated PR.

NivekT (approved):
LGTM! Thanks for contributing

Issue #865, opened 2022-10-27
Dataloader does not stop if one of the zipped DataPipes has a perpetual cycle


magehrig:
### 🐛 Describe the bug

The Zipper is supposed to stop as soon as the shortest input DataPipe is exhausted. This is not the case if one of the DataPipes is constructed with a perpetual cycle.

Minimal Example:
```Python
from torch.utils.data import DataLoader
from torchdata.datapipes.iter import IterableWrapper, IterDataPipe


class TestDataset(IterDataPipe):
    def __init__(self):
        self.data = IterableWrapper(list(range(10)))

    def __iter__(self):
        inf_cycle = IterableWrapper([42]).cycle(count=None)
        datapipe = self.data.zip(inf_cycle)
        return iter(datapipe)


if __name__ == '__main__':
    dataset = TestDataset()
    loader = DataLoader(dataset=dataset, batch_size=None, num_workers=0)

    for batch in loader:
        print(batch)
```
Prints the following but does not exit the program:
```bash
[0, 42]
[1, 42]
[2, 42]
[3, 42]
[4, 42]
[5, 42]
[6, 42]
[7, 42]
[8, 42]
[9, 42]
```

### Versions

```
PyTorch version: 1.14.0.dev20221026+cpu
Is debug build: False
CUDA used to build PyTorch: Could not collect
ROCM used to build PyTorch: N/A

OS: Ubuntu 18.04.6 LTS (x86_64)
GCC version: (Ubuntu 9.4.0-1ubuntu1~18.04) 9.4.0
Clang version: 8.0.0-3~ubuntu18.04.2 (tags/RELEASE_800/final)
CMake version: version 3.20.0
Libc version: glibc-2.27

Python version: 3.9.13 | packaged by conda-forge | (main, May 27 2022, 16:56:21)  [GCC 10.3.0] (64-bit runtime)
Python platform: Linux-4.15.0-194-generic-x86_64-with-glibc2.27
Is CUDA available: False
CUDA runtime version: 11.1.105
CUDA_MODULE_LOADING set to: N/A
GPU models and configuration:
GPU 0: Quadro RTX 8000
GPU 1: Quadro RTX 8000
GPU 2: Quadro RTX 8000
GPU 3: Quadro RTX 8000
GPU 4: Quadro RTX 8000
GPU 5: Quadro RTX 8000
GPU 6: Quadro RTX 8000
GPU 7: Quadro RTX 8000

Nvidia driver version: 515.65.01
[...]

PR #910, opened 2022-11-30
[3/4] Properly cleanup unclosed files within generator function

ejguan:
There is a case that `file.close` is never called because when generator function has never reached to the end. A simple example would be `zip` two datepipes with different length. The longer DataPipe would never reach the end of generator and then it will be cleaned up by `gc`. So, the line of `file.close` is not executed. (This is the reason that Vitaly has to create this [hack](https://github.com/pytorch/pytorch/blob/4451eb24e6287dff62ff8a7ec0eda6a6998807b0/torch/utils/data/datapipes/iter/combining.py#L573-L583) to retrieve all remaining data to make sure generator function is fully executed)

However, this hack introduces another problem where an infinite datapipe would make `zip` never end as it would try to deplete the infinite iterator. See: https://github.com/pytorch/data/issues/865

So, in this PR, I am adding a `try-finally` clause to make sure the `file.close` is always executed during the destruction of `generator` object. Then, we don't need the hack within `zip` any more.

- https://github.com/pytorch/pytorch/pull/89973
- https://github.com/pytorch/vision/pull/6997
- https://github.com/pytorch/pytorch/pull/89974

### DIFF
--- a/torchdata/datapipes/iter/util/combining.py
+++ b/torchdata/datapipes/iter/util/combining.py
@@ -115,13 +115,12 @@ class IterKeyZipperIterDataPipe(IterDataPipe[T_co]):
                 else:
                     yield res
         finally:
-            for remaining in ref_it:
-                janitor(remaining)
-
+            del ref_it
             # TODO(633): This should be Exception or warn when debug mode is enabled
-            if len(self.buffer) > 0:
+            if self.buffer:
                 for _, v in self.buffer.items():
                     janitor(v)
+                self.buffer.clear()
 
     def __len__(self) -> int:
         return len(self.source_datapipe)
@@ -156,7 +155,10 @@ class IterKeyZipperIterDataPipe(IterDataPipe[T_co]):
         self.buffer = OrderedDict()
 
     def __del__(self):
-        self.buffer.clear()
+        if self.buffer:
+            for _, v in self.buffer.items():
+                janitor(v)
+            self.buffer.clear()
 
 
 @functional_datapipe("zip_with_map")
--- a/torchdata/datapipes/iter/util/plain_text_reader.py
+++ b/torchdata/datapipes/iter/util/plain_text_reader.py
@@ -41,8 +41,10 @@ class PlainTextReaderHelper:
         with contextlib.suppress(StopIteration):
             for _ in range(self._skip_lines):
                 next(file)
-        yield from file
-        file.close()
+        try:
+            yield from file
+        finally:
+            file.close()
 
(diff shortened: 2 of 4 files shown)


torchdata/datapipes/iter/util/combining.py
```diff
@@ -115,13 +115,12 @@ def __iter__(self) -> Iterator:
                 else:
                     yield res
         finally:
-            for remaining in ref_it:
-                janitor(remaining)
-
+            del ref_it
```
NivekT, line 118: Question: IIUC, we are trusting there is nothing needs to be cleaned for `ref_it` or `ref_dp` will handle it?

NivekT (approved):
Just one question on this, but overall, this LGTM! 

Will look at the rest soon

torchdata/datapipes/iter/util/combining.py
wenleix: noob question: `del ref_it` will trigger `ref_it.__del__()` right? -- which seems to doing the cleaning? ~

torchdata/datapipes/iter/util/plain_text_reader.py
```diff
@@ -58,10 +60,9 @@ def strip_newline(self, stream: Union[Iterator[bytes], Iterator[str]]) -> Union[
     def decode(self, stream: Union[Iterator[bytes], Iterator[str]]) -> Union[Iterator[bytes], Iterator[str]]:
         if not self._decode:
             yield from stream
-            return
```
wenleix, line 61: like the new control flow. Originally doesn't understand why there is `return` after `yield`...

wenleix (approved):
Thanks!

torchdata/datapipes/iter/util/combining.py
ejguan: > IIUC, we are trusting there is nothing needs to be cleaned for `ref_it` or `ref_dp` will handle it?

> does that mean we believe `del ref_it` will do the clean up?

We have to. If the prior DataPipe is an infinite one like `cycle(None)`, it means we will never reach the end of `ref_it`. It's more like a safe guard to explicitly remove `ref_it` from the scope. 
And, this is the reason that all generators that yields opened file handles are required to using `finally` invoke `file.close` to guarantee the file is closed when `generator` object is out of scope.

torchdata/datapipes/iter/util/plain_text_reader.py
ejguan: Yeah. I feel like the current implementation is easier to understand.
Regarding `return` and `yield`, as long as there is a `yield` in the function, the function is automatically classified as generator function no matter if `return` exists or not.

PR #912, opened 2022-12-05
Add tests to valdiate iteration over combining DataPipe with infinite input

ejguan:
Add a few tests to validate the iterator is properly cleanup at the end of iteration, when the combining `DataPipe` takes input with infinite length like`cycle(None)`

Those tests should be timed out until https://github.com/pytorch/pytorch/pull/89974 is promoted to the nightly releases.


NivekT (approved):
LGTM!

PR #913, opened 2022-12-05
Fix CI to use stable numpy and networkx

ejguan:
Fixes torchvision broken CI caused by pre-release `numpy`.

In our CI, we install `torch` using `--pre` flag, which will make `pip` to use pre-release `numpy` and `networkx`. I have reported to TorchVision about the issue so they should be able to start to investigate how to prevent the problem when `numpy` promotes this release version to stable. For TorchData side, to mitigate the problem, the CI should use stable `numpy`.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -67,6 +67,7 @@ jobs:
       - name: Install dependencies
         run: |
           pip3 install -r requirements.txt
+          pip3 install networkx
           pip3 install --pre torch -f "${{ steps.pytorch_channel.outputs.value }}"
           pip3 install cmake ninja
           echo "/home/runner/.local/bin" >> $GITHUB_PATH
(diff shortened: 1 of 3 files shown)


pmeier:
See pytorch/vision##7009.

.github/workflows/ci.yml
```diff
@@ -67,6 +67,7 @@ jobs:
       - name: Install dependencies
         run: |
           pip3 install -r requirements.txt
+          pip3 install networkx
```
NivekT, line 70: Question: any reason to install here instead of within `test/requirements.txt`? Does installing from `requirements.txt` automatically install with `--pre`?

NivekT (approved):
Overall, looks good and makes sense to me. Thanks for looking into this!

Just one question in the comment.

.github/workflows/ci.yml
pmeier: `requirements.txt` does not contain `networkx`. This is a dependency coming from `torch`. And since we are installing that with the `--pre` flag for nightlies, we also pull in a pre-release of the other dependencies. By installing it explicitly before, the dependency is already satisfied and `pip` does not install the pre-release versions.

PR #916, opened 2022-12-06
Apply sharding based on priority & combine DistInfo and ExtraInfo

ejguan:
After https://github.com/pytorch/pytorch/pull/88424 is landed, we are able to invoke `apply_sharding` by sharding levels (distributed or multiprocessing). Then, we are able to give fine-control on sharding by `ReadingService`.
- For `DistributedReadingService`, we will only set sharding on the distributed level
- For `PrototypeMPReadingService`, we will set distributed sharding in the main process and set mp sharding in the worker processes. Previously, we set sharding in each worker process based on both distributed and mp information.
  - `worker_init_fn` doesn't need `DistInfo` anymore. As, the `DataPipe` has been distributed sharded in the main process.
  - Combine `DistInfo` and `ExtraInfo` for `worker_reset_fn` to synchronize the distributed seeds across distributed workers and set worker-local seeds based on both distributed and mp information.

### DIFF
--- a/docs/source/dataloader2.rst
+++ b/docs/source/dataloader2.rst
@@ -18,8 +18,7 @@ Note:
 - :class:`torchdata.datapipes.map.SequenceWrapper`: ``torch.utils.data.Dataset``
 - :class:`torchdata.datapipes.iter.IterableWrapper`: ``torch.utils.data.IterableDataset``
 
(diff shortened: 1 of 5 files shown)


NivekT:
I did not know it was possible to land without GH stamp.

ejguan:
@NivekT It's only possible for TorchData. PyTorch doesn't allow this anymore due to the GH first project

PR #917, opened 2022-12-08
Prevent reset iteration message sent twice to workers

ejguan:
Remove `end_datapipe.reset()` as it's going to be invoked at the beginning of iteration.
nit changes
- move `mp_ctx` creation out of loop
- Fix lint

### DIFF
--- a/torchdata/dataloader2/reading_service.py
+++ b/torchdata/dataloader2/reading_service.py
@@ -253,10 +253,11 @@ class PrototypeMultiProcessingReadingService(ReadingServiceInterface):
         if self.worker_prefetch_cnt > 0:
             datapipe = datapipe.prefetch(self.worker_prefetch_cnt)
 
+        ctx = mp.get_context(self.multiprocessing_context)
+
         for worker_id in range(self.num_workers):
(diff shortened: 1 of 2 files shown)


torchdata/dataloader2/reading_service.py
```diff
@@ -300,7 +301,6 @@ def initialize_iteration(self) -> None:
             dist_info = _DistInfo(shared_seed_int, self._world_size, self._rank)
             call_on_epoch_reset = partial(process_reset_fn, dist_info=dist_info, custom_reset_fn=self.worker_reset_fn)
             end_datapipe.reset_epoch(call_on_epoch_reset)
-            end_datapipe.reset()
```
NivekT, line 303: Do we need to call this when `self.main_prefetch_cnt <= 0`?

NivekT (approved):
LGTM! Just one question to make sure it is right.

torchdata/dataloader2/reading_service.py
ejguan: Not needed because `reset` is always called for each iteration of datapipe.

PR #919, opened 2022-12-09
[ReadingService] Add round robin sharding to support non-replicable DataPipe for Multiprocessing

ejguan:
This PR is created on top of https://github.com/pytorch/data/pull/555. And, this PR extends `PrototypeMultiprocessingReadingService` to accept non-replicable DataPipe.

And, this PR depends on https://github.com/pytorch/pytorch/pull/90769

### Main Changes
- Add a way to launch a process to fetch data from non-replicable DataPipes and send data to worker processes in a round-robin manner
  - Add `ShardingRoundRobinDispatcher` (functional name `sharding_round_robin_dispatch`) to indicate non-replicable DataPipe
  - Add `MultipleDataPipesToQueuesLoop` to connect non-sharding process to request/response queues
  - Add `find_lca_non_replicable_dp` as a graph function to determine the lowest common ancestor of all non-replicabble DataPipes. This would guarantee that all non-replicable DataPipes will be running in a single dispatching process
  - In each multiprocessing worker process,
    - If All datapipes are replicable, apply multiprocessing sharding to the graph
    - If not, worker would use `find_replicable_branches` to apply mp sharding to those replicable branches, because all non-replicable branches have been properly sharded by routing data round-robinly to worker processes.
- Properly get `ResetEpochResponse` from protocol via `get_response_reset_epoch`
- [x] Add tests for two graph functions
- [x] Add test to launch non-shardable DataPipe process
- Add documents
  - [x] Replicable DataPipe/Non-replicable DataPipe in multiprocessing
  - [x] How PrototypeMPRS handles the above two types of DataPipe
Please check the link for doc: https://ejguan.github.io/dataloader2.html#dynamic-sharding
## nit Changes
- Rename `Spawn` to `Create` as the process has not been started


test/dataloader2/test_dataloader2.py
```diff
@@ -302,7 +302,7 @@ class TestDataLoader2EventLoop(TestCase):
     #
     #     it = list(range(100))
     #     numbers_dp = IterableWrapper(it)
-    #     (process, req_queue, res_queue, _thread_local_datapipe) = communication.eventloop.SpawnThreadForDataPipeline(numbers_dp)
+    #     (process, req_queue, res_queue, _thread_local_datapipe) = communication.eventloop.CreateThreadForDataPipeline(numbers_dp)
```
ejguan, line 305: I am fine to revert this nit change as well.

torchdata/dataloader2/communication/iter.py
```diff
@@ -208,3 +213,66 @@ def nonblocking_next(self):
         if isinstance(response, communication.messages.InvalidStateResponse):
             raise NotAvailable
         return response.value
+
+
+class _IterateQueueDataPipes(IterDataPipe):
```
ejguan, line 218: Moving here to prevent circular import

torchdata/dataloader2/communication/protocol.py
```diff
@@ -225,6 +225,16 @@ def get_response_reset_iterator(self, block=False):
         if not isinstance(response, communication.messages.ResetIteratorResponse):
             raise Exception("Invalid response received")
 
+    def get_response_reset_epoch(self, block=False):
```
ejguan, line 228: Not sure why we didn't have this response before

torchdata/dataloader2/communication/eventloop.py
```diff
@@ -26,16 +29,62 @@
...
+        res_queues
+    ), "``MultipleDataPipesToQueuesLoop`` requires the same number of datapipes, request queues and response queues"
+
+    torch.set_num_threads(1)
+
+    loops = []
+
+    for source_datapipe, req_queue, res_queue in zip(source_datapipes, req_queues, res_queues):
+        # TODO(ejguan): non-blocking request/response introduces non-determinism
+        # because the order of operations is not going to be consistent in this process
```
ejguan, line 57: It's actually deterministic because `round_robin_demux` would make sure the order of fetching is preserved.

NivekT (commented):
I am going to need more time to look through this. What is the use case of non-shardable DataPipe?

NivekT (commented):
I think we should provide a definition of non-shardable DataPipe and why that may occur (an abstract example would be helpful as well). In particularly, as you mentioned, note that we do not want to duplicate such DataPipe into multiple workers and sharding filter should not be applied to it. Instead it should be read round-robin (or something else) by downstream DataPipes that are sharded?

ejguan:
> I think we should provide a definition of non-shardable DataPipe and why that may occur (an abstract example would be helpful as well).

Make sense. I will add doc regarding non-shardbable DataPipe/shardable DataPipe to the documents for dataloader2.

> In particularly, as you mentioned, note that we do not want to duplicate such DataPipe into multiple workers and sharding filter should not be applied to it. Instead it should be read round-robin (or something else) by downstream DataPipes that are sharded?

Actually, this is not True. `sharding_filter` is still needed if distributed sharding is needed. We only disabled multiprocessing sharding for the non-shardable branch. And, round-robin reading is also achieved automatically with `ProtoTypeMPRS`.

torchdata/dataloader2/communication/protocol.py
NivekT: Non-blocking but - hmmmm.... did this cause any bug or unhandled messages? Do you happen to know why?
I know you were looking into unusually messages and responses

torchdata/dataloader2/utils/worker.py
```diff
@@ -84,9 +153,22 @@ def process_reset_fn(
     reset the random state of the ``DataPipe`` graph and the global random states for ``torch``,
     ``random`` and ``numpy``.
     """
+    # Reset non-sharding process first
+    graph = traverse_dps(datapipe)
+    non_sharding_process_dps = find_dps(graph, communication.iter._IterateQueueDataPipes)
```
NivekT, line 158: Question: Is it always the case where `_IterateQueueDataPipes` is the only non-sharding process?

torchdata/dataloader2/communication/protocol.py
ejguan: When new iteration starts, `reset` will be called for `_IterateQueueDataPipes `. And, an extra `get_respondse_next` is invoked to drop / response. So, all requests are served.
https://github.com/pytorch/data/blob/0a0ae5d35be439c786c9ab798ea1596238d8640a/torchdata/dataloader2/reading_service.py#L152-L154

I can give a try to remove this part.

torchdata/dataloader2/utils/worker.py
ejguan: Yes because we will find the lowest common ancestor of all non-shardable DataPipes in the main process and replace it by this `_IterateQueueDataPipes` in the worker process.
So, it's guaranteed that there is only a single non-sharding process.

torchdata/dataloader2/communication/protocol.py
ejguan: It seems removing is fine

NivekT:
That makes sense. How do you plan to implement that? Will we still have users calling `.fullsync()` and make modification to the graph in DL2? Or making `full_sync` an option within DL2? Probably the former?

ejguan:
> How do you plan to implement that? Will we still have users calling `.fullsync()` and make modification to the graph in DL2?

We should allow either users calling `fullsync` or let `DistributedRS` attach `fullsync` automatically. Then, `ProtoMPRS` should do graph modification. WDYT? I am not going to implement it in this PR for `fullsync` though.

wenleix:
Replying to https://github.com/pytorch/data/pull/919?notification_referrer_id=NT_kwDOAAwycrE1MDQ4ODIwNTc5Ojc5OTM0Ng&notifications_query=reason%3Areview-requested#issuecomment-1351871160 : 

>  Actually, this is not True. sharding_filter is still needed if distributed sharding is needed. We only disabled multiprocessing sharding for the non-shardable branch. And, round-robin reading is also achieved automatically with ProtoTypeMPRS.

An alternative approach for distributed sharding would be distribute the workload based on filename or some compression/encoding unit in file (in Parquet it's called "Page": https://parquet.apache.org/docs/concepts/) and in ORC I think it's called "Stripe". So it avoid reading the original data multiple times?

torchdata/dataloader2/communication/protocol.py
wenleix: noob question: any docs / entry pointer to understand how these `dataloader2/communication` design / work? ~

test/dataloader2/test_dataloader2.py
```diff
@@ -352,6 +352,21 @@ def clean_me(process, req_queue, res_queue):
         clean_me(process, req_queue, res_queue)
 
 
+class NonShardableDataPipe(IterDataPipe):
```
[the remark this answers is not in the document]
ejguan: > BTW why do we want to use this sharding as default? seems so expensive (e.g. reading/decompressing data times and discarding most of them?...)

I think it's more OSS use case where users want to provide a list of file paths. When MP takes place, with the explicit `sharding_filter`, we only do read/decompress partial dataset. This requires users to place `sharding_filter` to the right place.

> I am wondering if we should give more specific name? Such as "filter_shard" vs. "roundrobin_shard" 

I see. We can let users to define `roundrobin_shard` explicitly (alternative of `sharding_filter`). Then, we find lowest common ancestor  of those `roundrobin_shard` in the graph. This is actually a great idea! Thank you.
Then, we don't necessarily bother to define the non-shardable DataPipe. As long as `roundrobin_shard` is presented, we would treat the prior graph as the non-sharding DataPipes.

torchdata/dataloader2/communication/protocol.py
ejguan: Unfortunately no doc. I can talk about the components based on my understanding:
- [ProtocolClient](https://github.com/pytorch/data/blob/main/torchdata/dataloader2/communication/protocol.py#L196) remains in the main process to pass `Request` via the `request_queue` to the corresponding worker process
- [ProtocolServer](https://github.com/pytorch/data/blob/56ed97e48da9934ce23453fc7b04418b64b02e8a/torchdata/dataloader2/communication/protocol.py#L168) is created in the worker process  that takes request then send `Response` back to main process  via `reqponse_queue`
- [DataPipeBehindQueues](https://github.com/pytorch/data/blob/main/torchdata/dataloader2/communication/iter.py#LL109C5-L109C25) is the worker loop that holds a ProtocolServer to maniputlate `DataPipe` based on the `Reqeust`
- [QueueWrapper](https://github.com/pytorch/data/blob/main/torchdata/dataloader2/communication/iter.py#L170) is the `DataPipe` that holds a `ProtocolClient` instance to issue `Request` and yield data from `response_queue` to the subsequent `DataPipe` graph.

We can talk about more detail offline if you want

test/dataloader2/test_dataloader2.py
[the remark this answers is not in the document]
ejguan: There is one more thing that needs to carefully design. When `sharding_filter` is presented, we will do distributed sharding first then do mp sharding. However, it becomes more complicated for `sharding_roundrobin`. There are two (three) different scenarios:
- If we don't want to do distributed sharding when the backend service has already done that, it's quite straight forward to launch one non-sharding process per distributed node. Proposed API: `sharding_roundrobin(sharding_group_filter=SHARDING_PRIORITIES.MULTIPROCESSING)`
- It becomes more complicated if we want to do distributed sharding as well, because distributed sharding can be achieved by two different ways.
  - Do distributed sharding like a `filter` (emit `world_size-1/world_size`) and run a non-sharding process to dispatch data from this process to worker processes on the same node. Proposed API:  `sharding_filter(sharding_group_filter=SHARDING_PRIORITIES.DISTRIBUTED).sharding_roundrobin(sharding_group_filter=SHARDING_PRIORITIES.MULTIPROCESSING)`.
  - Do a global round-robin distributed sharding. In that case, we can launch one process on rank 0 and distribute data to other distributed workers. Proposed API:  `sharding_roundrobin()` equivalent to `sharding_roundrobin(sharding_group_filter=(SHARDING_PRIORITIES.DISTRIBUTED, SHARDING_PRIORITIES.MULTIPROCESSING))`

This requires some changes to [`apply_sharding`](https://github.com/pytorch/pytorch/blob/d6efd25d1e51337f94229036825810dc09c80824/torch/utils/data/graph_settings.py#L35-L61) to accept multiple sharding in a branch, as long as they have exclusive `sharding_priorities`.

ejguan:
Non-shardable is an extremely bad name as it's actually being sharded by round-robin dispatching. The actual meaning here is to prevent copy of DataPipe to multiple processes. I might rename it to non-replicable DataPipe/coordinating process.

NivekT:
> Non-shardable is an extremely bad name as it's actually being sharded by round-robin dispatching. The actual meaning here is to prevent copy of DataPipe to multiple processes. I might rename it to non-replicable DataPipe/dispatching process.

I agree with renaming "non-shardable" to "non-replicated" DataPipe. I suppose sometimes it is replicable but the users won't want to?

ejguan:
@wenleix @NivekT This PR has been updated. And updated document can be found in https://ejguan.github.io/dataloader2.html#dynamic-sharding

torchdata/dataloader2/communication/protocol.py
wenleix: Thanks! Something like this? ~

319563983_491898436260282_330096271851500173_n

ejguan: Correct.

torchdata/dataloader2/utils/worker.py
```diff
@@ -73,6 +103,45 @@ def process_init_fn(
...
+    worker_seed_generator.manual_seed(dist_info.shared_seed)
+    torch.utils.data.graph_settings.apply_random_seed(
+        datapipe,
+        worker_seed_generator,
+    )
+
+    # Set global random states
+    _set_global_random_state(worker_seed_generator)
+
+    return datapipe
```
ejguan, line 142: Review Step 9: When new epoch starts, we want to control the random seed based on distributed information.
We need to guarantee all distributed dispatching process share the same random seed.

ejguan (commented):
The following review steps are added to make it easier to review the main logic in the PR. Let me know if there is anything fuzzy to you.

wenleix (commented):
"Review Step 1: Add ShardingRoundRobinDispatcher is introduced to indicate where the pipeline should be non-replicable."

LGTM. Name (`sharding_round_robin_dispatch`) is a bit long but let's keep it for now...

torchdata/datapipes/iter/util/sharding.py
```diff
@@ -0,0 +1,45 @@
...
+        self.source_datapipe = source_datapipe
+        if sharding_group_filter != SHARDING_PRIORITIES.MULTIPROCESSING:
+            raise NotImplementedError(
+                "`sharding_round_robin_dispatch` currently only supports `SHARDING_PRIORITIES.MULTIPROCESSING`."
+                "Please open issue on github for your feature request."
+            )
+        self.sharding_group_filter = sharding_group_filter
+
+    def __iter__(self) -> Iterator[T_co]:
+        yield from self.source_datapipe
```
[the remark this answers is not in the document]
wenleix: by "single-process use case". does it mean "eager mode"?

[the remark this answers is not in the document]
ejguan: Yeah, pure eager currently. In the future, we might provide a by-default `SingleProcessReadingService` for users.

torchdata/dataloader2/utils/dispatch.py
```diff
@@ -0,0 +1,122 @@
...
+        non_replicable_parents = []
+        for dp_id, (dp, src_graph) in root_graph.items():
+            res = helper(dp_id, dp, src_graph)
+            if res is not None:
+                non_replicable_parents.append(res)
+        # `root_dp` becomes the lowest common ancestor of this branch,
+        # if there are more than one unique non-replicable DataPipe prior to it.
+        if len(non_replicable_parents) > 0:
+            # One unique non-replicable DataPipe
+            if len(non_replicable_parents) == 1 or all(
```
wenleix, line 76: curious: can we just do something like 
```python
if len(set(non_replicable_parents)) == 1:
    ...
```
?

torchdata/dataloader2/utils/dispatch.py
```diff
@@ -0,0 +1,122 @@
...
+            continue
+        if _is_round_robin_sharding(dp):
+            parent_dps = list_dps(traverse_dps(dp))
+            for par_dp in parent_dps:
+                non_replicable_dps.add(id(par_dp))
+
+    root_dp_id = list(graph.keys())[0]
+    root_dp, root_graph = graph[root_dp_id]
+
+    cache: Dict[int, Optional[DataPipe]] = {}
```
wenleix, line 58: nit: what about call this `lca_for_subgraph`?

wenleix (commented):
"Review Step 2: Add this graph function to find the lowest common ancestor of the non-replicable DataPipes (ShardingRoundRobinDispatcher)"

LGTM. A seperate question is why we need to find LCA for non-repliate DataPipes (not aware of this in other dataflow systems before)

torchdata/dataloader2/utils/dispatch.py
```diff
@@ -0,0 +1,122 @@
...
+from torchdata.datapipes.iter import IterDataPipe, ShardingRoundRobinDispatcher
+
+
+__all__ = ["_DummyIterDataPipe", "find_lca_non_replicable_dp", "find_replicable_branches"]
+
+
+class _DummyIterDataPipe(IterDataPipe):
+    r"""
+    This DataPipe is a placeholder to be replaced by the ``QueueWrapper``
+    that connects the worker process for non-replicable DataPipe.
```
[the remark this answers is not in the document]
ejguan: No, this one connect dispatching process (containing non-replicable DataPipe) to worker process.
Let's say we have a graph as:
```mermaid
graph TD;
    DP0-->A["non-replicable DP"]-->DP1-->DP2-->DP3;
```
In the main process, the graph breaks into two pieces. One is the non-replicable graph, the another one is replicable graph.
```mermaid
graph TD;
    DP0-->A["non-replicable DP"];
    _DummyIterDataPipe-->DP1-->DP22-->DP3;
```
The non-replicable graph is sent to dispatching process and connect it with `num_workers` pairs of req_queue and res_queue. Then, each pair of queues will be connected to `_DummyIterDataPipe` before sent to worker process. Then, in each worker process, `_DummyIterDataPipe` will be replaced by `QueueWrapper` with ProtocolClient.
```mermaid
graph TD;
    ProtocolClient_0-->A0["DP1(worker_0)"]-->B0["DP2(worker_0)"]-->C0["DP3(worker_0)"];
    ProtocolClient_1-->A1["DP1(worker_1)"]-->B1["DP2(worker_1)"]-->C1["DP3(worker_1)"];
```

torchdata/dataloader2/utils/dispatch.py
ejguan: There can be case that multiple branches share the same non-replicable DataPipe. Then, the LCA should not be the new joined DataPipe. Example:
```mermaid
graph TD;
    DP0["non-replicable DP0"]-->DP1-->DP3-->DP4;
    DP0-->DP2-->DP4;
```
For `DP4`, `len(non_replicable_parents)==2` but both `non_replicable_parents` are the same `DP0`. Then, the LCA should be `DP0` rather than `DP4`.

ejguan:
> A seperate question is why we need to find LCA for non-replicate DataPipes (not aware of this in other dataflow systems before)

Using the sample graph in https://ejguan.github.io/dataloader2.html#dynamic-sharding. If there are multiple `sharding_round_robin_dispatch`, there are two ways to merge them:
- Use LCA
- Launch one dispatching process per non-replicable DataPipe

I choose to use LCA to reduce the complexity of multiple queues and processes communication. Let's talk more detail offline.

torchdata/dataloader2/utils/dispatch.py
wenleix: but in that case won't `all(dp == non_replicable_parents[0] for dp in non_replicable_parents)` also returns true?

torchdata/dataloader2/communication/protocol.py
wenleix: Let me try to draw it with mermaid

```mermaid
graph TD;
    Worker1_1-->ProtocolServer_1_1-->ProtocolClient_1
    Worker1_2-->ProtocolServer_1_2-->ProtocolClient_1
    Worker1_3-->ProtocolServer_1_3-->ProtocolClient_1
    ProtocolClient_1-->GPU1
    ProtocolClient_2-->GPU2
    ProtocolClient_3-->GPU3
```

wenleix:
> I choose to use LCA to reduce the complexity of multiple queues and processes communication. Let's talk more detail offline.

Makes sense. This helps keep our distribute data loading system (relatively) simple with reasonable perf~  . We can revisit the design decision upon use case~

test/test_graph.py
```diff
@@ -250,6 +258,130 @@ def test_multiprocessing_reading_service(self) -> None:
...
+        # Multi-branch
+        multi_br_dp = IterableWrapper(list(range(10)))
+        ch1, ch2 = multi_br_dp.fork(2)
+        fork_zip_dp = ch1.zip(ch2)
+
+        # Circular-branch
+        cir_br_dp = IterableWrapper(list(range(10)))
+        cir_map_dp = cir_br_dp.map(_x_mult_2)
+        # Force to circular reference
+        cir_br_dp.cir_dep = cir_map_dp
```
wenleix, line 299: noob question: where does the attribute `cir_dep` come from. ?

wenleix (commented):
Review Step 3: single/multiple non-replicable DataPipes: LGTM.

torchdata/dataloader2/utils/dispatch.py
wenleix: discussed offline. `DataPipe` can be unhashable so cannot use `set`

torchdata/dataloader2/utils/dispatch.py
```diff
@@ -0,0 +1,122 @@
...
+
+    return helper(root_dp_id, root_dp, root_graph)
+
+
+def find_replicable_branches(graph: DataPipeGraph) -> List[DataPipe]:
+    r"""
+    Given the graph of DataPipe generated by ``traverse_dps`` function, return the DataPipe
+    instances that don't have ``_DummyIterDataPipe`` (non-replicable DataPipe) in the prior graph.
+    """
+    assert len(graph) == 1, "DataPipeGraph should only contain a single output DataPipe"
```
wenleix, line 93: does that mean there should be only 1 connected component in the data pipe graph? curious when will there be more than 1 connected components?

torchdata/dataloader2/utils/dispatch.py
```diff
@@ -0,0 +1,122 @@
...
+
+    dps: List[DataPipe] = []
+    cache: Dict[int, bool] = {}
```
wenleix, line 96: i see `cache` used in other place such as `traverse` so didn't suggest any change. but curious why it's not called something like `cached_subgraph_is_replicable`? :)

torchdata/dataloader2/utils/dispatch.py
```diff
@@ -0,0 +1,122 @@
...
+            parent_dps = list_dps(traverse_dps(dp))
+            for par_dp in parent_dps:
+                non_replicable_dps.add(id(par_dp))
+
+    root_dp_id = list(graph.keys())[0]
+    root_dp, root_graph = graph[root_dp_id]
+
+    cache: Dict[int, Optional[DataPipe]] = {}
+
+    def helper(root_dp_id, root_dp, root_graph) -> Optional[DataPipe]:  # pyre-ignore
```
wenleix, line 60: also consider name this method as something like `get_lca_for_subgraph` .

torchdata/dataloader2/utils/dispatch.py
```diff
@@ -0,0 +1,122 @@
...
+    """
+    assert len(graph) == 1, "DataPipeGraph should only contain a single output DataPipe"
+
+    dps: List[DataPipe] = []
+    cache: Dict[int, bool] = {}
+
+    root_dp_id = list(graph.keys())[0]
+    root_dp, root_graph = graph[root_dp_id]
+
+    def helper(root_dp_id, root_dp, root_graph) -> bool:  # pyre-ignore
```
wenleix, line 101: nit: call it like `subgraph_is_replicable`? since doesn't seem immediately obvious what returned boolean indicates~

wenleix (commented):
"Review Step 4: Add a graph function to find replicable branches from the DataPipe graph in the worker process."

LGTM. Initially didn't understand why children dp is only added if root_dp is not replicable. But now it seems make sense to me (e.g. we don't want all the DataPipe node but only the "output DataPipe").

wenleix (commented):
"Review Step 5: Tests to find replicable branches. " % 1 minor comment.

torchdata/dataloader2/reading_service.py
```diff
@@ -187,14 +137,22 @@ class PrototypeMultiProcessingReadingService(ReadingServiceInterface):
             of each epoch in each worker process with ``WorkerInfo``
             and ``DataPipe`` as the expected arguments.
 
+    Note:
+        - This ``ReadingService`` is still in prototype mode and will replace
```
wenleix, line 141: unrelated: we should start to graduate it from "prototype mode" and starts find initial pioneering use case adoption~

torchdata/dataloader2/reading_service.py
```diff
@@ -187,14 +137,22 @@ class PrototypeMultiProcessingReadingService(ReadingServiceInterface):
...
+          :class:`MultiProcessingReadingService`.
+        - It currently does both distributed and multiprocessing sharding over the pipeline.
+          The distributed-related code is going to be removed when ``SequentialReadingService``
```
wenleix, line 144: any doc about `SequentialReadingService` and `DistributedReadingService`? :)

wenleix (commented):
"Review Step 6: Split original data pipeline by using the LCA of non-replicable DataPipe node and sending the first part to "dispatch process"": LGTM.

torchdata/dataloader2/utils/dispatch.py
ejguan: This is the end of `DataPipe` graph and we always expect only one output. There can be multiple connected within the graph.

The `graph` returned by `traverse_dps` is in the reversed order. Using the following example:
```mermaid
graph TD;
    A-->B-->C-->D;
    A-->C;
    E-->F-->D;
    D-->output;
```
The returned `graph` is:
```
{id(D): (D, {
    id(C): (C, {
            id(B): (B: {
                id(A}: (A, {})
            }),
            id(A): (A, {}),
        }),
    id(F): (F, {
        id(E): (E, {}),
    }),
})}
```
So, the length for first level of `graph` should always be 1

torchdata/dataloader2/utils/dispatch.py
ejguan: I will change those `cache` name to more meaning ones.

test/test_graph.py
```diff
@@ -250,6 +258,130 @@ def test_multiprocessing_reading_service(self) -> None:
...
+
+    def test_replicable_branches(self):
+        r"""
+        There should be a single DataPipe as the lowest common ancestor of all
+        non-replicable DataPipes that is replaced by ``DummyIterDataPipe``.
+        """
+        single_br_dp, *_, fork_zip_dp, _, cir_map_dp, _, graph = self._make_dp()
+        graph = replace_by_dummy(graph, single_br_dp)
+        dps = find_replicable_branches(graph)
+        self.assertTrue(all(dp in (fork_zip_dp, cir_map_dp) for dp in dps))
```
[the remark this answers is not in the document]
ejguan: Sounds reasonable to me.

torchdata/dataloader2/reading_service.py
ejguan: Yea, let me open an issue to track this

torchdata/dataloader2/reading_service.py
ejguan: Sequential: https://fburl.com/gdoc/bwmue33j
Distributed: https://pytorch.org/data/beta/generated/torchdata.dataloader2.DistributedReadingService.html#torchdata.dataloader2.DistributedReadingService
And, distributed is quite simple by doing two things:
- Add fullsync
- Launch distributed Process Group to share distributed random seeds

torchdata/dataloader2/communication/eventloop.py
```diff
@@ -26,16 +29,57 @@
...
+        loops.append(
+            _create_datapipe_queue_loop(source_datapipe, req_queue, res_queue, blocking_request_get=False)
+        )  # Non-blocking request
+
+    # Using `zip_longest` to guarantee the process is terminated only when
+    # all loops have received `TerminateRequest`
+    for _ in zip_longest(*loops):
+        # time.sleep to make Python switch context to get/send message in mp.Queue
+        # TODO(ejguan): Microbenchmarked a synthetic non-shardable case that sleep perform similar to pass.
+        #               A more comprehensive benchmarking in read-world scneario is needed.
```
wenleix, line 62: nit: real world? 😄

torchdata/dataloader2/communication/eventloop.py
```diff
@@ -26,16 +29,57 @@
...
+    for source_datapipe, req_queue, res_queue in zip(source_datapipes, req_queues, res_queues):
+        loops.append(
+            _create_datapipe_queue_loop(source_datapipe, req_queue, res_queue, blocking_request_get=False)
+        )  # Non-blocking request
+
+    # Using `zip_longest` to guarantee the process is terminated only when
+    # all loops have received `TerminateRequest`
+    for _ in zip_longest(*loops):
+        # time.sleep to make Python switch context to get/send message in mp.Queue
+        # TODO(ejguan): Microbenchmarked a synthetic non-shardable case that sleep perform similar to pass.
```
wenleix, line 61: nit: non-replicable

torchdata/dataloader2/communication/eventloop.py
```diff
@@ -26,16 +29,57 @@
...
+    r"""
+    Set the appropriate pipes and protocol server type, and create a loop over multiple datapipes
+    with the protocol server in a non-blocking manner.
+    """
+    assert call_on_process_init is None, "``MultipleDataPipesToQueuesLoop`` does not support call_on_process_init"
+    assert len(source_datapipes) == len(req_queues) and len(req_queues) == len(
+        res_queues
+    ), "``MultipleDataPipesToQueuesLoop`` requires the same number of datapipes, request queues and response queues"
+
+    torch.set_num_threads(1)
```
wenleix, line 48: noob question: what's this for?

wenleix (commented):
"Review Step 7: create req_queue and res_queueand execute with round_robin_demux"

High-level logic looks good. But didn't have enough low-level context yet~

torchdata/dataloader2/utils/worker.py
```diff
@@ -62,9 +65,36 @@ def process_init_fn(
...
+                dp, worker_info.num_workers, worker_info.worker_id, SHARDING_PRIORITIES.MULTIPROCESSING
+            )
+
+        req_queue = non_replicable_dp[0].req_queue
+        res_queue = non_replicable_dp[0].res_queue
+
+        queue_wrapper = communication.iter.QueueWrapper(
+            communication.protocol.IterDataPipeQueueProtocolClient(req_queue, res_queue)
+        )
+        dispatch_process_dp = communication.iter._IterateQueueDataPipes([queue_wrapper])
```
wenleix, line 95: IIUC: 

`IterDataPipeQueueProtocolClient` will be wrapped into a `QueueWrapper` (but still not a `IterDataPipe`), and further wrapped into a `_IterateQueueDataPipes` which is a `IterDataPIpe`?

wenleix (commented):
"Review Step 8: Worker process handling (graph rewrite and receiving demux result from dispatch process)"

LGTM.

torchdata/dataloader2/utils/worker.py
ejguan: Nope. Both `QueueWrapper` and `_IterateQueueDataPipes` are `IterDataPipe`, this is one of the thing that we can optimize later.

torchdata/dataloader2/utils/worker.py
```diff
@@ -73,6 +103,45 @@ def process_init_fn(
...
+
+def dispatch_process_reset_fn(
+    datapipe: DataPipe,
+    worker_info: WorkerInfo,
+    dist_info: _DistInfo,
+) -> DataPipe:
+    r"""
+    Based on the distributed shared random seed, this function is used to set the random state
+    of the non-repliable ``DataPipe`` graph and the global random states for the dispatch process.
+    This function would guarantee that all distributed non-sharding process share the
```
wenleix, line 129: nit: non-replicable ?

torchdata/dataloader2/utils/worker.py
```diff
@@ -84,9 +153,22 @@ def process_reset_fn(
     reset the random state of the ``DataPipe`` graph and the global random states for ``torch``,
     ``random`` and ``numpy``.
     """
+    # Reset non-sharding process first
+    graph = traverse_dps(datapipe)
+    dispatch_process_dps = find_dps(graph, communication.iter._IterateQueueDataPipes)
```
wenleix, line 158: so... this is is on worker process? -- the `_IterateQueueDataPipes` is really the "source data pipe" that receives data from "dispatch process" right? 

In the dispatch process, the "sink data pipe" is actually the round_robin_demux?

wenleix (commented):
Review Step 9: LGTM % minor question...

torchdata/dataloader2/communication/eventloop.py
ejguan: Lol. I mean the OSS use case. I will reword it

torchdata/dataloader2/communication/eventloop.py
ejguan: IIRC, this was introduced to disable `OpenMP` in dataloader workers. This is because OpenMP would create number of threads that equals to the number of CPU cores by default. And, with multiprocessing enabled, `num_workers x num_threads_per_worker` threads will be created. This won't provide any further benefit.

Besides, `OpenMP` features should not be enabled if any `OpenMP` features are utilized in the main process and before subprocesses are forked.

Any suggestion?

torchdata/dataloader2/utils/worker.py
ejguan: Correct!

wenleix: in that case shall we call the variable name as `dispatch_process_dps`? -- since it's a "source/receiving data pipe" that receives data on worker process? shall we call it something like `worker_source_dps`(don't have a good name suggestion :( )

wenleix (approved):

torchdata/dataloader2/utils/worker.py
ejguan: `dispatch_consumer_dps`?

torchdata/dataloader2/communication/iter.py
```diff
@@ -208,3 +213,62 @@ def nonblocking_next(self):
...
+                        continue
+                    if isinstance(response, communication.messages.InvalidStateResponse):
+                        raise communication.iter.InvalidStateResetRequired
+                    if isinstance(response, communication.messages.TerminateResponse):
+                        raise communication.iter.TerminateRequired
+                    self.datapipes[idx].protocol.request_next()
+                    yield response.value
+
+    def reset(self):
+        # NonBlocking DataPipes do not reset automatically, have to do it manually
```
NivekT, line 257: @ejguan Just noticed that the `reset` method has changed after the move. It used to have this:
```python
# Collect all existing requests results to clear queues
        for dp in self.datapipes:
            if dp.protocol.waiting_for_response():
                dp.protocol.get_response_next(block=True)
```

Is this no longer necessary?

ejguan: I don't think it's necessary because we always want to do `reset_epoch` as well for `NonBlocking`, which will discard all existing requests. So, when at the point of `reset`, we should expect no request within the worker process queues.

PR #920, opened 2022-12-12
Attemp to fix flaky [Connection refused] failure for elastic test

ejguan:
Add `dl.shutdown` to destroy the distributed process group. This potentially fixes the problem that the port is taken up by other process group.

The other failures in our nightly release are caused by PyTorch's dependency issue (`sympy` is missing from conda dependency)


NivekT (approved):
LGTM!

ejguan:
This PR does fix the flaky tests. So, we should recommend users to always call `shutdown` after finished iteration. I will send a PR to add `shutdown` to our test cases.

PR #921, opened 2022-12-13
Added Linux Workflow

DanilBaibak:
Added Linux Wheels workflow. The workflow will run in shadow mode without upload binaries.

### Changes

- Added Linux Wheels workflow
- Added Linux Conda workflow

### TODO

- [x] Add the `BUILD_S3` env variable. Depend on the PR https://github.com/pytorch/test-infra/pull/1269
- [x] Fixed smoke tests issue for Linux Conda - https://github.com/pytorch/test-infra/pull/1292
- [x] Add smoke test - https://github.com/pytorch/data/pull/923

### DIFF
--- a/.github/workflows/build_conda_linux.yml
+++ b/.github/workflows/build_conda_linux.yml
@@ -25,6 +25,7 @@ jobs:
           - repository: pytorch/data
             pre-script: ""
             post-script: ""
+            conda-package-directory: packaging/torchdata
             smoke-test-script: ""
             package-name: data
     name: ${{ matrix.repository }}
--- a/.github/workflows/build_conda_linux.yml
+++ b/.github/workflows/build_conda_linux.yml
@@ -26,7 +26,7 @@ jobs:
             pre-script: ""
             post-script: ""
             conda-package-directory: packaging/torchdata
-            smoke-test-script: ""
+            smoke-test-script: test/smoke_test/smoke_test.py
             package-name: torchdata
     name: ${{ matrix.repository }}
     uses: pytorch/test-infra/.github/workflows/build_conda_linux.yml@main
--- a/.github/workflows/build_conda_linux.yml
+++ b/.github/workflows/build_conda_linux.yml
@@ -42,8 +42,7 @@ jobs:
       smoke-test-script: ${{ matrix.smoke-test-script }}
       package-name: ${{ matrix.package-name }}
       env-var-script: packaging/env-var-script.txt
-      # Using "development" as trigger event so these binaries are not uploaded
-      # to official channels yet
+      # Using "development" as trigger event so these binaries are not uploaded to official channels yet
       trigger-event: development
     secrets:
       CONDA_PYTORCHBOT_TOKEN: ${{ secrets.CONDA_PYTORCHBOT_TOKEN }}
--- a/.github/workflows/build_conda_linux.yml
+++ b/.github/workflows/build_conda_linux.yml
@@ -29,18 +29,19 @@ jobs:
             smoke-test-script: test/smoke_test/smoke_test.py
             package-name: torchdata
     name: ${{ matrix.repository }}
-    uses: pytorch/test-infra/.github/workflows/build_conda_linux.yml@main
+    uses: pytorch/test-infra/.github/workflows/build_conda_linux.yml@Env-Var-Config
     with:
       conda-package-directory: ${{ matrix.conda-package-directory }}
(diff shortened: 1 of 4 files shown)


.github/workflows/build_conda_linux.yml
```diff
@@ -0,0 +1,48 @@
+name: Build Linux Conda
+
+on:
+  pull_request:
+  push:
+    branches:
+      - nightly
```
ejguan, line 7: I have seen a nightly branch created and updated daily. Wondering if there is a way for us to manual trigger nightly release and all those build workflows?

ejguan (commented):
It seems conda releases are broken because of missing `portalocker`. We do have a `meta.yml` to track those dependencies for conda. Is there way to provide it to the workflow?

.github/workflows/build_conda_linux.yml
DanilBaibak: We have a separate Chronos job to trigger a nightly build https://fburl.com/chronos/udouc4ht So we can trigger it also manually.

ejguan: Ohhh, thank you!

ejguan:
BTW, another comment regarding enabling `BUILD_S3=1`. I was only able to compile the AWSSDK plugin to support minimum `manylinux_2_17` with all the following manual installation steps (AWSSDK doesn't work well with dynamic linking).
https://github.com/pytorch/data/blob/9e217f7154c2ed96d95e9d9eb9152178a4e0c6a2/.github/workflows/_build_test_upload.yml#L111-L137

test/smoke_test/smoke_test.py
```diff
@@ -26,5 +26,3 @@ def s3_test():
     options = parser.parse_args()
     if options.s3:
         s3_test()
-
-    main()
```
ejguan, line 30: Thank you! I have no idea what I was thinking.

.github/workflows/build_wheels_linux.yml
```diff
@@ -0,0 +1,47 @@
...
+            package-name: torchdata
+    name: ${{ matrix.repository }}
+    uses: pytorch/test-infra/.github/workflows/build_wheels_linux.yml@main
+    with:
+      repository: ${{ matrix.repository }}
+      ref: ""
+      test-infra-repository: pytorch/test-infra
+      test-infra-ref: main
+      build-matrix: ${{ needs.generate-matrix.outputs.matrix }}
+      pre-script: ${{ matrix.pre-script }}
```
ejguan, line 38: I think we need to do an extra step in `pre-script` for linux wheels. This is needed to convert torchdata wheels to `manylinux` compatible. I just checked the 3.7 release, it's not `manylinux2014` compatible by `auditwheel show torchdata-0.6.0.dev20221216+cpu-cp37-cp37m-linux_x86_64.whl`. And, apparently, `auditwheel repair` won't help.

Reason behind it is our C-Extension relying on CMake system from AWSSDK, which requires OpenSSL. I am not sure my approach is the best to make it manylinux compatible. Just want to share what I have done, I basically install static OpenSSL/libcrypto library manually.
https://github.com/pytorch/data/blob/0a0ae5d35be439c786c9ab798ea1596238d8640a/.github/workflows/_build_test_upload.yml#L124-L131 and specify those paths to let AWSSDK statically linking against OpenSSL https://github.com/pytorch/data/blob/0a0ae5d35be439c786c9ab798ea1596238d8640a/.github/workflows/_build_test_upload.yml#L153-L156

ejguan (approved):
Thank you for this effort. This is super helpful. Considering we might need to spend way more time on the manylinux compatible support, I am not going to block this PR since the wheels are not going to be uploaded. We can have a follow up PR to support it.

.github/workflows/build_wheels_linux.yml
ejguan: It should be similar to what TorchAudio is doing in https://github.com/pytorch/audio/blob/main/packaging/pre_build_script.sh

NivekT (approved):
Thank you so much! I will re-import and land after the last issue is resolved.

.github/workflows/build_wheels_linux.yml
DanilBaibak: We have the `pre-script` option exactly for this purpose. I agree, let's merge this PR and then start working on follow up PR to support manylinux compatible.

PR #923, opened 2022-12-13
Add smoke test

ejguan:
Per title.
Smoke test will be used by project Nova to test compiled TorchData

### DIFF
--- a/test/smoke_test/smoke_test.py
+++ b/test/smoke_test/smoke_test.py
@@ -1,0 +1,30 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
+# All rights reserved.
+#
+# This source code is licensed under the BSD-style license found in the
+# LICENSE file in the root directory of this source tree.
+
+
+import argparse
+
+import torchdata
+import torchdata.dataloader2
+import torchdata.datapipes
+
+
+def s3_test():
+    from torchdata._torchdata import S3Handler
+
+
+if __name__ == "__main__":
+    r"""
+    TorchData Smoke Test
+    """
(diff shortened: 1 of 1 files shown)


test/smoke_test/smoke_test.py
```diff
@@ -0,0 +1,30 @@
...
+#
+# This source code is licensed under the BSD-style license found in the
+# LICENSE file in the root directory of this source tree.
+
+
+import argparse
+
+import torchdata
+import torchdata.dataloader2
+import torchdata.datapipes
```
dahsh, line 12: Do we need `datapipes` here?

ejguan: Sorry, I am not sure what do you mean

NivekT (approved):
LGTM!

Question:
This looks a bit different. from the one in TorchVision (not saying this is wrong), there is probably an internal post/guide that I missed? How does this work?

DanilBaibak:
@NivekT we use these smoke tests to test binaries:

1. First build binaries - https://github.com/pytorch/test-infra/blob/main/.github/workflows/build_wheels_macos.yml#L118
2. Install the current build - https://github.com/pytorch/test-infra/blob/main/.github/workflows/build_wheels_macos.yml#L164
3. Than try to import the main library blocks - https://github.com/pytorch/test-infra/blob/main/.github/workflows/build_wheels_macos.yml#L177

Smoke tests may different for different domain libraries. Any ideas are welcome 😄

DanilBaibak (approved):
LGTM

PR #933, opened 2022-12-22
Adapt OSS FullSync for Internal monitoring & error attribution

Miiira:
Differential Revision: D42120515


ejguan (approved):
Approve from OSS

Issue #934, opened 2022-12-28
`_DataPipeSerializationWrapper` doesn't work with multiprocessing Queue


ejguan:
### 🐛 Describe the bug

After https://github.com/pytorch/data/pull/919 is landed, a hanging problem happens on MacOS or Windows, where `spawn` is used to create subprocesses by default. See: https://github.com/pytorch/data/actions/runs/3794926183
I was able to mitigate the issue by removing the [`SeializationWrapper`](https://github.com/pytorch/pytorch/blob/08a47549af6f3992502dcc711ae93fb81eb3628b/torch/utils/data/datapipes/datapipe.py#L326) from https://github.com/pytorch/data/blob/e15e1453967ce2f25f6fcd2838caadfd0e2fa811/torchdata/dataloader2/dataloader2.py#L112

And, the reason that the `SerializationWrapper` doesn't work is `multiprocessing.Queue` is attached to a `DataPipe` and sent to subprocesses. Even though I am able to solve my hanging problem in a different way, it's better to solve this problem directly via `SerializationWrapper`.

The following should be a minimum repro example
```py
ctx = mp.get_context("spawn")
q = ctx.Queue()
dp = IterableWrapper(list(range(10)))
# Attach a Queue
dp.q = q
dl = DataLoader2(dp, reading_service=PrototypeMultiProcessingReadingService(2, "spawn"))
for d in dl:
    pass
```

cc: @NivekT 

### Versions

main

ejguan:
This is the similar issue reported in PyTorch Core: https://github.com/pytorch/pytorch/issues/87003

PR #935, opened 2022-12-28
Forward fix to enable passing dispatching Queue between processes

ejguan:
Due to the issue https://github.com/pytorch/data/issues/934, I can't attach `mp.Queue` to the `_DummyDataPipe` and transfer it to subprocesses when `spawn` or `forkserver` is used.
So, I did the following works

## Changes:
- Extend `process_init_fn` to accept `dispatching_req_queue` and `dispatching_res_queue` without relying on `_DummyDataPipe`
- Add tests to cover all available mp start methods per platform
- A few nit changes in `dataloader2.py` and `reading_service.py` to make sure `Error` is ignored during `__del__`.

### DIFF
--- a/test/dataloader2/test_dataloader2.py
+++ b/test/dataloader2/test_dataloader2.py
@@ -4,13 +4,14 @@
 # This source code is licensed under the BSD-style license found in the
 # LICENSE file in the root directory of this source tree.
 
-
+import multiprocessing as mp
 import os
 import pickle
 import unittest
 from unittest import TestCase
 
 import torch
+from torch.testing._internal.common_utils import instantiate_parametrized_tests, parametrize
 
 from torch.utils.data.datapipes.iter.grouping import SHARDING_PRIORITIES
 
@@ -41,6 +42,8 @@ except ImportError:
 
 skipIfNoDill = unittest.skipIf(not HAS_DILL, "no dill")
 TEST_WITH_TSAN = os.getenv("PYTORCH_TEST_WITH_TSAN", "0") == "1"
+
+mp_ctx_parametrize = parametrize("ctx", mp.get_all_start_methods())
 
 
 class _ReadingServiceWrapper:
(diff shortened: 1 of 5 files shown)


ejguan:
All related CI passed after this patch.

torchdata/dataloader2/reading_service.py
```diff
@@ -304,7 +314,10 @@ def clean_me(process, req_queue, res_queue):
             # TODO(619): Can send terminations simultaneously
             # TODO(620): Make termination a function of QueueWrapperDataPipe (similar to reset)
             req_queue.put(communication.messages.TerminateRequest())
-            _ = res_queue.get()
+            try:
+                _ = res_queue.get(timeout=default_dl2_worker_join_timeout_in_s)
+            except queue.Empty:
+                pass
```
NivekT, line 320: I am trying to understand why this was not necessary before? Or was it always needed?

torchdata/dataloader2/dataloader2.py
```diff
@@ -169,14 +169,18 @@ def shutdown(self) -> None:
...
+                self._reset_iter = True
+                self._datapipe_iter = None
+            if not self._terminated:
+                if self.reading_service is not None:
+                    self.reading_service.finalize_iteration()
+                    self.reading_service.finalize()
+                self._terminated = True
+        # Ignore AttributeError in case any attribute has been removed before `__del__`
+        except AttributeError:
+            pass
```
NivekT, line 183: Just to make sure I'm understanding correctly, seems like this was always needed?

test/dataloader2/test_dataloader2.py
```diff
@@ -42,6 +43,8 @@
 skipIfNoDill = unittest.skipIf(not HAS_DILL, "no dill")
 TEST_WITH_TSAN = os.getenv("PYTORCH_TEST_WITH_TSAN", "0") == "1"
 
+mp_ctx_parametrize = parametrize("ctx", mp.get_all_start_methods())
```
NivekT, line 46: I should do it this way in my PR too 👍

NivekT (approved):
LGTM! Just a few clarifying questions in the comments.

Should we also edit `dispatch.py` to remove the variables from the `_DummyIterDataPipe` class?

Or are they still needed elsewhere (at a glance, can't see any usage)?

```python
class _DummyIterDataPipe(IterDataPipe):
    r"""
    This DataPipe is a placeholder to be replaced by the ``QueueWrapper``
    that connects the worker process for non-replicable DataPipe.
    """
    req_queue: Queue
    res_queue: Queue
```

ejguan:
> Should we also edit `dispatch.py` to remove the variables from the `_DummyIterDataPipe` class?

I chose not to change it because I still want to use those two variables to hold `Queues` in the future, when we can figure out how to properly serialize them. I will add a comment to `_DummyIterDataPipe`

torchdata/dataloader2/dataloader2.py
ejguan: Yea, as long as we are accessing attributes within `__del__`, we should handle such scenario.

torchdata/dataloader2/reading_service.py
ejguan: I guess it's because there is noting complicated whene `TerminateRquest` is received.
https://github.com/pytorch/data/blob/e15e1453967ce2f25f6fcd2838caadfd0e2fa811/torchdata/dataloader2/communication/iter.py#L149-L151

I felt this `try-except` should be a standard protocol for request-response design.

PR #942, opened 2023-01-11
Execute smoketest for release workflow

ejguan:
To reduce the time spent for release workflow, disable all TorchData test in the release workflows since each PR/main branch should have been tested. There is no need to run an extra round of tests.
And, use smoketest to make sure the compiled TorchData package having all required modules.

### DIFF
--- a/.github/workflows/_build_test_upload.yml
+++ b/.github/workflows/_build_test_upload.yml
@@ -196,7 +196,7 @@ jobs:
             conda activate wheel_build_env
           fi
           pip3 install dist/torchdata*.whl
-      - name: Run DataPipes Tests with pytest
+      - name: Run Smoke Tests
         shell: bash -l {0}
         env:
           PYTHON_VERSION: ${{ matrix.python-version }}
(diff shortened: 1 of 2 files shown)


.github/workflows/_build_test_upload.yml
```diff
@@ -367,15 +366,6 @@ jobs:
...
           conda install pytorch torchdata cpuonly -c $(pwd)/conda-bld/ -c "$CONDA_CHANNEL"
-      - name: Run DataPipes Tests with pytest
-        shell: bash -l {0}
-        run: |
-          if ${{ ! startsWith( matrix.os, 'windows' ) }}; then
-            source "${MINICONDA_INSTALL_PATH}/etc/profile.d/conda.sh"
-          fi
-          conda activate "${CONDA_ENV_PATH}"
-          conda env update --file test/requirements.yml
-          python -m pytest --no-header -v test --ignore=test/test_period.py --ignore=test/test_text_examples.py --ignore=test/test_audio_examples.py
```
ejguan, line 378: Remove tests for conda because the smoketest has been carried out by `meta.yaml`

NivekT: This might have been mentioned before but when does `meta.yaml` gets executed? After binaries are built and before upload? And is it only for conda?

ejguan: It's a conda specific mechanism. 
`conda build` will execute the tests specified in `meta.yaml` after finishes building TorchData
 See: https://github.com/pytorch/data/actions/runs/3892182151/jobs/6643307007#step:8:2119

PR #950, opened 2023-01-19
Add support to keep non-replicable DataPipe in the main process

ejguan:
This is useful when `fullsync` is in the pipeline and we don't want to make this DataPipe running in the worker process

### Changes

- Change the function names that is dispatching-related to `dispatching_xxx`
- Make `fullsync` DataPipe non-replicable
- Add `_find_replicable_branches` to find the last DataPipe prior to any non-replicable DataPipe
  - Add graph tests
- In `PrototypeMultiprocessingReadingService`, make sure only `replicable_datapipe` sent to worker process. And, replace the `replicable_datapipe` with the `worker_consumer_datapipe`.


torchdata/dataloader2/reading_service.py
```diff
@@ -261,7 +277,7 @@ def initialize(self, datapipe: DataPipe) -> DataPipe:
             )
             (process, req_queue, res_queue) = communication.eventloop.CreateProcessForDataPipeline(
                 ctx,
-                datapipe,
+                replicable_dp,
```
Miiira, line 280: To confirm my understanding, after this change process.start will only execute the replicable DataPipe, then we don't have the initialize_iteration ordering issue in SequentialRS?

torchdata/dataloader2/reading_service.py
```diff
@@ -247,6 +250,19 @@ def initialize(self, datapipe: DataPipe) -> DataPipe:
                 for res_queue in res_queues:
                     res_queue.cancel_join_thread()
 
+        # Find replicable branches for worker processes
+        # The rest of non-replicable DataPipes will remain in the main process
+        replicable_dps = _find_replicable_branches(graph)
+        assert (
```
Miiira, line 256: Also a n00b question for my understanding, we can have multiple non-replicable DP, with single replicable branch in between, e.g. non-replicable -> map_dp -> fullsync?

Miiira (approved):

torchdata/dataloader2/reading_service.py
ejguan: Order still matters especially when we need to let the first RS to specify which part should be sent to dispatching process.

torchdata/dataloader2/reading_service.py
ejguan: Yeah

torchdata/dataloader2/reading_service.py
```diff
@@ -270,13 +286,24 @@ def initialize(self, datapipe: DataPipe) -> DataPipe:
...
+        self._worker_consumer_datapipe = end_datapipe
 
-        self.end_datapipe = communication.iter._IterateQueueDataPipes(self.datapipes)  # type: ignore[assignment]
         if self.main_prefetch_cnt > 0:
-            self.end_datapipe = self.end_datapipe.prefetch(self.main_prefetch_cnt)  # type: ignore[union-attr]
+            end_datapipe = self._worker_consumer_datapipe.prefetch(self.main_prefetch_cnt)  # type: ignore[union-attr]
+            self._main_prefetch_datapipe = end_datapipe
+
+        # Attach non-replicable DataPipes
+        if replicable_dp is not datapipe:
```
wenleix, line 299: I guess this happens when we have non-replicable data pipe (and need to keep it in the main process). 

And the reason to replace `replicable_dp` with `end_datapipe` is because `end_datapipe` has "exchange sink" attached? (i.e. the `_IterateQueueDataPipes` on line 291)?

test/test_graph.py
```diff
@@ -304,93 +316,180 @@ def _make_dp(self):
...
+        single_br_dp, *_, fork_zip_dp, cir_br_dp, cir_map_dp, _, graph = self._make_dp()
+        make_non_replicable_dp(cir_map_dp)
+        dps = _find_replicable_branches(graph)
+        self.assertEqual(len(dps), 3)
+        self.assertTrue(all(dp in (single_br_dp, fork_zip_dp, cir_br_dp) for dp in dps))
+
+        single_br_dp, *_, fork_zip_dp, _, cir_map_dp, end_dp, graph = self._make_dp()
+        make_non_replicable_dp(end_dp)
+        dps = _find_replicable_branches(graph)
+        self.assertEqual(len(dps), 3)
```
wenleix, line 440: I assume we don't support execution of this case yet? (e.g. there are 3 branches for replicable dp).

Not for this PR. -- can we have a test case simply test if we can return 1 branch of replicable dp (but it's not the end_dp). Essentially simulate the production use case (e.g. after all dp there is a full sync dp)

wenleix (approved):

test/dataloader2/test_dataloader2.py
```diff
@@ -592,6 +603,37 @@ def test_dispatching_worker_determinism(self, ctx):
         dl.seed(321)
         self.assertNotEqual(res, list(dl) + list(dl))
 
+    @mp_ctx_parametrize
+    def test_non_replicable_datapipe(self, ctx) -> None:
+        r"""
+        For the pipeline with non-replicable DataPipe, make sure
```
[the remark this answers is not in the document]
ejguan: No, it can't. Like I said, `ShardingRRDispatchDP` labels the prior graph non-shardable and other non-shardable DataPipe labels itself and subsequent graph non-shardable.

test/test_graph.py
```diff
@@ -262,15 +264,24 @@ def test_multiprocessing_reading_service(self) -> None:
...
+def insert_round_robin_sharding(graph, datapipe):
+    dispatch_dp = ShardingRoundRobinDispatcher(datapipe, SHARDING_PRIORITIES.MULTIPROCESSING)
+    return replace_dp(graph, datapipe, dispatch_dp), dispatch_dp
 
 
 def replace_by_dummy(graph, datapipe):
     return replace_dp(graph, datapipe, _DummyIterDataPipe())
 
 
+def _is_replicable(self):
```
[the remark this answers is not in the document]
ejguan: I am trying to reduce the usage of any local function or lambda function, which are not recommended in TorchData due to pickle.
I can give a try to make it nested to see if it's working.

torchdata/dataloader2/reading_service.py
ejguan: Doing that because we send `replicable_dp` to worker process, which returns an `end_datapipe` with the exchange sink.
```
non_rep_dp1 (datapipe) -> non_rep_dp2 -> rep_dp_1 (replicable_dp) -> rep_dp2 -> ...
```
After replacement, the graph becomes
```
non_rep_dp1 (datapipe) -> non_rep_dp2 -> end_datapipe
```
In this case, we want to return `non_rep_dp1` (`datapipe`) rather than the `end_datapipe`.

For the case that the whole graph is replicable
```
rep_dp_1 (replicable_dp/datapipe) -> rep_dp2 -> ...
```
After launching mp, graph becomes
```
end_datapipe
```
So, we just need to keep `end_datapipe`.

test/test_graph.py
ejguan: > I assume we don't support execution of this case yet? (e.g. there are 3 branches for replicable dp).

No, we don't. Those tests aim to test the graph function.

> can we have a test case simply test if we can return 1 branch of replicable dp (but it's not the end_dp). Essentially simulate the production use case (e.g. after all dp there is a full sync dp)

I will add one shortly.

test/test_graph.py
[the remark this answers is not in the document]
ejguan: It works!

NivekT (approved):

PR #951, opened 2023-01-19
Replace easy_install by pip install to fix CI for MacOS

### LINKED ISSUE #948: Mysterious issue when install TorchData on CI MacOS machines with Python 3.9/3.10
### 🐛 Describe the bug

See examples:
- https://github.com/pytorch/data/actions/runs/3943431056
- https://github.com/pytorch/data/actions/runs/3932455178

Based on the [log](https://github.com/pytorch/data/actions/runs/3943431056/jobs/6748247866#step:9:1595), it looks like the installation script wants to find `torchdata` from pypi. It seems like `torchdata` has been added as a dependency of itself during installation. However, the same workflow works for other python versions.

It can't be produced on my local environment with the same version of python 3.9.16 and the same command in workflow.
And, I have validated the nightly releases of TorchData are working properly.

### Versions

main branch

ejguan:
Fixes #948

### Changes
- To fix MacOS CI, uses `pip install .` or `pip install -e .` to replace `python setup.py install` or `python setup.py develop`
  - `pip install` would use an isolated environment to prevent the problem in our CI
  - Add `ninja` to [`pyproject.toml`](https://github.com/pytorch/data/blob/main/pyproject.toml) to make sure CMake works properly using `pip install`
  - Easy install is deprecated https://setuptools.pypa.io/en/latest/deprecated/easy_install.html#easy-install
- Change documents to advocate `pip install` rather than `python setup.py`

### DIFF
--- a/README.md
+++ b/README.md
@@ -101,7 +101,7 @@ You can then proceed to run [our examples](https://github.com/pytorch/data/tree/
 ### From source
 
 ```bash
-python setup.py install
+pip install .
 ```
 
 If you'd like to include the S3 IO datapipes and aws-sdk-cpp, you may also follow
--- a/.github/workflows/_build_test_upload.yml
+++ b/.github/workflows/_build_test_upload.yml
@@ -481,7 +481,7 @@ jobs:
           fi
 
           pip3 install -r requirements.txt
-          python3 setup.py install
+          pip3 install .
       - name: Check env
         run: echo `which spinx-build`
       - name: Build the docset
--- a/.github/workflows/aistore_ci.yml
+++ b/.github/workflows/aistore_ci.yml
@@ -52,7 +52,7 @@ jobs:
         uses: NVIDIA/aistore@master
       - name: Build TorchData
         run: |
-          python setup.py install
+          pip3 install .
       - name: Install test requirements
         run: pip3 install -r test/requirements_aistore.txt
       - name: Run AIStore DataPipe tests with pytest
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -73,7 +73,7 @@ jobs:
           echo "/home/runner/.local/bin" >> $GITHUB_PATH
       - name: Build TorchData
         run: |
-          python setup.py install
+          pip3 install .
         env:
           BUILD_S3: 1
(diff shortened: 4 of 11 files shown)


ejguan:
Based on https://github.com/pytorch/data/actions/runs/3960562456, we are able to narrow down that it only happens when `BUILD_S3=1`.

ejguan:
Based on the log, TorchData has been successfully built for MacOS on python 3.9 and 3.10:
- Python 3.9: https://github.com/pytorch/data/actions/runs/3971317713/jobs/6808073890
- Python 3.10: https://github.com/pytorch/data/actions/runs/3971317713/jobs/6808074007

The failure should have been fixed by #957, which I will rebase onto to validate my change.

README.md
```diff
@@ -101,7 +101,7 @@ You can then proceed to run [our examples](https://github.com/pytorch/data/tree/
 ### From source
 
 ```bash
-python setup.py install
+pip install .
```
NivekT, line 104: Question:
1. Just to confirm, this is recommended for all users (not just MacOS)?
2. I wonder if PyTorch Core is impacted as well?

NivekT (approved):
LGTM! Just some questions

README.md
wenleix: Curious why replace `python setup.py install` with `pip install .` solve the issue? (maybe it's explained somewhere but didn't see it 😳)

ejguan: > 1. Just to confirm, this is recommended for all users (not just MacOS)?

Yea, it's a recommendation from Python

> 2. I wonder if PyTorch Core is impacted as well?

IDK. As far as I know, it doesn't

> Curious why replace `python setup.py install` with `pip install .` solve the issue?

The issue itself is mysterious enough to me as when doing `python setup.py install` it added `torchdata` as its dependency -> circular dependency. See log: https://github.com/pytorch/data/actions/runs/3972365525/jobs/6810200117#step:9:1639
I don't know why it happens only on macos python 3.9 and python 3.10.

So, I have no clue why `pip install` can solve this problem. Since it does solve it anyway, I decide not to spend extra time to dive deeper on this topic until I have more time.

PR #952, opened 2023-01-19
Huggingface all options

### LINKED ISSUE #944: Full Support for HuggingFace-Datasets
### 🚀 The feature

Currently `HuggingFaceHubReaderIterDataPipe` is a wrapper for `load_dataset()` (https://huggingface.co/docs/datasets/v2.6.1/en/package_reference/loading_methods#datasets.load_dataset). `load_dataset()` however supports many more optional arguments. 
If we want to fully support downloading from HuggingFace we need to add these options to `HuggingFaceHubReaderIterDataPipe` as well.

### Motivation, pitch

E.g.  the [oscar](https://huggingface.co/datasets/oscar)-dataset requires you to also provide a name, which is not possible when using `HuggingFaceHubReaderIterDataPipe`.

I could provide a fix for this.

### Alternatives

_No response_

### Additional context

_No response_

SvenDS9:
Fixes #944

### Changes

- Changed test setup for HuggingFaceHubReader. Do not test against production but ensure `load_dataset` (from HuggingFace) is called with correct parameters
- Include HuggingFaceHubReader in documentation

@ejguan could you please have a look

### DIFF
--- a/test/test_huggingface_datasets.py
+++ b/test/test_huggingface_datasets.py
@@ -4,15 +4,12 @@
 # This source code is licensed under the BSD-style license found in the
 # LICENSE file in the root directory of this source tree.
 
-import os
 import unittest
-import warnings
+from unittest.mock import patch
 
 import expecttest
 
-from _utils._common_utils_for_test import create_temp_dir, create_temp_files, reset_after_n_next_calls
-
-from torchdata.datapipes.iter import HuggingFaceHubReader, IterableWrapper
+from torchdata.datapipes.iter import HuggingFaceHubReader
 
 try:
     import datasets
@@ -25,32 +22,24 @@ skipIfNoDatasets = unittest.skipIf(not HAS_DATASETS, "no datasets")
 
 
 class TestHuggingFaceHubReader(expecttest.TestCase):
-    def setUp(self):
-        self.temp_dir = create_temp_dir()
-        self.temp_files = create_temp_files(self.temp_dir)
-        self.temp_sub_dir = create_temp_dir(self.temp_dir.name)
-        self.temp_sub_files = create_temp_files(self.temp_sub_dir, 4, False)
-
-        self.temp_dir_2 = create_temp_dir()
-        self.temp_files_2 = create_temp_files(self.temp_dir_2)
-        self.temp_sub_dir_2 = create_temp_dir(self.temp_dir_2.name)
-        self.temp_sub_files_2 = create_temp_files(self.temp_sub_dir_2, 4, False)
-
-    def tearDown(self):
-        try:
-            self.temp_sub_dir.cleanup()
-            self.temp_dir.cleanup()
-            self.temp_sub_dir_2.cleanup()
-            self.temp_dir_2.cleanup()
-        except Exception as e:
-            warnings.warn(f"HuggingFace datasets was not able to cleanup temp dir due to {e}")
-
     @skipIfNoDatasets
-    def test_huggingface_hubreader(self):
-        datapipe = HuggingFaceHubReader(dataset="lhoestq/demo1", revision="main", streaming=True)
+    @patch("datasets.load_dataset")
+    def test_huggingface_hubreader(self, mock_load_dataset):
+        mock_load_dataset.return_value = datasets.Dataset.from_dict(
+            {
+                "id": ["7bd227d9-afc9-11e6-aba1-c4b301cdf627", "7bd22905-afc9-11e6-a5dc-c4b301cdf627"],
+                "package_name": ["com.mantz_it.rfanalyzer"] * 2,
+            }
+        )
+
+        datapipe = HuggingFaceHubReader("lhoestq/demo1", revision="branch", streaming=False, use_auth_token=True)
+
         elem = next(iter(datapipe))
         assert type(elem) is dict
         assert elem["package_name"] == "com.mantz_it.rfanalyzer"
+        mock_load_dataset.assert_called_with(
+            path="lhoestq/demo1", streaming=False, split="train", revision="branch", use_auth_token=True
+        )
 
 
 if __name__ == "__main__":
--- a/test/test_huggingface_datasets.py
+++ b/test/test_huggingface_datasets.py
@@ -34,12 +34,17 @@ class TestHuggingFaceHubReader(expecttest.TestCase):
 
         datapipe = HuggingFaceHubReader("lhoestq/demo1", revision="branch", streaming=False, use_auth_token=True)
 
-        elem = next(iter(datapipe))
+        iterator = iter(datapipe)
+        elem = next(iterator)
         assert type(elem) is dict
+        assert elem["id"] == "7bd227d9-afc9-11e6-aba1-c4b301cdf627"
         assert elem["package_name"] == "com.mantz_it.rfanalyzer"
         mock_load_dataset.assert_called_with(
             path="lhoestq/demo1", streaming=False, split="train", revision="branch", use_auth_token=True
         )
+        with self.assertRaises(StopIteration):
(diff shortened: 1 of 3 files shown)


torchdata/datapipes/iter/load/huggingface.py
```diff
@@ -17,42 +17,48 @@
 
 
 def _get_response_from_huggingface_hub(
-    dataset: str, split: str, revision: str, streaming: bool, data_files: Optional[Dict[str, str]]
+    path: str, streaming=True, split="train", revision="main", **config_kwargs
```
ejguan, line 20: nit: Could you please add type annotation to each argument?

torchdata/datapipes/iter/load/huggingface.py
```diff
@@ -17,42 +17,48 @@
...
 class HuggingFaceHubReaderIterDataPipe(IterDataPipe[Tuple[str, StreamWrapper]]):
     r"""
-    Takes in dataset names and returns an Iterable HuggingFace dataset
-    Args format is the same as https://huggingface.co/docs/datasets/loading
+    Takes in dataset names and returns an Iterable HuggingFace dataset.
+    Args format and meaning are the same as https://huggingface.co/docs/datasets/loading.
+    Contrary to their implementation, default behavior differs in the following:
+    split is set to "train".
+    revision is set to "main".
+    streaming is set to True.
```
ejguan, line 33: nit: Add a tab before each line
Add one empty line before `Args`

torchdata/datapipes/iter/load/huggingface.py
```diff
@@ -17,42 +17,48 @@
...
+
+        datasets.load_dataset = MagicMock(return_value=datasets.Dataset.from_dict(
+            {"id": ["7bd227d9-afc9-11e6-aba1-c4b301cdf627", "7bd22905-afc9-11e6-a5dc-c4b301cdf627" ], "package_name": ["com.mantz_it.rfanalyzer"] * 2}
+        ))
+
+    .. testcode::
+
+        huggingface_reader_dp = HuggingFaceHubReader("lhoestq/demo1", revision="main")
+        elem = next(iter(huggingface_reader_dp))
+        assert elem["package_name"] == "com.mantz_it.rfanalyzer"
```
ejguan, line 52: Thank you!

test/test_huggingface_datasets.py
```diff
@@ -25,32 +22,24 @@
...
+        )
+
+        datapipe = HuggingFaceHubReader("lhoestq/demo1", revision="branch", streaming=False, use_auth_token=True)
+
         elem = next(iter(datapipe))
         assert type(elem) is dict
         assert elem["package_name"] == "com.mantz_it.rfanalyzer"
+        mock_load_dataset.assert_called_with(
+            path="lhoestq/demo1", streaming=False, split="train", revision="branch", use_auth_token=True
+        )
```
ejguan, line 42: Could you please add one more line to test if there is only one element yielded from the datapipe?

torchdata/datapipes/iter/load/huggingface.py
SvenDS9: Would it be better to align this with the default arguments of the load_dataset() function (e.g. split is None)? 
Or would this lead to problems elsewhere? A quick Github search reveals no usage of the HuggingFaceHubReader outside of Pytorch Data.

test/test_huggingface_datasets.py
SvenDS9: What exactly do you mean? `_get_response_from_huggingface_hub()` returns an iterator over the dataset and we look at the first element in line 37-39.

torchdata/datapipes/iter/load/huggingface.py
ejguan: Good question.
For PyTorch project, we want to have a deprecation cycle for those kind of BC-breaking changes. You can keep `split` as `train` for now and add a deprecation warning to indicate users that `train` will be changed to `None` in the release `0.7`

test/test_huggingface_datasets.py
ejguan: This test validates the first output is the right one. I want to check a StopIteration should be raised when calling next over the iterator one more time

torchdata/datapipes/iter/load/huggingface.py
NivekT: nit: can we preserve the previous ordering of arguments for the most backward compatibility? (unless there is a reason why it should be changed)

torchdata/datapipes/iter/load/huggingface.py
```diff
@@ -61,20 +73,14 @@ def __init__(
...
-        self.dataset = dataset
-        self.split = split
-        self.revision = revision
-        self.streaming = streaming
-        self.data_files = data_files
+        self.datset = dataset
+        self.config_kwargs = config_kwargs
+        warnings.warn(
+            "default behavior of HuggingFaceHubReader will change in version 0.7", DeprecationWarning, stacklevel=2
+        )
```
ejguan, line 80:
```suggestion
if "split" not in self.config_kwargs:
    warnings.warn("Default value of `split` will be changed to None in version 0.7", FutureWarning)

```

ejguan (approved):
LGTM. @NivekT Do you want to chime in for any concern?

torchdata/datapipes/iter/load/huggingface.py
```diff
@@ -62,19 +74,14 @@ def __init__(
...
         self.dataset = dataset
-        self.split = split
-        self.revision = revision
-        self.streaming = streaming
-        self.data_files = data_files
+        self.config_kwargs = config_kwargs
+        if "split" not in self.config_kwargs:
+            warnings.warn("Default value of `split` will be changed to None in version 0.7", FutureWarning)
+        if "revision" not in self.config_kwargs:
+            warnings.warn("Default value of `revision` will be changed to None in version 0.7", FutureWarning)
```
NivekT, line 81: Looks like the default arguments are changed. @ejguan it will be slightly BC-breaking. WDYT?

ejguan: Wait, the default arguments remain the same right?
`split="train"`
`revision="main"`
`streaming=True`

The default arguments are assigned in `_get_response_from_huggingface_hub`.

NivekT: Oh in that case I'm fine with it, only that users will not be able to see those default arguments from IDE autocomplete. which is suboptimal but not a blocker. 

Is the warning for Streaming missing or we want it to stay `True`? The default is `False` for [HG's version](https://huggingface.co/docs/datasets/v2.8.0/en/package_reference/loading_methods#datasets.load_dataset).

ejguan: > Is the warning for Streaming missing or we want it to stay `True`?

Good question. I personally like streaming=True to incorporate the style of large-dataset.

ejguan:
@SvenDS9 Could you please do a rebase onto main branch?

torchdata/datapipes/iter/load/huggingface.py
```diff
@@ -17,42 +16,55 @@
 
 
 def _get_response_from_huggingface_hub(
-    dataset: str, split: str, revision: str, streaming: bool, data_files: Optional[Dict[str, str]]
+    dataset: str,
+    split: Union[str, datasets.Split] = "train",
+    revision: Union[str, datasets.Version] = "main",
```
ejguan, line 21: Let's keep the annotation as `str` to make sure `datasets` as optional dependency

NivekT (approved):

NivekT (approved):

PR #967, opened 2023-01-25
Improve reset epoch sent to all workers before blocking to receive

ejguan:
### Changes

- Send request to workers before requesting response to reduce the latency for each new epoch
- Minor change for resetting `replicable_dp`

### DIFF
--- a/torchdata/dataloader2/reading_service.py
+++ b/torchdata/dataloader2/reading_service.py
@@ -259,8 +259,7 @@ class PrototypeMultiProcessingReadingService(ReadingServiceInterface):
         replicable_dp = replicable_dps[0]
 
         if self.worker_prefetch_cnt > 0:
-            worker_prefetch_dp = replicable_dp.prefetch(self.worker_prefetch_cnt)
-            replicable_dp = worker_prefetch_dp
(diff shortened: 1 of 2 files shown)


torchdata/dataloader2/reading_service.py
```diff
@@ -259,8 +259,7 @@ def initialize(self, datapipe: DataPipe) -> DataPipe:
         replicable_dp = replicable_dps[0]
 
         if self.worker_prefetch_cnt > 0:
-            worker_prefetch_dp = replicable_dp.prefetch(self.worker_prefetch_cnt)
-            replicable_dp = worker_prefetch_dp
+            replicable_dp = replicable_dp.prefetch(self.worker_prefetch_cnt)
```
NivekT, line 262: Interesting `worker_prefetch_dp` wasn't used anywhere else

NivekT (approved):
LGTM! Thanks!

torchdata/dataloader2/reading_service.py
ejguan: It was my bad, I was testing the graph earlier in the PR of dispatching process but forgot to optimize this part.

PR #968, opened 2023-01-26
Exception handling in online datapipes

### LINKED ISSUE #963: Modify exception handling of online Datapipes
### 🚀 The feature

Add an optional argument to `HTTPReaderIterDataPipe `to specify error handling.

https://github.com/pytorch/data/blob/2ca1fa6483e58c6428319393e1aab4c26f576bec/torchdata/datapipes/iter/load/online.py#L89-L94
`_get_response_from_http()` raises an exception if an error occurs. 

### Motivation, pitch

While working with `HTTPReaderIterDataPipe` I have noticed that failing HTTP-requests are not handled gracefully. This means that the DataPipe cannot be continued after an error occurs. Therefore I propose to add an optional argument which allows the user to continue the datapipe with the next url in case of an error.

### Alternatives

See https://discuss.pytorch.org/t/exception-handling-when-using-datapipes/170959

### Additional context

This also affects the other Datapipes in online.py which read from remote. It would probably make sense to add this feature there as well.

SvenDS9:
Fixes #963

### Changes

- Add option to `HTTPReader` to skip over urls causing problems
- Add test for this behavior


SvenDS9:
While working on this, I noticed a few things:
https://github.com/pytorch/data/blob/2ca1fa6483e58c6428319393e1aab4c26f576bec/torchdata/datapipes/iter/load/online.py#L21-L33
As the issue https://github.com/python/cpython/issues/86793 seems to have been resolved, this can probably be removed.

https://github.com/pytorch/data/blob/2ca1fa6483e58c6428319393e1aab4c26f576bec/torchdata/datapipes/iter/load/online.py#L40-L53
Here we open a requests-session, which is probably unnecessary as we only do one GET-request and then close the session again.

In addition we convert the exceptions from `HTTPError`/`RequestException `to a simple Exception, why?

NivekT:
@SvenDS9 Thanks for spotting these!

1. Feel free to remove the proxies but please test to make sure it is fine.
2. Session object is probably not needed unless we want to use it multiple times (e.g. retry). Feel free to change that if you confirm it is not needed.
3. I don't really remember why they are turned into general Exceptions. I think raising the original ones should be fine. We do want to make sure that whatever error message it produces, it is easy for user to identify that it comes from HttpReader.

torchdata/datapipes/iter/load/online.py
```diff
@@ -36,21 +35,11 @@ def _get_proxies() -> Optional[Dict[str, str]]:
...
-        return url, StreamWrapper(r.raw)
-    except HTTPError as e:
-        raise Exception(f"Could not get the file. [HTTP Error] {e.response}.")
-    except RequestException as e:
-        raise Exception(f"Could not get the file at {url}. [RequestException] {e.response}.")
-    except Exception:
-        raise
+    with requests.Session() as session:
+        proxies = _get_proxies()
+        r = session.get(url, timeout=timeout, proxies=proxies, stream=True, **query_params)  # type: ignore[arg-type]
```
NivekT, line 40: This looks good to me since I think `timeout` is `None` default.

torchdata/datapipes/iter/load/online.py
```diff
@@ -80,18 +70,26 @@ class HTTPReaderIterDataPipe(IterDataPipe[Tuple[str, StreamWrapper]]):
...
         self.query_params = kwargs
 
     def __iter__(self) -> Iterator[Tuple[str, StreamWrapper]]:
         for url in self.source_datapipe:
-            if self.query_params:
+            try:
                 yield _get_response_from_http(url, timeout=self.timeout, **self.query_params)
-            else:
-                yield _get_response_from_http(url, timeout=self.timeout)
+            except RequestException as e:
```
NivekT, line 88: Question: What makes you decide to catch `RequestException` vs other ones or all of them?

NivekT (commented):
Overall looks good to me. One question on what type of exception we should skip over.

I think retry will be very helpful if we can have that.

cc: @ejguan to have a look at the API

torchdata/datapipes/iter/load/online.py
SvenDS9: Good question. All exceptions explicitly thrown by Requests inherit from RequestException (see [doc](https://requests.readthedocs.io/en/latest/user/quickstart/#errors-and-exceptions)), but just to be safe I'll change this to `Exception`.

torchdata/datapipes/iter/load/online.py
```diff
@@ -205,18 +220,33 @@ class OnlineReaderIterDataPipe(IterDataPipe[Tuple[str, StreamWrapper]]):
...
-                yield _get_response_from_google_drive(url, timeout=self.timeout)
-            else:
-                yield _get_response_from_http(url, timeout=self.timeout)
+            try:
+                parts = urllib.parse.urlparse(url)
+
+                if re.match(r"(drive|docs)[.]google[.]com", parts.netloc):
+                    yield _get_response_from_google_drive(url, timeout=self.timeout, **self.query_params)
+                else:
+                    yield _get_response_from_http(url, timeout=self.timeout, **self.query_params)
```
ejguan, line 244: Can you please wrap `try-except` around `_get_response_from_google_drive` or `_get_response_from_http`?
In your current implementation, there is a chance the skipped Error comes from `parts = urllib.parse.urlparse(url)` or `re.match(r"(drive|docs)[.]google[.]com", parts.netloc)`.

ejguan (commented):
Overall LGTM with one comment

torchdata/datapipes/iter/load/online.py
NivekT: I will import and merge this after this change. Thanks!

SvenDS9: I have implemented this change, but I don't really understand why it is necessary. In my opinion the source of the exception doesn't really matter if we want to skip over them anyway. With this change exceptions caused by trying to parse the url will not be caught.

NivekT: > With this change exceptions caused by trying to parse the url will not be caught.

I think that is the point. If the URL cannot be parsed, perhaps users want to know and fix it. If you cannot get a response, then they may want to skip it.

NivekT: > With this change exceptions caused by trying to parse the url will not be caught.

I think that is the point. If the URL cannot be parsed, perhaps users want to know and fix it. If you cannot get a response, then they may want to skip it.

NivekT (approved):

Issue #971, opened 2023-01-27
Drop python3.7 and experiment python 3.11


ejguan:
### 🚀 The feature

PyTorch and other domains will drop support for python 3.7. So, TorchData will follow the same schedule to drop 3.7 support in 0.6.0 release.
And, since PyTorch might start to support python 3.11 in the next release, we should start to experiment torchdata with Python 3.11

### Motivation, pitch

Align the supported python versions across pytorch org.

ejguan:
To drop python 3.7 support, we should remove python 3.7 from our release CI on both *main* and *release/0.6* branches, right after we finish the branch cut.

ejguan:
Closing it for now

PR #974, opened 2023-01-27
Drop python 3.7 support as aligned with PyTorch

ejguan:
First step for #971

per title


NivekT (approved):
Wooooo!

https://docs.python.org/3/whatsnew/3.8.html

ejguan:
As far as I know, Python 3.8 introduced two good things:
- `asyncio` becomes stable
- `Protocol` is supported in `typing`

PR #977, opened 2023-01-30
Only run wheels/package build step for PRs with changes of the CI workflows

ejguan:
Reduce the waiting time on CI for PRs.


NivekT (commented):
n00b question: IIUC normal code changes in Python cannot break wheels/package build steps?

ejguan:
> IIUC normal code changes in Python cannot break wheels/package build steps?

Actually, it's a good question. It might be reasonable to test if any change would break the release. Let me keep it as it is..

PR #983, opened 2023-02-02
Fix lint for SequentialReadingService and unify the API for process_reset_fn

ejguan:
This is the follow up for 807db8f8c7282b2f48b48b1e07439c119a2ba12f

### Changes

- Fix lint
- Fix checkpoint result
- Add inline doc
- Unify the API for process_reset_fn and dispatch_process_reset_fn

After this PR, I will land a test case and tutorial for SequentialRS

### DIFF
--- a/torchdata/dataloader2/reading_service.py
+++ b/torchdata/dataloader2/reading_service.py
@@ -328,7 +328,7 @@ class PrototypeMultiProcessingReadingService(ReadingServiceInterface):
                 self._main_prefetch_datapipe.reset()  # type: ignore[union-attr]
             # Send the shared seed to subprocesses
             call_on_epoch_reset = partial(
-                process_reset_fn, custom_reset_fn=self.worker_reset_fn, custom_dispatch_process_reset_fn=iter_reset_fn
+                process_reset_fn, iter_reset_fn=iter_reset_fn, custom_reset_fn=self.worker_reset_fn
             )
             assert self._worker_consumer_datapipe is not None
             self._worker_consumer_datapipe.reset_epoch(call_on_epoch_reset, seed_generator)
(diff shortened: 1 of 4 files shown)


torchdata/dataloader2/reading_service.py
```diff
@@ -303,7 +310,9 @@ def initialize(self, datapipe: DataPipe) -> DataPipe:
 
         return self._end_datapipe  # type: ignore[return-value]
 
-    def initialize_iteration(self, seed_generator: SeedGenerator, iter_reset_fn: Optional[Callable[[DataPipe], DataPipe]] = None) -> Optional[Callable[[DataPipe], DataPipe]]:
+    def initialize_iteration(
+        self, seed_generator: SeedGenerator, iter_reset_fn: Optional[Callable[[DataPipe], DataPipe]] = None
+    ) -> Optional[Callable[[DataPipe], DataPipe]]:
```
NivekT, line 315: Question: where is the case when it returns something other than `None`?

torchdata/dataloader2/reading_service.py
```diff
@@ -540,9 +560,12 @@ def checkpoint(self) -> bytes:
         states = []
         for rs in self.reading_services:
             states.append(rs.checkpoint())
+        return b"\n".join(states)
```
NivekT, line 563: This is more of a placeholder for now? To be changed later?

NivekT (approved):
LGTM with some questions!

torchdata/dataloader2/reading_service.py
ejguan: This is a valid solution to combine multiple `bytes`, unless you have any concern on this solution.

PR #985, opened 2023-02-03
Enable SequentialReadingService to support MP + Distributed

### LINKED ISSUE #911: `DistributedReadingService` supports multi-processing reading
### 🚀 The feature

`TorchData` is a great work for better data loading! I have tried it and it gives me a nice workflow with tidy code-style.❤️

When using DDP, I work with the `DataLoader2` where `reading_service=DistributedReadingService()`. I find this service runs one worker for outputting datas per node. This means it has lower reading throughput than the legacy `DataLoader`, which utilizes multiple workers with the total worker number = `num_workers * world_size`.

Therefore, is it possible to combine `DistributedReadingService` with multi-processing reading? This could be possibly done by introducing `PrototypeMultiProcessingReadingService` into  `DistributedReadingService` (Just guessing. I'm not a pro for handling this.).

### Motivation, pitch

I think this feature could be a part of #427 . The detailed motivation is declared above.

### Alternatives

_No response_

### Additional context

_No response_

ejguan:
Fixes #911

### Changes

- Remove distributed code from PrototypeMPRS
- Fix a bug of `blocking_request_get` not sent to worker process
- Enable `SequentialReadingService` to combine both Distributed and MP ReadingService
- Add tests for `SequentialReadingService`
- Add tutorial for `SequentialReadingService`

### DIFF
--- a/docs/source/dlv2_tutorial.rst
+++ b/docs/source/dlv2_tutorial.rst
@@ -50,3 +50,21 @@ Distributed
         for d in dl:
             model(d)
     dl.shutdown()
+
+Multiprocessing + Distributed
+------------------------------
+
+``SequentialReadingService`` can be used to combine both ``ReadingServices`` together to achive multiprocessing and distributed training at the same time.
+
+.. code:: python
+
+    mp_rs = PrototypeMultiProcessingReadingService(num_workers=4)
+    dist_rs = DistributedReadingService()
+    rs = SequentialReadingService(dist_rs, mp_rs)
+
+    dl = DataLoader2(datapipe, reading_service=rs)
+    for epoch in range(10):
+        dl.seed(epoch)
+        for d in dl:
+            model(d)
+    dl.shutdown()
--- a/test/dataloader2/test_dataloader2.py
+++ b/test/dataloader2/test_dataloader2.py
@@ -12,7 +12,6 @@ import random
 import socket
 import unittest
 
-from functools import partial
 from unittest import TestCase
 
 import numpy as np
--- a/test/dataloader2/test_dataloader2.py
+++ b/test/dataloader2/test_dataloader2.py
@@ -711,6 +711,10 @@ def _dist_one_epoch(dl):
 
 
 def _finalize_distributed_queue(rank, q):
+    r"""
+    Synchronize all distributed processes to guarantee all data have been put into
+    the Multiprocessing Queue.
+    """
     pg = dist.new_group(backend="gloo")
     end_tensor = torch.tensor([rank], dtype=torch.int64)
     dist.all_reduce(end_tensor, group=pg)
@@ -721,6 +725,9 @@ def _finalize_distributed_queue(rank, q):
 
 
 def _random_fn(data):
+    r"""
+    Used to validate the randomness of subprocess-local RNGs are set deterministically.
+    """
     py_random_num = random.randint(0, 2 ** 32)
     np_random_num = np.random.randint(0, 2 ** 32)
     torch_random_num = torch.randint(0, 2 ** 32, size=[]).item()
--- a/test/dataloader2/test_dataloader2.py
+++ b/test/dataloader2/test_dataloader2.py
@@ -7,10 +7,18 @@
 import multiprocessing as mp
 import os
 import pickle
+import queue
+import random
+import socket
 import unittest
+
+from functools import partial
 from unittest import TestCase
 
+import numpy as np
+
 import torch
+import torch.distributed as dist
 from torch.testing._internal.common_utils import instantiate_parametrized_tests, parametrize
 
 from torch.utils.data.datapipes.iter.grouping import SHARDING_PRIORITIES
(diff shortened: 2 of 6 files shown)


torchdata/dataloader2/communication/eventloop.py
```diff
@@ -92,7 +92,7 @@ def _create_datapipe_queue_loop(source_datapipe, req_queue, res_queue, blocking_
     return pipe_type.DataPipeBehindQueues(
         source_datapipe,
         protocol_type(req_queue, res_queue),
-        blocking_request_get=True,
+        blocking_request_get=blocking_request_get,
```
ejguan, line 95: This was a bug that would blocking the dispatching process since there will be multiple loops running in the same process. And, we need to make sure each loop won't block each other.
And, it should fix the problem for DPP + MPRS with Fullsync

torchdata/dataloader2/reading_service.py
```diff
@@ -206,23 +203,13 @@ def __init__(
...
-        self._world_size = 1
-        self._rank = 0
 
     def initialize(self, datapipe: DataPipe) -> DataPipe:
         r"""
         ``PrototypeMultiProcessingReadingService`` finds information about sharding,
         separates graph by multiple pieces and reconnects it using queues.
         creates subprocesses.
         """
-        if dist.is_available() and dist.is_initialized():
```
wenleix, line 219: Nice :)

test/dataloader2/test_dataloader2.py
```diff
@@ -635,7 +653,204 @@ def test_non_replicable_datapipe(self, ctx) -> None:
...
+    port = s.getsockname()[1]
+    s.close()
+    return str(port)
+
+
+class TerminateSignal:
+    pass
+
+
+def _launch_distributed_training(world_size, *args, fn):
```
wenleix, line 672: I asked ChatGPT what does this function do: 

> The program is a function _launch_distributed_training that launches a distributed training process. The function takes in parameters world_size, *args, and fn. The environment variable MASTER_ADDR is set to TEST_MASTER_ADDR, and the environment variable MASTER_PORT is set to a value returned from a call to the _get_open_port function. The function creates a multiprocessing context using the spawn method, and creates a queue q using the context.

> The function then creates world_size processes using the Process method of the context and starts each process. The target of each process is the function fn, and the arguments for each process are rank, world_size, q, and *args. The function stores the created processes in a list ps.

> The function then uses a while loop to get data from the queue q and append it to a list res. The loop breaks when a TerminateSignal is received from the queue. After the loop, the function joins all processes in the ps list. Finally, the function returns the res list.

Seems quite correct? ~

wenleix (approved):
LGTM.

torchdata/dataloader2/communication/iter.py
```diff
@@ -263,8 +263,6 @@ def reset(self):
     def reset_epoch(
         self, reset_fn: Callable[[WorkerInfo, SeedGenerator, DataPipe], DataPipe], seed_generator: SeedGenerator
     ):
-        for dp in self.datapipes:
-            dp.protocol.discard_existing_request()
```
NivekT, line 267: Is this no longer needed?

docs/source/dlv2_tutorial.rst
```diff
@@ -50,3 +50,21 @@ Distributed
         for d in dl:
             model(d)
     dl.shutdown()
+
+Multiprocessing + Distributed
```
NivekT, line 54: Non-blocking but it will be nice to add this to our Colab example as well!

NivekT (approved):
LGTM!

test/dataloader2/test_dataloader2.py
dracifer: what happens if we give the description back to ChatGPT and ask it to write a code~~ "write a function that (quote~)"

docs/source/dlv2_tutorial.rst
```diff
@@ -50,3 +50,21 @@ Distributed
...
+------------------------------
+
+``SequentialReadingService`` can be used to combine both ``ReadingServices`` together to achive multiprocessing and distributed training at the same time.
```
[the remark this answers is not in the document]
ejguan: I guess so when AIStoreRS or RayRS is provided.

torchdata/dataloader2/communication/iter.py
```diff
@@ -263,8 +263,6 @@ def reset(self):
     def reset_epoch(
         self, reset_fn: Callable[[WorkerInfo, SeedGenerator, DataPipe], DataPipe], seed_generator: SeedGenerator
     ):
-        for dp in self.datapipes:
```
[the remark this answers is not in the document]
ejguan: This is not useful at all. There is supposed to be non-request when doing `reset_epoch`.

test/dataloader2/test_dataloader2.py
ejguan: Haha, writing those comments would take me more time than writing code.

torchdata/dataloader2/communication/iter.py
ejguan: I mean I can keep it but TBH I don't think it's really useful.

docs/source/dlv2_tutorial.rst
ejguan: Sounds reasonable to me.

torchdata/dataloader2/communication/iter.py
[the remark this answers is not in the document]
ejguan: Let me keep it then.

PR #987, opened 2023-02-06
Refactor `ShardingFilterIterDataPipe` into a seperate file (#94095)

wenleix:
Summary:
Move `ShardingFilterIterDataPipe` into a dedicated file.

Also, propose to have a dedicated parent class (`_ShardingIterDataPipe`) for sharding data pipe, as this seems more like a "system/engine-level" datapipe that gives strong hints to RS on how to execute, and needs first-class citizen treatment in RS (compared with other "user-level" datapipe that are mostly composable `Callable[[Iterable], Iterable]`.  But open to other discussions.

Open question: Should
[ShardingRoundRobinDispatcherIterDataPipe](https://github.com/pytorch/data/blob/01fc76200354501b057bb439b43a1f05f609dd0a/torchdata/datapipes/iter/util/sharding.py#L16-L17) also be considered as a `_ShardingIterDataPipe`? (e.g. this sharding is executed by replicating (the metadata), while `ShardingRoundRobinDispatcherIterDataPipe` hints too expensive to replicate so requires round robin data exchange/dispatch).

Differential Revision:
D43014692

X-link: https://github.com/pytorch/pytorch/pull/94095

D43014692

### DIFF
--- a/test/dataloader2/test_dataloader2.py
+++ b/test/dataloader2/test_dataloader2.py
@@ -20,7 +20,7 @@ import torch
 import torch.distributed as dist
 from torch.testing._internal.common_utils import instantiate_parametrized_tests, parametrize
 
-from torch.utils.data.datapipes.iter.grouping import SHARDING_PRIORITIES
+from torch.utils.data.datapipes.iter.sharding import SHARDING_PRIORITIES
 
 from torchdata.dataloader2 import (
     communication,
--- a/test/test_graph.py
+++ b/test/test_graph.py
@@ -14,7 +14,7 @@ import expecttest
 from _utils._common_utils_for_test import IS_WINDOWS
 
 from torch.utils.data import IterDataPipe
-from torch.utils.data.datapipes.iter.grouping import SHARDING_PRIORITIES
+from torch.utils.data.datapipes.iter.sharding import SHARDING_PRIORITIES
 
 from torchdata.dataloader2 import DataLoader2, MultiProcessingReadingService, ReadingServiceInterface
 from torchdata.dataloader2.graph import find_dps, list_dps, remove_dp, replace_dp, traverse_dps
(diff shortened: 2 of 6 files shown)


NivekT (approved):

pmeier:
This is BC breaking. The original module (`torch.utils.data.datapipes.iter.grouping`) as well as the class (`ShardingFilterIterDataPipe`) were public. For example, `torchvision` uses it [here](https://github.com/pytorch/vision/blob/c05ad81bcb8a042a67abfd57efabb0778584b922/torchvision/prototype/datasets/utils/_internal.py#L107).

For us it is not a large problem to change, but other libraries that rely on that might break in the upcoming release.

ejguan:
@pmeier You are right. How about we reference those two classes back to `grouping` from `sharding.py` and add a deprecation warning? @wenleix

pmeier:
Just to give some unsolicited advice: I would push to make all modules below `torch.utils.data.datapipes.iter` or equivalent `torchdata.datapipes.iter` private while you are still in beta. You can expose all the functionality in the top namespaces and don't have to worry about BC for the structure in the long run. For an example, have a look at how we are doing it for our new functionality, e.g. https://github.com/pytorch/vision/tree/main/torchvision/prototype/transforms

Using such an architecture would have prevented this, because the datapipe would still be exposed under `torch.utils.data.datapipes.iter` and you are free to move it around as you like.

ejguan:
I was more concern on the `SHARDING_PRIORITIES`
https://github.com/pytorch/pytorch/blob/fe0e28ab87ddda38a82efe74dd3eaf05e926f117/torch/utils/data/datapipes/iter/sharding.py#L16

pmeier:
> For `ShardingFilter`, it is exposed at parent level as `torch.utils.data.datapipes.iter.ShardingFilter`.

Huh, I guess we never looked for that given that the module it is in is public. But fair enough, I'm gonna change that on our side.

wenleix:
> How about we reference those two classes back to grouping from sharding.py and add a deprecation warning? @wenleix

Sounds good. Let me import them back . Thanks for reporting this!

PR #988, opened 2023-02-06
Add pure python platform-agnostic wheel

### LINKED ISSUE #973: Add a pure python releases for platform-agnostic support
### 🚀 The feature

Currently, we have the release coverage on `macos`, `windows`, `manylinux x86`, `macos-arm64` on Python 3.7 - 3.10. There is still at least one missing support for aarch64, the easiest way to extend our support is to add a pure python package without integration with AWSSDK.

For `PyPI` (`pip`), users would install `torchdata` in the order of platform-specific first then platform-agnostic.

I am not sure about conda. But, users should be able to specify the platform directly via `conda install`

### Motivation, pitch

Make TorchData have full coverage of support

### Alternatives

_No response_

### Additional context

_No response_

ejguan:
Fixes #973

Release workflow can be found: https://github.com/pytorch/data/actions/runs/4108506783/jobs/7089283333

### DIFF
--- a/.github/workflows/_build_test_upload.yml
+++ b/.github/workflows/_build_test_upload.yml
@@ -69,18 +69,27 @@ jobs:
           - 3.8
           - 3.9
           - "3.10"
+          - pure
+        exclude:
+          - os: macos-latest
+            python-version: pure
+          - os: windows-latest
+            python-version: pure
+          - os: macos-m1-12
+            python-version: pure
     steps:
       - name: Checkout Source Repository
         uses: actions/checkout@v3
         with:
           ref: ${{ inputs.branch }}
           submodules: recursive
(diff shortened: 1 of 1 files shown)


NivekT (commented):
Questions:
1. Will this be the default installation going forward?
2. Do we need to update the installation instructions in README?

ejguan:
> 1. Will this be the default installation going forward?

Like I said in the issue. This wheel would be the backup package when none of platform specific wheels fit into users' machine. If `pip` finds any suitable wheel for the specific platform, it should install the platform-specific one.



> 2\. Do we need to update the installation instructions in README?

No thing needs to be updated IMO.

NivekT (approved):

PR #994, opened 2023-02-07
Guarantee datapipe being reset iterator when all loops have received reset request in the dispatching process

### LINKED ISSUE #984: Fix tests for S3
### 🐛 Describe the bug

S3 test is broken due to the the datasets have been updated in the public bucket, which we don't have any control.
Test case:
https://github.com/pytorch/data/blob/807db8f8c7282b2f48b48b1e07439c119a2ba12f/test/test_remote_io.py#L256-L291

And, we previously just fix the test by updating the number of files per bucket whenever the dataset update happened. It's not a long-term solution to maintain CI. To fix it, we might choose from the following solutions:
- Only validate some files existing in the output, not the total file count from each bucket
- Use mock to simulate the result
- Add our own stable bucket for testing

I prefer the first solution for two reasons:
- We want to test the functionality provided by `_torchdata.so`. Even though mocking the result of this extension would guarantee test green, it doesn't really cover the test over the extension.
- The third option might work but it also means our own bucket will be exposed on Github, which is not ideal IMHO.

### Versions

main

ejguan:
### Changes

- ~Fix S3 Tests~ in #997
- Ignore AttributeError for ReadingService when `gc` gets involved
- Add a reset_iterator_counter to dispatching process to guarantee that the datapipe is reset when all loops have received the request. Otherwise, datapipe can be reset in the middle of iteration of the other loops.
- Remove reference of the `thread` from `Prefetcher`. This would prevent racing condition when both `finally` in generator and `reset` function are accessing the same `thread`.

### DIFF
--- a/torchdata/dataloader2/communication/eventloop.py
+++ b/torchdata/dataloader2/communication/eventloop.py
@@ -50,8 +50,6 @@ class _ResetCounter:
         assert self.cnt <= self.exp_cnt
 
     def is_reached(self) -> bool:
-        if self._reached:
-            return self._reached
         if self.cnt == self.exp_cnt:
             self._reached = True
         return self._reached
--- a/torchdata/dataloader2/communication/eventloop.py
+++ b/torchdata/dataloader2/communication/eventloop.py
@@ -35,24 +35,59 @@ __all__ = [
 ]
 
 
+class _ResetCounter:
+    exp_cnt: int
+    cnt: int
+    _reached: bool
+
+    def __init__(self, exp_cnt: int):
+        self.exp_cnt = exp_cnt
+        self.cnt = 0
+        self._reached = False
+
+    def increment(self) -> None:
+        self.cnt += 1
+        assert self.cnt <= self.exp_cnt
+
+    def is_reached(self) -> bool:
+        if self._reached:
+            return self._reached
+        if self.cnt == self.exp_cnt:
+            self._reached = True
+        return self._reached
+
+    def reset(self) -> None:
+        if self._reached:
+            self._reached = False
+            self.cnt = 0
+
+
 def MultipleDataPipesToQueuesLoop(source_datapipes, req_queues, res_queues, call_on_process_init=None):
     r"""
     Set the appropriate pipes and protocol server type, and create a loop over multiple datapipes
     with the protocol server in a non-blocking manner.
     """
     assert call_on_process_init is None, "``MultipleDataPipesToQueuesLoop`` does not support call_on_process_init"
-    assert len(source_datapipes) == len(req_queues) and len(req_queues) == len(
+    num_loops = len(source_datapipes)
+    assert num_loops == len(req_queues) and num_loops == len(
         res_queues
     ), "``MultipleDataPipesToQueuesLoop`` requires the same number of datapipes, request queues and response queues"
 
     torch.set_num_threads(1)
 
     loops = []
+    reset_iterator_counter = _ResetCounter(num_loops)
 
     for source_datapipe, req_queue, res_queue in zip(source_datapipes, req_queues, res_queues):
         loops.append(
-            _create_datapipe_queue_loop(source_datapipe, req_queue, res_queue, blocking_request_get=False)
-        )  # Non-blocking request
+            _create_datapipe_queue_loop(
+                source_datapipe,
+                req_queue,
+                res_queue,
+                blocking_request_get=False,
+                reset_iterator_counter=reset_iterator_counter,
+            )
+        )  # Non-blocking request with reset counters
 
     # Using `zip_longest` to guarantee the process is terminated only when
     # all loops have received `TerminateRequest`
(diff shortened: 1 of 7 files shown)


torchdata/dataloader2/reading_service.py
```diff
@@ -94,6 +94,12 @@ def finalize_iteration(self) -> None:
         """
         pass
 
+    def __del__(self):
+        try:
+            self.finalize()
+        except AttributeError:
+            pass
+
```
ejguan, line 102: Change 2

torchdata/dataloader2/communication/iter.py
```diff
@@ -145,6 +145,15 @@ def DataPipeBehindQueues(source_datapipe, protocol, blocking_request_get=False):
...
         elif isinstance(request, communication.messages.ResetIteratorRequest):
+            # Ensure only reset iterator once for the dispatching process
+            if reset_iterator_counter is not None:
+                reset_iterator_counter.increment()
+                while not reset_iterator_counter.is_reached():
+                    yield True
+                # Sync between loops within the dispatching process
+                source_datapipe.reset_iterator()
+                yield True
+                reset_iterator_counter.reset()
```
ejguan, line 156: Change 3

ejguan:
I will rerun all tests tmrw when PyTorch nightly has been updated.

torchdata/datapipes/iter/util/prefetcher.py
```diff
@@ -99,9 +100,7 @@ def __iter__(self):
                         time.sleep(CONSUMER_SLEEP_INTERVAL)
             finally:
                 prefetch_data.run_prefetcher = False
-                if self.thread is not None:
-                    self.thread.join()
-                    self.thread = None
+                thread.join()
```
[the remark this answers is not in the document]
NivekT: Should we set `self.thread = None` still?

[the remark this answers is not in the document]
NivekT: If `reset` was called first, wouldn't the `self.thread` be joined within `reset` anyway? Does that not join the local `thread` here as well?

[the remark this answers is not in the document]
ejguan: There is a chance that `reset` called and new thread is assigned to `self.thread`. But the previous generator was released by `gc` and call `finally`. Then, it means we are going to join the thread that is not supposed to be joined.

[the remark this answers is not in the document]
NivekT: I see, so `reset` and `__iter__` is called before the first one cleans up. Makes sense

torchdata/dataloader2/reading_service.py
NivekT: Just a comment - a little bit surprised we need this given `ProtoMultiRS`'s `finalize` seems to be catching every possible `AttributeError`? Maybe the issue is `Distributed`? Thoughts?

torchdata/dataloader2/communication/eventloop.py
```diff
@@ -35,24 +35,59 @@
...
+        self.cnt = 0
+        self._reached = False
+
+    def increment(self) -> None:
+        self.cnt += 1
+        assert self.cnt <= self.exp_cnt
+
+    def is_reached(self) -> bool:
+        if self._reached:
+            return self._reached
```
NivekT, line 54: nit: These two lines can be removed? But I guess it is slightly faster, so I'm indifferent.

torchdata/dataloader2/communication/iter.py
NivekT: Just to confirm I understand - this is to handle the situation where some workers are handling `GetNextRequest` while some are trying to `reset`? You want all `GetNext` to be done before the dispatching process executes `reset`?

NivekT (approved):
LGTM! Hope the nightly gets uploaded and all the CIs get fixed

torchdata/dataloader2/communication/iter.py
ejguan: Not workers. It only happens to the dispatching process when multiple leaf DataPipes shares the same data source (Round robin demux on the same DataPipe) in a single process.
It handles the case when some loops have received `reset` while the others haven't. We want to wait to request `getNext` until all loops have received `reset`. Otherwise, there will be a case that the data source is reset during the middle of iteration for other loops

torchdata/dataloader2/reading_service.py
ejguan: `ProtoMPRS` doesn't handle all of `AttributeError` like the `self._worker_processes` from 
https://github.com/pytorch/data/blob/98222ad72ee7a29e676646e6b3f9173576410320/torchdata/dataloader2/reading_service.py#L345

Technical speaking, I should remove those try-except clauses in `finalize` to simplify the codebase

torchdata/dataloader2/communication/eventloop.py
ejguan: Lol, you are right. I will remove them tmrw

NivekT (approved):

ejguan:
The test takes way more time to finish. However, I can't really reproduce it either on linux or mac

ejguan:
And, I am going to remove all S3 related commits. To fix S3 test, I plan to rely on https://github.com/pytorch/data/pull/997

NivekT (approved):

PR #997, opened 2023-02-08
Fix test_remote_io.py due to mutating public s3 bucket

### LINKED ISSUE #984: Fix tests for S3

ShiboXing:
creating your pull request.

Fixes #984

### Changes
- Add a private function to TestDataPipeRemoteIO in test_remote_io.py to get s3 objects count label through aws cli
- add awscli in requirements

### DIFF
--- a/test/requirements.txt
+++ b/test/requirements.txt
@@ -13,3 +13,4 @@ protobuf >= 3.9.2, < 3.20
 datasets
 graphviz
 adlfs
+awscli
--- a/test/requirements.txt
+++ b/test/requirements.txt
@@ -13,4 +13,4 @@ protobuf >= 3.9.2, < 3.20
 datasets
 graphviz
 adlfs
-awscli
+awscli>=1.27.66
--- a/test/test_remote_io.py
+++ b/test/test_remote_io.py
@@ -225,7 +225,7 @@ class TestDataPipeRemoteIO(expecttest.TestCase):
             res = list(dl)
             self.assertEqual(sorted(expected), sorted(res))
 
-    def __get_s3_cnt(self, s3_pths: list):
+    def __get_s3_cnt(self, s3_pths: list, recursive=True):
         """Return the count of the total objects collected from a list s3 paths"""
         tot_objs = set()
         for p in s3_pths:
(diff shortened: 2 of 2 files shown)


ejguan (commented):
Do you mind removing this test?
https://github.com/pytorch/data/blob/e14927f32a6cf99658f72b765425bf032116f724/test/test_remote_io.py#L337-L343

It takes too much time as it would download the whole archives. We have used the test to validated `S3FileLoader` for an image.

test/test_remote_io.py
```diff
@@ -223,24 +225,43 @@ def _filepath_fn(url):
             res = list(dl)
             self.assertEqual(sorted(expected), sorted(res))
 
+    def __get_s3_cnt(self, s3_pths: list):
```
ejguan, line 228: You might want to add an argument as `recursive`. IIRC, `fsspec` would behave like `recursive=False` but `S3FileLister` would behave like `recursive=True`.

ejguan (commented):
And, you don't need to rely on CI to execute those tests. You can run the test on your local env by
`pytest -v test/test_remote_io.py -k test_fsspec_io_iterdatapipe`
`pytest -v test/test_remote_io.py -k test_s3_io_iterdatapipe`

test/requirements.txt
```diff
@@ -13,3 +13,4 @@ protobuf >= 3.9.2, < 3.20
 datasets
 graphviz
 adlfs
+awscli>=1.27.66
```
NivekT, line 16: Question: curious why we need to specify the version here?
1. I assume it doesn't work with version prior to `1.27.66`?
2. Does it automatically try to install a version before `1.27.66`?

NivekT, line 16: Question: curious why we need to specify the version here?
1. I assume it doesn't work with version prior to `1.27.66`?
2. Does it automatically try to install a version before `1.27.66`?

ShiboXing: Hi @NivekT, actually the version doesn't matter here.

ShiboXing: Hi @NivekT, actually the version doesn't matter here as far as I know. I pinned it since I noticed that pip install -r requirements.txt would download and cache all versions of awscli, which is taking a bit long. Do you recommend reverting this or haivng some other edit advice?

ShiboXing:
Hi @NivekT, @ejguan, have you seen a version conflict similar to the one we are getting in 3.11 conda? I wonder if you have any quick fix

ejguan:
> Hi @NivekT, @ejguan, have you seen a version conflict similar to the one we are getting in 3.11 conda? I wonder if you have any quick fix

You ignore it. Not related

test/requirements.txt
ejguan: > pip install -r requirements.txt would download and cache all versions of awscli

Emmm. I am not aware of that. I think it would only download the latest version.

ShiboXing: This is what I see when I try the command locally. It only installs the latest version though. Please let me know if I need to revert this change, as well as other changes I need to make based on the test results! @ejguan @NivekT

ejguan:
@ShiboXing Can you please do a `rebase` onto `main` branch and force push? Otherwise, I think this PR is ready.

ShiboXing:
@ejguan, rebased and force-pushed!

test/requirements.txt
NivekT: @ejguan Are you fine with keeping the version number here?

ejguan: not a big deal as it's `>=` not `==`.

ejguan (approved):
Thanks for your effort to make TochData' CI green again

PR #1001, opened 2023-02-09
Insert/Extract from Serialization Wrapper only if it is necessary

### LINKED ISSUE #960: Place to insert `SerializationWrapper`
### 🐛 Describe the bug

Currently, `DataLoader2` always [wrap](https://github.com/pytorch/data/blob/2ca1fa6483e58c6428319393e1aab4c26f576bec/torchdata/dataloader2/dataloader2.py#L114) the DataPipe graph with a `SerializationWrapper` to make sure the DataPipe graph can be cloned with either `pickle` or `dill`.
When `PrototypeMultiprocessingReadingService` gets involved with DataLoader2, a `prefetch` on worker is appended to the graph. So, the graph sent to subprocess now ends with the `Prefetcher` rather than the `SerialzationWrapper`. Besides, `PrototypeMultiprocessingReadingService` now supports to move partial graph to dispatching process and worker processes. This means the partial graph might potentially not wrapped by this `SerializationWrapper`.

To fix this problem, I propose to do the following changes:
- The `clone` function would do wrap and unwrap so that the DataPipe graph remains the same structure after provided to DataLoader2.
- Wrap each partial graph with the serialization wrapper when needed
  - Dispatching process: https://github.com/pytorch/data/blob/2ca1fa6483e58c6428319393e1aab4c26f576bec/torchdata/dataloader2/reading_service.py#L236
  - Worker process: https://github.com/pytorch/data/blob/2ca1fa6483e58c6428319393e1aab4c26f576bec/torchdata/dataloader2/reading_service.py#L279

### Versions

main

ejguan:
Fixes #960

### Changes
- Remove `SerializationWrapper` from the graph. When serialization is needed like multiprocess, we attach the wrapper before serialization and remove the wrapper after deserialization
- When `clone`, we don't need to rely on `SerializationWrapper` to do all `dill` trick. We can do it within `clone` function

### DIFF
--- a/test/dataloader2/test_dataloader2.py
+++ b/test/dataloader2/test_dataloader2.py
@@ -450,7 +450,7 @@ class PrototypeMultiProcessingReadingServiceTest(TestCase):
             end_dp, reading_service=PrototypeMultiProcessingReadingService(num_workers=2, multiprocessing_context=ctx)
         )
         # Determinism and dynamic sharding
-        _assert_deterministic_dl_res(dl, [i * 4 for i in range(10)])
+        #  _assert_deterministic_dl_res(dl, [i * 4 for i in range(10)])
 
         # Non-replicable before sharding_filter
         # shuffle in dispatch process
--- a/test/dataloader2/test_dataloader2.py
+++ b/test/dataloader2/test_dataloader2.py
@@ -283,13 +283,6 @@ class DataLoader2IntegrationTest(TestCase):
     def _get_mp_reading_service():
         return MultiProcessingReadingService(num_workers=2)
 
-    @staticmethod
-    def _access_datapipe(dl):
-        """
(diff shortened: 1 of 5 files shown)


Miiira (approved):
LGTM, thanks for simplifying the layer and enhancing debuggability!

NivekT (commented):
Question: How does this impact `state_dict`? IIUC, the wrapper won't be present in serialized DataPipe?

ejguan:
> Question: How does this impact `state_dict`? IIUC, the wrapper won't be present in serialized DataPipe?

If you check the chagne, `serialize_datapipe` function would attach the wrapper and `deserialize_datapipe` function will extract the inner DadtaPipe from the wrapper.

So, wrapper will be present in the serialized state but users would not notice that.

PR #1003, opened 2023-02-09
Forward worker exceptions and exit with it

priyaramani:
When a worker thread fails in PrototypeMultiProcessingReadingService, exception was silently neglected and the thread hanged. This PR fixes that by catching and propagating exception to response queue, showing that to user and exiting the process instead of hanging.

### DIFF
--- a/test/dataloader2/test_dataloader2.py
+++ b/test/dataloader2/test_dataloader2.py
@@ -27,7 +27,6 @@ from torchdata.dataloader2 import (
     DataLoader2,
     DistributedReadingService,
     MultiProcessingReadingService,
-    PrototypeMultiProcessingReadingService,
     ReadingServiceInterface,
     SequentialReadingService,
 )
@@ -119,7 +118,7 @@ class DataLoader2Test(TestCase):
         dp = MakeMistakeDataPipe(dp)
         for worker_prefetch_cnt in [0, 5, 10]:
             for num_workers in [1, 4]:
-                rs = PrototypeMultiProcessingReadingService(
+                rs = MultiProcessingReadingService(
                     num_workers=num_workers, worker_prefetch_cnt=worker_prefetch_cnt
                 )
                 dl = DataLoader2(dp, reading_service=rs)
--- a/test/dataloader2/test_dataloader2.py
+++ b/test/dataloader2/test_dataloader2.py
@@ -122,9 +122,9 @@ class DataLoader2Test(TestCase):
                 dl = DataLoader2(dp, reading_service=rs)
                 it = iter(dl)
                 for i in range(EXCEPTION_ITERATION_NUM*num_workers):
-                    item = next(it)
+                    next(it)
                 with self.assertRaises(communication.iter.WorkerException):
-                    item = next(it)
+                    next(it)
 
     def test_dataloader2_state_dict(self) -> None:
         test_data_pipe = IterableWrapper(range(3))
--- a/test/dataloader2/test_dataloader2.py
+++ b/test/dataloader2/test_dataloader2.py
@@ -46,6 +46,8 @@ TEST_WITH_TSAN = os.getenv("PYTORCH_TEST_WITH_TSAN", "0") == "1"
 
 mp_ctx_parametrize = parametrize("ctx", mp.get_all_start_methods())
 
+DATAPIPE_ITERATION = 7
+
 
 class _ReadingServiceWrapper:
     def __init__(self, dp):
@@ -61,6 +63,18 @@ class _ReadingServiceWrapper:
     @staticmethod
     def return_one():
         return 1
+
+class MakeMistakeDataPipe(IterDataPipe):
+    def __init__(self, source_datapipe, iteration=DATAPIPE_ITERATION):
+        self.source_datapipe = source_datapipe
+        # TODO generate random num within 100 to raise expection at
+        self.iteration = iteration
+
+    def __iter__(self):
+        for i, x in enumerate(self.source_datapipe):
+            if i == self.iteration:
+                raise Exception("oops")
+            yield x
 
 
 class TestReadingService(ReadingServiceInterface):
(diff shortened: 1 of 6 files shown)


torchdata/datapipes/iter/util/prefetcher.py
```diff
@@ -93,7 +95,11 @@ def __iter__(self):
                 self.thread.start()
                 while prefetch_data.run_prefetcher:
                     if len(prefetch_data.prefetch_buffer) > 0:
-                        yield prefetch_data.prefetch_buffer.popleft()
+                        item = prefetch_data.prefetch_buffer.popleft()
+                        if isinstance(item, Exception):
+                            prefetch_data.run_prefetcher = False
+                            raise item
+                        yield item
```
ejguan, line 102: But keep this while loop working until `Exception` is popped from buffer.

It seems `run_prefetcher` needs to have three states:
- Normal
- Stop thread while loop
- Stop iter while loop

torchdata/datapipes/iter/util/prefetcher.py
```diff
@@ -74,6 +74,8 @@ def thread_worker(prefetch_data):
                     stop_iteration = True
                 except communication.iter.TerminateRequired:
                     prefetch_data.run_prefetcher = False
+                except Exception as e:
+                    prefetch_data.prefetch_buffer.append(e)
```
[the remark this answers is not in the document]
priyaramani: Thanks, added break after appending exception

torchdata/dataloader2/communication/iter.py
```diff
@@ -264,6 +277,8 @@ def __iter__(self):
                         raise communication.iter.InvalidStateResetRequired
                     if isinstance(response, communication.messages.TerminateResponse):
                         raise communication.iter.TerminateRequired
+                    if isinstance(response, communication.messages.WorkerExceptionResponse):
+                        raise communication.iter.WorkerException(f"Exception from worker {idx}: {response.exception}")
```
ejguan, line 281: The argument of `WorkerException` should be an exception rather than string. The easiest way to chain multiple exceptions should be:
```py
raise communication.iter.WorkerException(idx) from response.exception
```
In `WorkerException`, you can do:
```py
class WorkerException:
    def __init__(self, worker_id):
        super().__init__(f"Got exception from worker {idx}")
```

torchdata/dataloader2/communication/iter.py
```diff
@@ -179,6 +190,9 @@ def DataPipeBehindQueues(source_datapipe, protocol, blocking_request_get=False,
                     protocol.response_invalid_state()
                     yield True
                     break
+                except Exception as e:
+                    protocol.response_worker_exception(str(e))
```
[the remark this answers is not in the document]
priyaramani: thanks, updated!

torchdata/dataloader2/communication/iter.py
```diff
@@ -57,6 +57,14 @@ class TerminateRequired(Exception):
     pass
 
 
+class WorkerException(Exception):
+    """
+    Returned by DataPipe when there is a failure/exception from a worker process
+    """
+
+    pass
```
ejguan, line 65: nit: `pass` can be removed here

ejguan (approved):
You might want to rebase onto `main` branch again. I have already renamed `PrototypeMultiProcessingReadingService` to `MultiProcessingReadingService`

test/dataloader2/test_dataloader2.py
```diff
@@ -27,6 +27,7 @@
     DataLoader2,
     DistributedReadingService,
     MultiProcessingReadingService,
+    PrototypeMultiProcessingReadingService,
```
ejguan, line 30: No need to add it.

test/dataloader2/test_dataloader2.py
```diff
@@ -99,6 +114,21 @@ def test_dataloader2_shutdown(self) -> None:
         data_loader: DataLoader2 = DataLoader2(datapipe=test_data_pipe)
         data_loader.shutdown()
 
+    def test_worker_exception_raised(self):
+        dp = IterableWrapper(range(100)).sharding_filter()
+        dp = MakeMistakeDataPipe(dp)
+        for worker_prefetch_cnt in [0, 5, 10]:
+            for num_workers in [1, 4]:
+                rs = PrototypeMultiProcessingReadingService(
```
ejguan, line 122: Let's use `MultiProcessingReadingService`

priyaramani: Done, updated!

ejguan: Great. Can you import this PR to Phab then land it? There should be a button to do that.
image

torchdata/datapipes/iter/util/prefetcher.py
```diff
@@ -76,6 +76,9 @@ def thread_worker(prefetch_data: _PrefetchData):
                     except communication.iter.TerminateRequired:
                         prefetch_data.run_prefetcher = False
                         prefetch_data.stop_iteration = True
+                    except Exception as e:
+                        prefetch_data.prefetch_buffer.append(e)
+                        break
```
NivekT, line 81: What's the behavior without changing `Prefetcher`? It hangs and does show the error?

test/dataloader2/test_dataloader2.py
```diff
@@ -99,6 +113,19 @@ def test_dataloader2_shutdown(self) -> None:
         data_loader: DataLoader2 = DataLoader2(datapipe=test_data_pipe)
         data_loader.shutdown()
 
+    def test_worker_exception_raised(self):
+        dp = IterableWrapper(range(100)).sharding_filter()
+        dp = MakeMistakeDataPipe(dp)
+        for worker_prefetch_cnt in [0, 5, 10]:
+            for num_workers in [1, 4]:
+                rs = MultiProcessingReadingService(num_workers=num_workers, worker_prefetch_cnt=worker_prefetch_cnt)
```
NivekT, line 121: What happens if there is a DataPipe after `Prefetcher`? Does it still work properly?

e.g. `dp.map().prefetch().map()`

NivekT (commented):
LGTM overall with some questions:

Issue #978, opened 2023-01-31
`torchdata` cannot be installed from nightly channel.


pmeier:
Following the official installation instructions, running 

```bash
pip install --pre torchdata --index-url=https://download.pytorch.org/whl/nightly/cpu
```

results in an "infinite" resolving loop like

```
Looking in indexes: https://download.pytorch.org/whl/nightly/cpu
Collecting torchdata
  Using cached https://download.pytorch.org/whl/nightly/torchdata-0.6.0.dev20230131-cp38-cp38-manylinux_2_17_x86_64.manylinux2014_x86_64.whl (4.6 MB)
Requirement already satisfied: urllib3>=1.25 in /home/philip/.conda/envs/pytorch-vision-dev/lib/python3.8/site-packages (from torchdata) (1.26.14)
Requirement already satisfied: requests in /home/philip/.conda/envs/pytorch-vision-dev/lib/python3.8/site-packages (from torchdata) (2.28.2)
  Using cached https://download.pytorch.org/whl/nightly/torchdata-0.6.0.dev20230130-cp38-cp38-manylinux_2_17_x86_64.manylinux2014_x86_64.whl (4.6 MB)
Collecting torch==2.0.0.dev20230130
  Using cached https://download.pytorch.org/whl/nightly/cpu/torch-2.0.0.dev20230130%2Bcpu-cp38-cp38-linux_x86_64.whl (194.7 MB)
Collecting torchdata
  Using cached https://download.pytorch.org/whl/nightly/torchdata-0.6.0.dev20230129-cp38-cp38-manylinux_2_17_x86_64.manylinux2014_x86_64.whl (4.6 MB)
  Using cached https://download.pytorch.org/whl/nightly/torchdata-0.6.0.dev20230128-cp38-cp38-manylinux_2_17_x86_64.manylinux2014_x86_64.whl (4.6 MB)
  Using cached https://download.pytorch.org/whl/nightly/torchdata-0.6.0.dev20230127-cp38-cp38-manylinux_2_17_x86_64.manylinux2014_x86_64.whl (4.6 MB)
Collecting torch==2.0.0.dev20230127
  Using cached https://download.pytorch.org/whl/nightly/cpu/torch-2.0.0.dev20230127%2Bcpu-cp38-cp38-linux_x86_64.whl (194.6 MB)
Collecting torchdata
[...]

PR #1007, opened 2023-02-13
Make portalocker optional dependency

### LINKED ISSUE #940: Make portalocker optional to guarantee minimum dependency
### 🚀 The feature

This was a concern raised earlier in https://github.com/pytorch/data/issues/441
When users don't need to use the on-disk cache, they don't necessarily need to handle the installation of `portalocker`, especially for windows users.

### Motivation, pitch

Make TorchData more easy to install

### Alternatives

_No response_

### Additional context

_No response_

ejguan:
Fixes #940 #978

## Changes:
- Make portalocker optional
- Add `workflow_dispatch` to `pull_release.yml` to enable manual triggering on nightly release testing.
- Remove unused `test/requirements.yml`

After this PR, I need to make changes to TorchText's CI to install `portalocker`.


test/requirements.yml
```diff
@@ -1,9 +0,0 @@
-dependencies:
```
ejguan, line 1: This file was introduced to run test when we finished conda release. But, now we only do smoke test after release. So this file is not used anywhere in the code base.

wenleix (approved):

NivekT (approved):

NivekT:
If I recall correctly, this should not break any TorchText test or build. Do we need to do any confirmation before landing or is the CI here sufficient?

ejguan:
> If I recall correctly, this should not break any TorchText test or build. Do we need to do any confirmation before landing or is the CI here sufficient?

It won't break TT build. But as there are `on_disk_cache` calls in TT's dataset, it would raise Error when `portalocker` is not installed with `torchdata`. I will send a PR to fix it.

PR #1015, opened 2023-02-15
Remove fsspec from conda build test for Python 3.11

ejguan:
`fsspec` doesn't support python 3.11 for now. In order to enable TorchData conda release for python 3.11, I have to remove `fsspec` from `meta.yml` to prevent installation during building process.
See: https://anaconda.org/anaconda/fsspec/files

You can find failing release workflow in https://github.com/pytorch/data/actions/runs/4183212525


atalman (approved):
Approving since optional dependency

PR #1017, opened 2023-02-15
Laion5b dataset example

SvenDS9:
This is an example that uses Datapipes to download and preprocess the [laion5b](https://laion.ai/blog/laion-5b/)-dataset (to be more precise [this subset](https://huggingface.co/datasets/laion/laion2B-en-joined))

### Changes

- Load metadata from Huggingface and filter
- Load images from the urls
- access metadata of image and print out label and copyright information

Unfortunately loading images is still very slow as this is not done in parallel. For this we would a need a new Datapipe that supports multithreading or `asyncyio`.


examples/vision/laion5b.py
```diff
@@ -0,0 +1,74 @@
...
+
+
+# For more information about the dataset see: https://laion.ai/blog/laion-5b/
+# name of the dataset to be used
+NAME = "laion/laion2B-en-joined"
+
+
+# As the dataset is too large to store locally we use a streaming approach
+def laion2b_en(name=NAME):
+    dp = HuggingFaceHubReader(name)
```
ejguan, line 44: Noob question: Is `HuggingFaceHub` the only source of the meta data?

SvenDS9: Yes and no. The dataset with prejoined watermark and unsafe probabilities is only available on `HuggingFaceHub` (afaik). The dataset without this info is also available on [the-eye.eu](https://the-eye.eu/public/AI/cah/laion5b/metadata/laion2B-en/)

ejguan: Sounds good to me.

examples/vision/laion5b.py
```diff
@@ -0,0 +1,74 @@
...
+    dp = Filter(dp, has_no_watermark)
+    dp = Filter(dp, is_sfw)
```
ejguan, line 46: I have a nit request. Can we use functional API? like:
```py
    dp = dp.filter(has_no_watermark)
    dp = dp.filter(is_sfw)
    ...

```

examples/vision/laion5b.py
```diff
@@ -0,0 +1,74 @@
...
+    dp = dp.shuffle().sharding_filter()
+    dp = Slicer(dp, index=["TEXT", "URL"])
+    dp = Mapper(dp, fn=load_image, input_col="URL", output_col="IMAGE")  # this needs multithreading
+    dp = Filter(dp, filter_fn=image_was_loaded, input_col="IMAGE")
+    dp = Dropper(dp, "URL")
```
ejguan, line 51: Let's add a `dp.batch(batch_size)` here.

ejguan (commented):
Overall LGTM with two minor comments.

examples/vision/laion5b.py
```diff
@@ -0,0 +1,78 @@
...
+                print(f"image {i}: {label=}, {copyright_info=} ")
+            else:
+                print(f"image {i}: {label=}")
+    except PIL.UnidentifiedImageError:
+        print(f"image {i}: corrupted")
+
+
+if __name__ == "__main__":
+    i = 0
+    for batch in laion2b_en():
```
ejguan, line 75: Forget to mention, can you please use DataLoader2 as well?

```py
from torchdata.dataloader2 import DataLoader2, MultiprocessingReadingService

dp = laion2b_en()
rs = MultiprocessingReadingService(num_workers=4)
dl = DataLoader2(dp, reading_service=rs)
for batch in dl:
    ...
```

SvenDS9:
Sorry I made a mistake. I will have to reopen this.

examples/vision/laion5b.py
SvenDS9: See https://github.com/pytorch/data/pull/1034#issuecomment-1437304407

Issue #42, opened 2021-10-05
TarArchiveReader is not functioning with HTTPReader or GDriveReader


NivekT:
This issue was discovered as part of #40. The `TarArchiveReader` implementation is likely wrong:

1. An error is raised when we attempt to use `TarArchiveReader` immediately after `HTTPReader` because the HTTP stream does not support the operation `seek`:

```
file_url = "http://ai.stanford.edu/~amaas/data/sentiment/aclImdb_v1.tar.gz"
http_reader_dp = HttpReader(IterableWrapper([file_url]))
tar_dp = http_reader_dp.read_from_tar()
for fname, stream in tar_dp:
    print(f"{fname}: {stream.read()}")
```
It returns an error that looks something like this:
```
Traceback (most recent call last):
  File "/Users/ktse/data/test/test_stream.py", line 66, in <module>
    for fname, stream in tar_dp:
  File "/Users/.../data/torchdata/datapipes/iter/util/tararchivereader.py", line 62, in __iter__
    raise e
  File "/Users/.../data/torchdata/datapipes/iter/util/tararchivereader.py", line 48, in __iter__
    tar = tarfile.open(fileobj=cast(Optional[IO[bytes]], data_stream), mode=self.mode)
  File "/Users/.../miniconda3/envs/pytorch/lib/python3.9/tarfile.py", line 1609, in open
    saved_pos = fileobj.tell()
io.UnsupportedOperation: seek
```
Currently, you can work around by downloading the file in advance (or caching it with `OnDiskCacheHolderIterDataPipe`). In those cases, `TarArchiveReader` works as intended.

2. `TarArchiveReader` also doesn't work with `GDriveReader` because of the return type

```
amazon_review_url = "https://drive.google.com/uc?export=download&id=0Bz8a_Dbh9QhbaW12WVVZS2drcnM"
gdrive_reader_dp = OnlineReader(IterableWrapper([amazon_review_url]))
tar_dp = gdrive_reader_dp.read_from_tar()
```
This is because `validate_pathname_binary_tuple` requires `BufferedIOBase`. Perhaps it should accept HTTP response as well?

https://github.com/pytorch/data/blob/85d8bbe235cd58f270c17367a5577de107b0095f/torchdata/datapipes/utils/common.py#L66-L76
```
test/test_stream.py:None (test/test_stream.py)
test_stream.py:79: in <module>
    for fname, stream in tar_dp:
[...]

PR #1034, opened 2023-02-20
Add laion5b-example with Dataloader2

SvenDS9:
This is an example that uses Datapipes to download and preprocess the [laion5b](https://laion.ai/blog/laion-5b/)-dataset (to be more precise [this subset](https://huggingface.co/datasets/laion/laion2B-en-joined)). Also uses Dataloader2 for multiprocessing.

### Changes
- Load metadata from Huggingface and filter
- Load images from the urls
- access metadata of image and print out label and copyright information

Unfortunately I made a mistake while rebasing in #1017 so I had to reopen the PR.

### DIFF
--- a/examples/vision/laion5b.py
+++ b/examples/vision/laion5b.py
@@ -1,0 +1,84 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
+# All rights reserved.
+#
+# This source code is licensed under the BSD-style license found in the
+# LICENSE file in the root directory of this source tree.
+from io import BytesIO
+
+import requests
+
+from torchdata.dataloader2 import DataLoader2, MultiProcessingReadingService
+from torchdata.datapipes.iter import HuggingFaceHubReader
+
+try:
+    import PIL
+    from PIL import Image
+except ImportError:
+    PIL = None
+    Image = None
+
+
+def has_no_watermark(x):
+    return x["pwatermark"] is not None and x["pwatermark"] < 0.8
+
+
+def is_sfw(x):
+    return x["punsafe"] is not None and x["punsafe"] < 0.5
+
+
+def load_image(url):
+    try:
+        r = requests.get(url, timeout=5)
+        return Image.open(BytesIO(r.content))
+    except Exception:
+        return None
+
+
+def image_was_loaded(x):
+    return x is not None
+
+
+# For more information about the dataset see: https://laion.ai/blog/laion-5b/
+# name of the dataset to be used
+NAME = "laion/laion2B-en-joined"
+
+
+# As the dataset is too large to store locally we use a streaming approach
+def laion2b_en(name=NAME):
+    dp = HuggingFaceHubReader(name)
+    dp = dp.filter(has_no_watermark)
+    dp = dp.filter(is_sfw)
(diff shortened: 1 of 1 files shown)


SvenDS9:
To use the Dataloader2 I have to use `r.content` instead of:
https://github.com/pytorch/data/blob/a3b34a00e7d2b6694ea0d5e21fcc084080a3abae/torchdata/datapipes/iter/load/online.py#L41
This causes the following problem:
  File "\lib\multiprocessing\queues.py", line 244, in _feed
    obj = _ForkingPickler.dumps(obj)
  File "\lib\multiprocessing\reduction.py", line 51, in dumps
    cls(buf, protocol).dump(obj)
TypeError: cannot pickle '_io.BufferedReader' object
r.raw doesn't serialize very well. We probably need to change this this in HTTPReader in the future. It was originally introduced in #51 to allow using `seek` in `TarArchiveLoader` (#42)

ejguan (approved):
Thank you. LGTM

ejguan:
> To use the Dataloader2 I have to use `r.content` instead of:
> 
> https://github.com/pytorch/data/blob/a3b34a00e7d2b6694ea0d5e21fcc084080a3abae/torchdata/datapipes/iter/load/online.py#L41
> 
> 
> This causes the following problem:
> Traceback (most recent call last):
> File "\lib\multiprocessing\queues.py", line 244, in _feed
> obj = _ForkingPickler.dumps(obj)
> File "\lib\multiprocessing\reduction.py", line 51, in dumps
> cls(buf, protocol).dump(obj)
> TypeError: cannot pickle '_io.BufferedReader' object
> `r.raw` doesn't serialize very well. We probably need to change this in `HTTPReader` in the future. It was originally introduced in #51 to allow using `seek` in `TarArchiveLoader` (#42)

 I am not sure why `r.raw` is sent through mp.queue. The whole pipeline is supposed to stay in worker processes, then `r.raw` has been processed by `load_image`. The only thing being passed from worker process to main process is the `batch` of data, right?

SvenDS9:
> I am not sure why `r.raw` is sent through mp.queue. The whole pipeline is supposed to stay in worker processes, then `r.raw` has been processed by `load_image`. The only thing being passed from worker process to main process is the `batch` of data, right?

Previously load `load_image()` didn't call `PIL.Image.open()` but just returned the result of `_get_response_from_http()` or `None` if it failed to load. Therefore the batches still contained the `r.raw` objects. If the batches contain `r.content` instead, you do not run into any issues.

So one needs to be careful when using `HTTPReader` with multiprocessing and process the result before passing it to another process or else one will run into issues.

ejguan:
> Previously `load_image()` didn't call `PIL.Image.open()` but just returned the result of `_get_response_from_http()` or `None` if it failed to load. Therefore the batches still contained the `r.raw` objects. If the batches contain `r.content` instead, you do not run into any issues.

I see. IIRC, `r.content` basically would download every byte into memory, which would eliminate this problem. 

> So one needs to be careful when using `HTTPReader` with multiprocessing and process the result before passing it to another process or else one will run into issues.

Make sense. TBH, in the majority of use cases, I would expect users run all decoding operations in worker processes. So, I would assume they don't have to pass `r.raw` from worker to main.

PR #1039, opened 2023-02-22
RFC: Add snapshot to MultiProcessingReadingService

### LINKED ISSUE #1033: Accessing DataPipe state with MultiProcessingReadingService
Hi TorchData team,

I'm wondering how to access the state of the datapipe in the multi-processing context with DataLoader2 + MultiProcessingReadingService. When using no reading service, we can simply access the graph using `dataloader.datapipe`, then I can easily access the state of my datapipe using the code shown below.

However, in the multi processing case, the datapipe graph is replaced with QueueWrapper instances, and I cannot find any way to communicate with the workers to get access to the state of the data pipe (and I get the error that my StatefulIterator cannot be found on the datapipe). If I access `dl2._datapipe_before_reading_service_adapt` I do get the initial state only which makes sense since there is no state sync between the main and worker processes.

As far as I understand, this will also be a blocker for state capturing for proper DataLoader checkpointing when the MultiProcessingReadingService is being used.

Potentially, could we add a `getstate` communication primitive in `communication.messages` in order to capture the state (via getstate) of a datapipe in a worker process?
We're also open to using `sharding_round_robin_dispatch` in order to keep more information in the main process but I'm a bit confused on how to use it, if you have some sample code for me for the following case?

Running against today's master (commit a3b34a00e7d2b6694ea0d5e21fcc084080a3abae):

```python
import torchdata.datapipes as dp
from torch.utils.data.graph_settings import get_all_graph_pipes, traverse_dps
from torchdata.dataloader2 import DataLoader2, MultiProcessingReadingService


class StatefulIterator(dp.iter.IterDataPipe):
    def __init__(self, datapipe):
        self.datapipe = datapipe
        self.custom_index = 0

    def __iter__(self):
        self.custom_index = 0
        for item in self.datapipe:
            self.custom_index += 1
            yield item
        self.custom_index = 0


def get_datapipe():
    initial_data = dp.iter.IterableWrapper([1, 2, 3, 4])
    stateful_data = StatefulIterator(initial_data)
    sharded_data = stateful_data.sharding_filter()
    return sharded_data


def get_datapipe_state(datapipe):
    graph = traverse_dps(datapipe)
    all_pipes = get_all_graph_pipes(graph)
    for pipe in all_pipes:
        if hasattr(pipe, "custom_index"):
            return pipe.custom_index

    raise ValueError("This datapipe does not contain a StatefulIterator.")


def main_no_multiprocessing():
    dp = get_datapipe()
    dl2 = DataLoader2(dp)
    for item in dl2:
        print("Custom index", get_datapipe_state(dl2.datapipe))
        print("Item", item)


def main_multiprocessing():
    dp = get_datapipe()
    dl2 = DataLoader2(dp, reading_service=MultiProcessingReadingService(num_workers=4))
    for item in dl2:
        print("Custom index", get_datapipe_state(dl2.datapipe))
        print("Item", item)


if __name__ == "__main__":
    main_no_multiprocessing()
    main_multiprocessing()
```

cc: @ejguan @VitalyFedyunin @NivekT

jhoareau:
- Added communication methods for state pickling in worker processes
  - The worker datapipes are briefly paused for the state to be gathered
  - The state is then gathered by the `QueueWrapper` datapipes then `_IterateQueueDataPipes`
  - We only gather the state of worker processes and not of prefetch processes in this first iteration
- Added tests for the feature

Fixes #1033 

@NivekT @ejguan

### DIFF
--- a/test/dataloader2/test_mprs.py
+++ b/test/dataloader2/test_mprs.py
@@ -276,10 +276,44 @@ class TestMultiProcessingReadingService(TestCase):
     # cumulative_res.extend(res)
     # self.assertEqual(list(range(n_elements)), sorted(cumulative_res))
 
+    @mp_ctx_parametrize
+    @dp_parametrize
+    @parametrize("n_workers,worker_prefetch_cnt,main_prefetch_cnt", [(2, 1, 1), (4, 1, 0), (4, 0, 0)])
+    def test_reading_service_snapshot(self, ctx, dp, n_workers, worker_prefetch_cnt, main_prefetch_cnt) -> None:
+        # Functional Test: Confirms that `snapshot` does capture the state of the underlying DataPipes properly
+        rs = MultiProcessingReadingService(
+            num_workers=n_workers,
+            worker_prefetch_cnt=worker_prefetch_cnt,
+            main_prefetch_cnt=main_prefetch_cnt,
+            multiprocessing_context=ctx,
+        )
+        dl: DataLoader2 = DataLoader2(dp, reading_service=rs)
+        res = []
+        stop_index = 3
+        for i, x in enumerate(dl):
+            res.append(x)
+            if i == stop_index:
+                snapshot = dl.reading_service.snapshot()
+                break
+        self.assertEqual(
+            n_workers,
+            len(snapshot),
+            msg=f"The test is failing with '{ctx}', num_workers = {rs.num_workers}, "
+            f"worker_prefetch_cnt = {rs.worker_prefetch_cnt}, main_prefetch_cnt = {rs.main_prefetch_cnt}",
+        )
+
+        if worker_prefetch_cnt == 0 and main_prefetch_cnt == 0 and dp == dp1:
+            for snapshot_worker in snapshot:
+                self.assertAlmostEqual(
+                    stop_index + worker_prefetch_cnt,
+                    snapshot_worker["_number_of_samples_yielded"],
+                    delta=2,
+                    msg=f"The test is failing with '{ctx}', num_workers = {rs.num_workers}, "
+                    f"worker_prefetch_cnt = {rs.worker_prefetch_cnt}, main_prefetch_cnt = {rs.main_prefetch_cnt}",
+                )
+        dl.shutdown()
+
     # TODO: Implemented in an upcoming PR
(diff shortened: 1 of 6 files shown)


NivekT:
Hi @jhoareau,

Thanks for opening the issue and the PR. I have read both of them. Some thoughts (please correct me if I'm wrong):

1. Getting the serialized state of each worker's DataPipe is probably more info that you need
    - Based on your description, it sounds like you only want the length before sharding, and not other serialized state data
    - While I have a few open PRs related to checkpoint, there is a reason why we have not provided that feature yet. That is because it gets very complicated in order to work for every single DataPipe implementation, especially since we plan to allow state restoration. 
        - For example `shuffle` currently doesn't store the state of its buffer during serialization. We do not want to make a public API that returns an incomplete state of the DataPipes, at least not yet.
    - This `__len__` PR (#728) might be more relevant but we haven't prioritized that
2. As things stand, `sharding_round_robin_dispatch` is probably what you want since you aren't sharding until you have fully tokenized. Therefore, I don't think you will be benefiting from having multiple workers tokenizing the same data anyway?
    - Based on your description, each worker should return the same `length` value as well? Unless there is randomness in your transformation that I'm not aware of.
    - Let me know if there is something I missed here.
3. Another ideally solution is the have the `DataLoader2` `__len__` feature fully implemented, then you can use that in combination with inserting `dp.custom_set_length()` (which will be a custom DataIpe) after sharding to ensure that you get the desired from each worker.

jhoareau:
@NivekT Thanks for your reply. I'll focus on 1 here:
Ultimately, if I understand correctly, full state checkpointing will be necessary in some shape or form with all reading services. Right now, we are able to access the state of the datapipe when on the main process, but not in the MPRS - if we reduce the problem purely to state accessing, there's a "gap" in the current feature set.

I understand that the state as defined now is not satisfactory to you yet in terms of resuming capabilities, but technically right now the DataLoader2 does have a `state_dict`/`from_state` public API that allows us to snapshot and resume from this partial state, just not with the MPRS. I don't think this PR is introducing any extra "partiality" to the snapshotting, but do feel free to correct me on that :D

Would you accept this PR under any shape, for example if I convert the methods to protected, making this a non-public API for now? Or do you consider that the reading service is not mature enough for it yet?

PR #1044, opened 2023-02-24
Implement BatchAsyncMapper

ejguan:
### Changes

- Add `BatchAsyncMapper` to support async processing
  - Add syntax sugar to combine `batch().async_map().flatmap()`
  - `input_col`/`output_col` are added to align the behavior to `Mapper`
  - Add unit tests

### DIFF
--- a/torchdata/datapipes/iter/transform/callable.py
+++ b/torchdata/datapipes/iter/transform/callable.py
@@ -538,6 +538,16 @@ class BatchAsyncMapperIterDataPipe(IterDataPipe):
         >>> dp = dp.async_map_batches(mul_ten, 16, input_col=1, output_col=-1)
         >>> list(dp)
         [(0, 0, 0), (1, 1, 10), (2, 2, 20), (3, 3, 30), ...]
+        # Async fetching html from remote
+        >>> from aiohttp import ClientSession
+        >>> async def fetch_html(url: str, **kwargs):
+        ...     async with ClientSession() as session:
+        ...         resp = await session.request(method="GET", url=url, **kwargs)
+        ...         resp.raise_for_status()
+        ...         html = await resp.text()
+        ...     return html
+        >>> dp = IterableWrapper(urls)
+        >>> dp = dp.async_map_batches(fetch_html, 16)
     """
 
     def __new__(
--- a/torchdata/datapipes/iter/transform/callable.py
+++ b/torchdata/datapipes/iter/transform/callable.py
@@ -435,7 +435,7 @@ class _BatchAsyncMapperIterDataPipe(IterDataPipe):
         self.source_datapipe = source_datapipe
         if not inspect.iscoroutinefunction(async_fn):
             raise ValueError(f"Expected a corotine function with an async def syntax, but got a {type(async_fn)}")
-        self.async_fn = async_fn
+        self.async_fn = async_fn  # type: ignore[assignment]
         if input_col is None and output_col is not None:
             raise ValueError("`output_col` must be None when `input_col` is None.")
         self.input_col = input_col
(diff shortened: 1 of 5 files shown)


torchdata/datapipes/iter/transform/callable.py
```diff
@@ -414,3 +418,138 @@ def __len__(self) -> int:
...
+                data[self.output_col] = res
+
+            if t_flag:
+                data = tuple(data)
+
+            new_batch.append(data)
+        return new_batch
+
+
+@functional_datapipe("async_map_batches")
```
wenleix, line 501: map_batches, lol: https://docs.ray.io/en/latest/data/api/doc/ray.data.Dataset.map_batches.html#ray.data.Dataset.map_batches

torchdata/datapipes/iter/transform/callable.py
```diff
@@ -414,3 +418,138 @@ def __len__(self) -> int:
...
+
+class _BatchAsyncMapperIterDataPipe(IterDataPipe):
+    datapipe: IterDataPipe
+    async_fn: Callable
+
+    def __init__(
+        self,
+        source_datapipe: IterDataPipe,
+        async_fn: Callable,
+        input_col=None,
```
wenleix, line 431: curious if it will becomes something like `input_cols/output_cols` or it always has to be single column?

wenleix (approved):
Thanks!

torchdata/datapipes/iter/transform/callable.py
ejguan: Yeah, any naming suggestion is welcome

torchdata/datapipes/iter/transform/callable.py
```diff
@@ -414,3 +418,138 @@ def __len__(self) -> int:
...
+            - Key(s) is used for dict.
+        output_col: Index of data where result of ``fn`` is placed. ``output_col`` can be specified
+            only when ``input_col`` is not ``None``
+            - ``None`` as default to replace the index that ``input_col`` specified; For ``input_col`` with
+              multiple indices, the left-most one is used, and other indices will be removed.
+            - Integer is used for list/tuple. ``-1`` represents to append result at the end.
+            - Key is used for dict. New key is acceptable.
+        max_concurrency: Maximum concurrency to call async functions. (Default value: 32)
+
+    Example:
```
NivekT, line 524: Non-block nit: an example related to IO/download will be nice

NivekT (approved):
LGTM!

Question:
1. Will we add a `AsyncMapper` or this will be all? It seems like it will be mostly be the same except maybe instead of asking for `batch_size`, automatically set it to `batch_size = max-concurrency`.
2. Providing some async options for IO DataPipes will be nice. Or somehow use this DataPipe in combination with those as an example.

torchdata/datapipes/iter/transform/callable.py
ejguan: Current state:
- `input_col` accepts multiple elements
- `output_col` is not supported

ejguan:
> 1\. Will we add a `AsyncMapper` or this will be all? It seems like it will be mostly be the same except maybe instead of asking for `batch_size`, automatically set it to `batch_size = max-concurrency`.

I am not sure. Probably not until there is a solid request. Rather than setting `batch_size = max-concurrency`, i think it might be better to do `max-concurrency=batch_size`

> 2\. Providing some async options for IO DataPipes will be nice. Or somehow use this DataPipe in combination with those as an example.

It's hard to combine with other DataPipe because this DataPipe takes async function directly rather than rely on if prior DataPipe has async operation.
I will try to come up with an example for async io operations.

Issue #106, opened 2021-11-10
[DataPipe] Ensure all DataPipes Meet Testing Requirements


NivekT:
## 🚀 Feature

We have many tests for existing DataPipes (both in PyTorch Core and TorchData). However, over time, they have become less organized. Moreover, as the [testing requirements](https://github.com/pytorch/pytorch/wiki/DataPipes-testing-requirements) expand, older DataPipes may not have tests to cover the newly added requirements.

This issue aims to track the status of tests for all DataPipes.

## Motivation

We want to ensure test coverage for all DataPipe is complete to reduce bugs and unexpected behavior.

## Alternative

We also should create some testing templates for `IterDataPipe` and `MapDataPipe` that can be widely applied.

## `IterDataPipe` Tracker

X - Done
NA - Not Applicable
Blank - Unclear

| Name | Module | Functional Test | Reset | `__len__` | Serializable | Pickable | Graph
| ------------- | ------------- | ------------- | ------------- | ------------- | ------------- | ------------- | ------------- |
| Batcher | Core |  
| BucketBatcher | Core |
| Collator | Core |
| Concater | Core |
| Demultiplexer | Core |
| FileLister | Core |
| FileLoader | Core |
| Filter | Core |
| Forker | Core |
| Grouper | Core |
| IterableWrapper | Core |
| Mapper | Core |
| Multiplexer | Core |
| RoutedDecoder | Core |
| Sampler | Core |
| Shuffler | Core |
| StreamReader | Core |
| UnBatcher | Core |
| Zipper | Core |
| CSVDictParser | Data | 
| CSVParser | Data | 
| Cycler | Data | 
| Enumerator | Data | 
| Extractor | Data | 
| GDriveReader | Data | 
| HashChecker | Data | 
| Header | Data | 
| HttpReader | Data | 
| InMemoryCacheHolder | Data | 
| IndexAdder | Data | 
| IoPathFileLister | Data | 
| IoPathFileLoader | Data | 
| IterKeyZipper | Data | 
| JsonParser | Data | 
| LineReader | Data | 
| MapKeyZipper | Data | 
| OnDiskCacheHolder | Data | 
| OnlineReader | Data | 
| ParagraphAggregator | Data | 
| Rows2Columnar | Data | 
| SampleMultiplexer | Data | 
| Saver | Data | 
| TarArchiveReader | Data | 
| XzFileReader | Data | 
| ZipArchiveReader | Data | 

## `MapDataPipe` Tracker

[...]

PR #1052, opened 2023-02-27
Add ThreadPoolMapper

### LINKED ISSUE #1045: ThreadMapperDatapipe
### 🚀 The feature

Similar to #1044 (thanks @ejguan!) I propose to add a new datapipe that uses `ThreadPoolExecutor` to multithread mapping.

### Motivation, pitch

Speed up mapping by using Multithreading

### Alternatives

 Three possible implementations come to my mind.

- Similar to #1044 construct batches and use [Executor.map()](https://docs.python.org/3/library/concurrent.futures.html#concurrent.futures.Executor.map) and then unbatch again. One disadvantage of this is that the first item can only be returned once all operations in the batch have finished. 
This may change in a future python version see https://github.com/python/cpython/issues/74028 and https://github.com/python/cpython/pull/18566
- Only allow batches as input and apply the operation to each element in the batch. Then return the processed batch.
- Use [concurrent.futures.as_completed](https://docs.python.org/3/library/concurrent.futures.html#concurrent.futures.as_completed) with a parameter like `scheduled_tasks` to schedule a finite number of tasks. This would return results as soon as they are completed but not preserve order.

Which option do you prefer? We can of course also implement e.g. both option 1 and 3.

### Additional context

_No response_

SvenDS9:
Fixes #1045 

### Changes

- Add ThreadPoolMapper datapipe
- Add tests

### DIFF
--- a/docs/source/torchdata.datapipes.iter.rst
+++ b/docs/source/torchdata.datapipes.iter.rst
@@ -165,9 +165,9 @@ These DataPipes apply the a given function to each element in the DataPipe.
 
     BatchAsyncMapper
     BatchMapper
-    BatchThreadPoolMapper
     FlatMapper
     Mapper
+    ThreadPoolMapper
 
 Other DataPipes
 -------------------------
--- a/docs/source/torchdata.datapipes.iter.rst
+++ b/docs/source/torchdata.datapipes.iter.rst
@@ -165,6 +165,7 @@ These DataPipes apply the a given function to each element in the DataPipe.
 
     BatchAsyncMapper
     BatchMapper
+    BatchThreadPoolMapper
     FlatMapper
     Mapper
 
--- a/test/test_iterdatapipe.py
+++ b/test/test_iterdatapipe.py
@@ -1621,7 +1621,7 @@ class TestIterDataPipe(expecttest.TestCase):
             self.assertEqual(v1, exp)
             self.assertEqual(v2, exp)
 
-    def test_threadpool_map_batches(self):
+    def test_threadpool_map(self):
         batch_size = 16
         target_length = 30
         input_dp = IterableWrapper(range(target_length))
@@ -1669,7 +1669,7 @@ class TestIterDataPipe(expecttest.TestCase):
         self.assertEqual(list(range(target_length)), res_after_reset)
 
     @suppress_warnings  # Suppress warning for lambda fn
-    def test_threadpool_map_batches_tuple_list_with_col_iterdatapipe(self):
+    def test_threadpool_map_tuple_list_with_col_iterdatapipe(self):
         batch_size = 3
 
         def fn_11(d):
@@ -1775,7 +1775,7 @@ class TestIterDataPipe(expecttest.TestCase):
         _helper(lambda data: (data[0], data[1], int(data[2])), int, 2)
 
     @suppress_warnings  # Suppress warning for lambda fn
-    def test_threadpool_map_batches_dict_with_col_iterdatapipe(self):
+    def test_threadpool_map_dict_with_col_iterdatapipe(self):
         batch_size = 3
 
         def fn_11(d):
--- a/test/test_iterdatapipe.py
+++ b/test/test_iterdatapipe.py
@@ -1632,13 +1632,13 @@ class TestIterDataPipe(expecttest.TestCase):
             return data if not sum else data.sum()
 
         # Functional Test: apply to each element correctly
-        map_dp = input_dp.thread_map_batches(fn, batch_size)
+        map_dp = input_dp.threadpool_map(fn, batch_size)
         # self.assertEqual(target_length, len(map_dp))
         for x, y in zip(map_dp, range(target_length)):
             self.assertEqual(x, torch.tensor(y, dtype=torch.float))
 
         # Functional Test: works with partial function
-        map_dp = input_dp.thread_map_batches(partial(fn, dtype=torch.int, sum=True), batch_size)
+        map_dp = input_dp.threadpool_map(partial(fn, dtype=torch.int, sum=True), batch_size)
         for x, y in zip(map_dp, range(target_length)):
             self.assertEqual(x, torch.tensor(y, dtype=torch.int).sum())
 
@@ -1647,7 +1647,7 @@ class TestIterDataPipe(expecttest.TestCase):
         # self.assertEqual(target_length, len(map_dp))
 
         input_dp_nl = IDP_NoLen(range(target_length))
-        map_dp_nl = input_dp_nl.thread_map_batches((lambda x: x), batch_size)
+        map_dp_nl = input_dp_nl.threadpool_map((lambda x: x), batch_size)
         for x, y in zip(map_dp_nl, range(target_length)):
             self.assertEqual(x, torch.tensor(y, dtype=torch.float))
 
@@ -1657,7 +1657,7 @@ class TestIterDataPipe(expecttest.TestCase):
         #     len(map_dp_nl)
 
         # Test: two independent ThreadPoolExecutors running at the same time
-        map_dp_parallel = input_dp_parallel.thread_map_batches(fn, batch_size)
+        map_dp_parallel = input_dp_parallel.threadpool_map(fn, batch_size)
         for x, y, z in zip(map_dp, map_dp_parallel, range(target_length)):
             self.assertEqual(x, torch.tensor(z, dtype=torch.float))
             self.assertEqual(y, torch.tensor(z, dtype=torch.float))
@@ -1707,16 +1707,16 @@ class TestIterDataPipe(expecttest.TestCase):
                 datapipe = IterableWrapper([constr((0, 1, 2)), constr((3, 4, 5)), constr((6, 7, 8))])
                 if ref_fn is None:
                     with self.assertRaises(error):
-                        res_dp = datapipe.thread_map_batches(fn, batch_size, input_col, output_col)
+                        res_dp = datapipe.threadpool_map(fn, batch_size, input_col, output_col)
                         list(res_dp)
                 else:
-                    res_dp = datapipe.thread_map_batches(fn, batch_size, input_col, output_col)
+                    res_dp = datapipe.threadpool_map(fn, batch_size, input_col, output_col)
                     ref_dp = datapipe.map(ref_fn)
                     if constr is list:
                         ref_dp = ref_dp.map(list)
-                    self.assertEqual(list(res_dp), list(ref_dp))
+                    self.assertEqual(list(res_dp), list(ref_dp), "First test failed")
                     # Reset
-                    self.assertEqual(list(res_dp), list(ref_dp))
+                    self.assertEqual(list(res_dp), list(ref_dp), "Test after reset failed")
 
         _helper(lambda data: data, fn_n1_def, 0, 1)
         _helper(lambda data: (data[0], data[1], data[0] + data[1]), fn_n1_def, [0, 1], 2)
@@ -1825,14 +1825,14 @@ class TestIterDataPipe(expecttest.TestCase):
             datapipe = IterableWrapper([{"x": 0, "y": 1, "z": 2}, {"x": 3, "y": 4, "z": 5}, {"x": 6, "y": 7, "z": 8}])
             if ref_fn is None:
                 with self.assertRaises(error):
-                    res_dp = datapipe.thread_map_batches(fn, batch_size, input_col, output_col)
+                    res_dp = datapipe.threadpool_map(fn, batch_size, input_col, output_col)
                     list(res_dp)
             else:
-                res_dp = datapipe.thread_map_batches(fn, batch_size, input_col, output_col)
+                res_dp = datapipe.threadpool_map(fn, batch_size, input_col, output_col)
                 ref_dp = datapipe.map(ref_fn)
-                self.assertEqual(list(res_dp), list(ref_dp))
+                self.assertEqual(list(res_dp), list(ref_dp), "First test failed")
                 # Reset
-                self.assertEqual(list(res_dp), list(ref_dp))
+                self.assertEqual(list(res_dp), list(ref_dp), "Test after reset failed")
 
         _helper(lambda data: data, fn_n1_def, "x", "y")
         _helper(lambda data: data, p_fn_n1, "x", "y")
--- a/test/test_iterdatapipe.py
+++ b/test/test_iterdatapipe.py
@@ -1622,7 +1622,6 @@ class TestIterDataPipe(expecttest.TestCase):
             self.assertEqual(v2, exp)
 
     def test_threadpool_map(self):
-        batch_size = 16
         target_length = 30
         input_dp = IterableWrapper(range(target_length))
         input_dp_parallel = IterableWrapper(range(target_length))
@@ -1632,32 +1631,30 @@ class TestIterDataPipe(expecttest.TestCase):
             return data if not sum else data.sum()
 
         # Functional Test: apply to each element correctly
-        map_dp = input_dp.threadpool_map(fn, batch_size)
-        # self.assertEqual(target_length, len(map_dp))
+        map_dp = input_dp.threadpool_map(fn)
+        self.assertEqual(target_length, len(map_dp))
         for x, y in zip(map_dp, range(target_length)):
             self.assertEqual(x, torch.tensor(y, dtype=torch.float))
 
         # Functional Test: works with partial function
-        map_dp = input_dp.threadpool_map(partial(fn, dtype=torch.int, sum=True), batch_size)
+        map_dp = input_dp.threadpool_map(partial(fn, dtype=torch.int, sum=True))
         for x, y in zip(map_dp, range(target_length)):
             self.assertEqual(x, torch.tensor(y, dtype=torch.int).sum())
 
-        # this doesn't work atm
         # __len__ Test: inherits length from source DataPipe
-        # self.assertEqual(target_length, len(map_dp))
+        self.assertEqual(target_length, len(map_dp))
 
         input_dp_nl = IDP_NoLen(range(target_length))
-        map_dp_nl = input_dp_nl.threadpool_map((lambda x: x), batch_size)
+        map_dp_nl = input_dp_nl.threadpool_map(lambda x: x)
         for x, y in zip(map_dp_nl, range(target_length)):
             self.assertEqual(x, torch.tensor(y, dtype=torch.float))
 
-        # this doesn't work atm
         # __len__ Test: inherits length from source DataPipe - raises error when invalid
-        # with self.assertRaisesRegex(TypeError, r"instance doesn't have valid length$"):
-        #     len(map_dp_nl)
+        with self.assertRaisesRegex(TypeError, r"instance doesn't have valid length$"):
+            len(map_dp_nl)
 
         # Test: two independent ThreadPoolExecutors running at the same time
-        map_dp_parallel = input_dp_parallel.threadpool_map(fn, batch_size)
+        map_dp_parallel = input_dp_parallel.threadpool_map(fn)
         for x, y, z in zip(map_dp, map_dp_parallel, range(target_length)):
             self.assertEqual(x, torch.tensor(z, dtype=torch.float))
             self.assertEqual(y, torch.tensor(z, dtype=torch.float))
@@ -1670,8 +1667,6 @@ class TestIterDataPipe(expecttest.TestCase):
 
     @suppress_warnings  # Suppress warning for lambda fn
     def test_threadpool_map_tuple_list_with_col_iterdatapipe(self):
-        batch_size = 3
-
         def fn_11(d):
             return -d
 
@@ -1707,10 +1702,10 @@ class TestIterDataPipe(expecttest.TestCase):
                 datapipe = IterableWrapper([constr((0, 1, 2)), constr((3, 4, 5)), constr((6, 7, 8))])
                 if ref_fn is None:
                     with self.assertRaises(error):
-                        res_dp = datapipe.threadpool_map(fn, batch_size, input_col, output_col)
+                        res_dp = datapipe.threadpool_map(fn, input_col, output_col)
                         list(res_dp)
                 else:
-                    res_dp = datapipe.threadpool_map(fn, batch_size, input_col, output_col)
+                    res_dp = datapipe.threadpool_map(fn, input_col, output_col)
                     ref_dp = datapipe.map(ref_fn)
                     if constr is list:
                         ref_dp = ref_dp.map(list)
@@ -1776,8 +1771,6 @@ class TestIterDataPipe(expecttest.TestCase):
 
     @suppress_warnings  # Suppress warning for lambda fn
     def test_threadpool_map_dict_with_col_iterdatapipe(self):
-        batch_size = 3
-
         def fn_11(d):
             return -d
 
@@ -1825,10 +1818,10 @@ class TestIterDataPipe(expecttest.TestCase):
             datapipe = IterableWrapper([{"x": 0, "y": 1, "z": 2}, {"x": 3, "y": 4, "z": 5}, {"x": 6, "y": 7, "z": 8}])
             if ref_fn is None:
                 with self.assertRaises(error):
-                    res_dp = datapipe.threadpool_map(fn, batch_size, input_col, output_col)
+                    res_dp = datapipe.threadpool_map(fn, input_col, output_col)
                     list(res_dp)
             else:
-                res_dp = datapipe.threadpool_map(fn, batch_size, input_col, output_col)
+                res_dp = datapipe.threadpool_map(fn, input_col, output_col)
                 ref_dp = datapipe.map(ref_fn)
                 self.assertEqual(list(res_dp), list(ref_dp), "First test failed")
                 # Reset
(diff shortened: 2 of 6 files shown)


SvenDS9:
Currently I am unhappy with a few things and open for suggestions:

- the naming is confusing: the `BatchMapper` applies a function to the whole batch while the `BatchAsyncMapper` and `BatchThreadPoolMapper` apply a function to every element in the batch
- By using `flatmap` we lose access to `len`
- test coverage for ThreadPoolMapper must be improved, we should be able to copy all tests of the normal `Mapper` (+ additional tests?)

I also noticed two unrelated things, that should be improved:
- flatmap without a function and unbatch do the same thing
- Some dp like `UnBatcher` and `Grouper` (both in Core) do not overwrite the `len` function and do not have a test for `len` even though they are marked as completed here: https://github.com/pytorch/data/issues/106

ejguan:
> * the naming is confusing: the `BatchMapper` applies a function to the whole batch while the `BatchAsyncMapper` and `BatchThreadPoolMapper` apply a function to every element in the batch. Use prepositon 'in'?

Good catch. It might be better to just name them `ThreadPoolMapper` and `AsyncMapper`. WDYT? 
cc: @wenleix @NivekT 

> * By using `flatmap` we lose access to `len`

Yeah. IMHO, `len` is currently not well supported in the regime of DataPipe. To resolve it, we might want to make those two DataPipe self-containing all the `batch` and `flatmap` functionality.

> test coverage for ThreadPoolMapper must be improved, we should be able to copy all tests of the normal Mapper (+ additional tests?)

I think one extra test to validate multiple `ThreadPoolMapper` running at the same time should be good.

> * code duplicaton in `merge_batch_with_result` with AsyncMapper ->  extract function?

Yea, that would be an ideal case.

> flatmap without a function and unbatch do the same thing

I think they are not exactly same. `unbatch` support multiple-level of unnested.


> * Some dp like `UnBatcher` and `Grouper` (both in Core) do not overwrite the `len` function and do not have a test for `len` even though they are marked as completed here: [[DataPipe] Ensure all DataPipes Meet Testing Requirements #106](https://github.com/pytorch/data/issues/106)

Thank you for catching that. Updated the issue. There has been always a project in my mind to do a auto-detection on those uncovered test cases.

SvenDS9:
> Yeah. IMHO, `len` is currently not well supported in the regime of DataPipe. To resolve it, we might want to make those two DataPipe self-containing all the `batch` and `flatmap` functionality.

Alternatively we could modify lengthsetter to not only accept `int` but also Exceptions to be thrown when calling `len` to circumvent the issue, but TBH I don't particularly like this approach.

Noob question:
Is the [test_map_tuple_list_with_col_iterdatapipe](https://github.com/SvenDS9/pytorch/blob/05943712a443138497c185405b575043b2916f34/test/test_datapipe.py#L1227) actually executed? When I run it locally it fails due to mismatched types. E.g.:

```python
_helper(lambda data: (data[0], data[1], data[0] + data[1]), fn_n1_def, [0, 1], 2)
```
The return value of the lambda function is always a tuple. but in `_helper` we also test with lists.

ejguan:
> Alternatively we could modify lengthsetter to not only accept `int` but also Exceptions to be thrown when calling `len` to circumvent the issue, but TBH I don't particularly like this approach.

Actually, we can extend `flatmap` to accept a `len_fn` 

> Is the [test_map_tuple_list_with_col_iterdatapipe](https://github.com/SvenDS9/pytorch/blob/05943712a443138497c185405b575043b2916f34/test/test_datapipe.py#L1227) actually executed? When I run it locally it fails due to mismatched types.

I think it's executed. This test passes locally on my machine.

SvenDS9:
I am confused...but okay.

test/test_iterdatapipe.py
```diff
@@ -1619,6 +1621,280 @@ def _helper(input_data, exp_res, async_fn, input_col=None, output_col=None, max_
...
+                datapipe = IterableWrapper([constr((0, 1, 2)), constr((3, 4, 5)), constr((6, 7, 8))])
+                if ref_fn is None:
+                    with self.assertRaises(error):
+                        res_dp = datapipe.thread_map_batches(fn, batch_size, input_col, output_col)
+                        list(res_dp)
+                else:
+                    res_dp = datapipe.thread_map_batches(fn, batch_size, input_col, output_col)
+                    ref_dp = datapipe.map(ref_fn)
+                    if constr is list:
+                        ref_dp = ref_dp.map(list)
```
SvenDS9, line 1716: In order for this test (copied from `test_map_tuple_list_with_col_iterdatapipe`) to work I need to convert the tuple produced by the `ref_fn` to a list.
Frankly I do not understand why the test for `map `supposedly doesn't fail in Core, when I copy it here it understandbly fails due to the aforementioned reason

docs/source/torchdata.datapipes.iter.rst
```diff
@@ -164,6 +164,7 @@ These DataPipes apply the a given function to each element in the DataPipe.
 
     BatchAsyncMapper
     BatchMapper
+    BatchThreadPoolMapper
```
ejguan, line 167: BTW, I plan to change the name of `BatchAsyncMapper` to `AsyncMapper`. Let's change `BatchThreadPoolMapper` to `ThreadPoolMapper`

SvenDS9: Will do

torchdata/datapipes/iter/transform/callable.py
```diff
@@ -585,3 +592,156 @@ def __new__(
...
+        source_datapipe,
+        fn: Callable,
+        batch_size: int,
+        input_col=None,
+        output_col=None,
+        max_workers: Optional[int] = None,
+        **threadpool_kwargs,
+    ):
+        dp = source_datapipe.batch(batch_size)
+        dp = _BatchThreadPoolMapperIterDataPipe(dp, fn, input_col, output_col, max_workers, **threadpool_kwargs)
```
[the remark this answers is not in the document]
ejguan: I think it's fine as you said, a `prefetch` would help this problem. The main reason we do it rather yielding data whenever one data has been processed by `map` is that we need to guarantee the order of data.

docs/source/torchdata.datapipes.iter.rst
SvenDS9: The functional form will also be changed to `async_map`?

test/test_iterdatapipe.py
ejguan: Do you know which function is failing?

test/test_iterdatapipe.py
```diff
@@ -1619,6 +1621,280 @@ def _helper(input_data, exp_res, async_fn, input_col=None, output_col=None, max_
...
+                    self.assertEqual(list(res_dp), list(ref_dp))
+                    # Reset
+                    self.assertEqual(list(res_dp), list(ref_dp))
```
ejguan, line 1719: We should add a `msg` to `self.assertEqual` so that we will be able to know which test is failing.

docs/source/torchdata.datapipes.iter.rst
ejguan: Yeah

test/test_iterdatapipe.py
SvenDS9: > Noob question: Is the [test_map_tuple_list_with_col_iterdatapipe](https://github.com/SvenDS9/pytorch/blob/05943712a443138497c185405b575043b2916f34/test/test_datapipe.py#L1227) actually executed? When I run it locally it fails due to mismatched types. E.g.:
> 
> ```python
> _helper(lambda data: (data[0], data[1], data[0] + data[1]), fn_n1_def, [0, 1], 2)
> ```
> 
The very first test where the return value of the lambda function is always a tuple. Here this is in Line 1722
```
Expected :[[0, 1, 1], [3, 4, 7], [6, 7, 13]]
Actual   :[(0, 1, 1), (3, 4, 7), (6, 7, 13)]
```

SvenDS9: Ah! I see, thank you for clearing this up!

torchdata/datapipes/iter/transform/callable.py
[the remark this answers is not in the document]
SvenDS9: Should we expose the `buffer_size` or set it to some fixed value (2 to start working on only the next batch or default_value of `Prefetcher`)?
We probably should also make recommendations for `batch_size`. 
For both we probably need to do some performance tests.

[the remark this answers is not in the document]
ejguan: IMHO, prefetching 1 element should be sufficient as requesting the first element would lead to launching all threads for the next batch.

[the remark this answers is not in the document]
SvenDS9: Ah, I thought about adding Prefetcher before `flatmap`, but yeah the next batch is probably enough.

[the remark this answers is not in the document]
SvenDS9: I have done a small test running in the background:
```python
def wait_a_bit(x):
    time.sleep(random.random() * 3)
    return x * 10

dp = IterableWrapper(range(10000))
print("threads, batch_size, time until 10000 ")
for batch_size in map(lambda x : pow(2,x), reversed(range(7,13))):
    start = time.time()
    for _ in dp.threadpool_map(wait_a_bit,batch_size=batch_size, max_workers=128):
        pass
    end = time.time()
    print(f"{threads}, {batch_size}, {end - start}")
```
without `prefetching`: 
| threads | batch_size | time until 10000 |
|---------|------------|------------------|
| 128     | 4096       | 122.96           |
| 128     | 2048       | 126.84           |
| 128     | 1024       | 135.94           |
| 128     | 512        | 154.29           |
| 128     | 256        | 192.07           |
| 128     | 128        | 236.32           |

with `prefetching` and `buffer_size=2`:
| threads | batch_size | time until 10000   |
|---------|------------|--------------------|
| 128     | 4096       | 117.93 |
| 128     | 2048       | 120.48  |
| 128     | 1024       | 120.39 |
| 128     | 512        | 119.98    |
| 128     | 256        | 120.47 |
| 128     | 128        | 119.05 |

The theoretical minimum is at around 117s. As you can see adding a `Prefetcher` is necessary. WIth `buffer_size = 1` we get worse results if the `batch_size` is the same (or smaller) than the number of threads but similar results if it is larger. Therefore I think `buffer_size=2` is the right choice.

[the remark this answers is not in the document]
ejguan: I agree `Prefetch` is necessary. I was incorrect about using 1 would be sufficient. I guess the more complex of the map function, the larger `buffer_size` is needed.
On top of `time until 10000`, it might be good to have the time to get the first element.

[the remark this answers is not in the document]
SvenDS9: Interistingly this also depends on the use of `Prefetcher` and `batch_size` but also on `buffer_size`

without prefetching:
| threads | batch_size | time until first |
|---------|------------|------------------|
| 128     | 4096       | 4.61             |
| 128     | 2048       | 4.53             |
| 128     | 1024       | 4.86             |
| 128     | 512        | 3.34             |
| 128     | 256        | 3.20             |
| 128     | 128        | 3.01             |
| 128     | 64         | 3.01             |
| 128     | 32         | 2.93             |
| 32      | 128        | 4.48             |
| 32      | 64         | 3.36             |
| 32      | 32         | 2.99             |

with `buffer_size=2`:
| threads | batch_size | time until first |
|---------|------------|------------------|
| 128     | 2048       | 37.62            |
| 128     | 1024       | 28.21            |
| 128     | 512        | 16.08            |
| 128     | 256        | 9.45             |
| 128     | 128        | 6.06             |
| 128     | 64         | 3.91             |
| 128     | 32         | 3.00             |
| 32      | 128        | 17.34 |
| 32      | 64         | 10.44  |
| 32      | 32         | 5.86  |

with `buffer_size=1`

| threads | batch_size | time until first |
|---------|------------|------------------|
| 128     | 2048       | 27.07            |
| 128     | 1024       | 16.39            |
| 128     | 512        | 10.42            |
| 128     | 256        | 6.88             |
| 128     | 128        | 4.90             |
| 128     | 64         | 3.02             |
| 128     | 32         | 3.03             |
| 32      | 128        | 10.35 |
| 32      | 64         | 6.78  |
| 32      | 32         | 4.64 |

Without prefetching (AND without specifying an `input_col` !) we make use of the fact that the result of `executor.map` is an iterator and not the whole batch. 

So it really is a tradeoff. Maybe add an argument to enable/disable prefetching depending on whats more important and set it to `True` by default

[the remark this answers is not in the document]
ejguan: Yeah, I agree that it's just tradeoff that users need to tune based on their own requirements. 
How about we remove `prefetch` from the factory DataPipe? And, just give them suggestions in the documents. Two major reasons:
- Different scenarios require different `buffer_size`. Users can always do it by themselves to attach a `prefetch`.
- If `MultiProcessingReadingService` is used, a `prefetch` is attached automatically, which might help the result.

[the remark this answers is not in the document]
SvenDS9: Adding a `Prefetch `after `flatmap` doesn't help improve performance as the next batch is only started once all elements from previous batch have been yielded.

Instead I rewrote ThreadPoolMapper to no longer use batches but schedule only a finite number of tasks at any time. This new implementation is better in every way:
If we e.g. choose `scheduled_tasks = 2 * threads` (64 threads, 128 scheduled tasks)

- First element is yielded relatively fast (4.78s)
- throughput (time until 5000) is still fast (119s)
- Access to `len` from previous dp
- Less dps in the graph
- Elements are still returned in correct order

Higher value of `scheduled_tasks` will lead to longer waiting period until first element is yielded as tasks are executed out of order.

SvenDS9:
After https://github.com/pytorch/pytorch/pull/95067 got merged some tests are currently broken in torchvision (transforms.Identity(),  AttributeError: 'Identity' object has no attribute '__name__')

torchdata/datapipes/iter/transform/callable.py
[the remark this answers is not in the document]
ejguan: > Instead I rewrote ThreadPoolMapper to no longer use batches but schedule only a finite number of tasks at any time. This new implementation is better in almost every way:

It sounds reasonable to me. I will take a deeper look at your PR later today. For `asyncio`, I might need to revamp it as well. And, `asyncio` has another problem with nested asyncio enabled.

torchdata/datapipes/iter/transform/callable.py
```diff
@@ -585,3 +591,200 @@ def __new__(
...
+    fn: Callable
+
+    def __init__(
+        self,
+        source_datapipe: IterDataPipe,
+        fn: Callable,
+        input_col=None,
+        output_col=None,
+        scheduled_tasks: int = 64,
+        max_workers: Optional[int] = None,
```
ejguan, line 692: I think now we can consolidate those two arguments to a single one. To align with `ThreadPoolExecutor`, let's keep `max_workers` then.

torchdata/datapipes/iter/transform/callable.py
```diff
@@ -585,3 +591,200 @@ def __new__(
...
+            itr = iter(self.datapipe)
+            for _ in range(self.scheduled_tasks):
+                try:
+                    futures_deque.append(executor.submit(self._apply_fn, next(itr)))
+                except StopIteration:
+                    has_next = False
+                    break
+
+            # Yield must be hidden in closure so that the futures are submitted
+            # before the first iterator value is required.
```
ejguan, line 768: I don't fully understand this part. The futures should have been submitted to `executor`, right? Do you mean Python has not switched context to threadpool?

ejguan (commented):
I just realize this is almost a same DataPipe as `prefetch` except we provided a function to be applied.
https://github.com/pytorch/data/blob/main/torchdata/datapipes/iter/util/prefetcher.py

This can be a base class for both `prefetch` and `pin_memory` where `prefetch` takes a noop function but `pin_memory` takes a default `pin_memory_fn` and is not replicable.

Even fullsync takes the similar [approach](https://github.com/pytorch/data/blob/main/torchdata/datapipes/iter/util/distributed.py).

Would you pls take a reference from those them to see if we can consolidate them?

torchdata/datapipes/iter/transform/callable.py
SvenDS9: TBH I don't fully understand it either why hiding yield is necessary (or is it? see below).

In `executor.map()` it totally makes sense to schedule tasks before calling `next` on the iterator.
When I don't hide it in a closure the first element is only yielded once `scheduled_tasks` many tasks have been completed.

At least I thought that's what happened as time until first increased (by a lot) with when increasing `scheduled_tasks` but I cannot reproduce it right now. Will probably change it back.

ejguan: This is potentially the problem of python threading. It needs a way to switch context as all threads share the same GIL. Normally we can do `time.sleep(0)` in the main thread to make the background thread picking up jobs.

torchdata/datapipes/iter/transform/callable.py
SvenDS9: How exactly? If we want to leave `max_workers` as `None` by default (as is the case in ThreadPoolExecutor, "Changed in version 3.8: Default value of max_workers is changed to min(32, os.cpu_count() + 4")
how would we choose `scheduled_tasks`? Set it to `2*threads` if `threads` is specified and some default value otherwise?

SvenDS9:
Are you sure? `Prefetcher` parallelizes `next` on the previous dp in the background, while here `next` is blocking and we execute a function in parallel.

torchdata/datapipes/iter/transform/callable.py
SvenDS9: Changed it back to a generator as I could not reproduce the time issues. Currently without `time.sleep(0)`. Where should I add this or is it not really necessary?

ejguan:
> Are you sure? `Prefetcher` calls `next` on the previous dp in the background, while here `next` is blocking and we execute a function in parallel.

Emm you are right. Thanks for correcting me! Please let me know when the PR is ready to be reviewed. You can ask for review by doing re-request review to the reviewer section.

torchdata/datapipes/iter/transform/callable.py
ejguan: I am thinking with a default number of `max_workers`, we can always schedule the same amount of tasks to maximize the number of workers, right?

torchdata/datapipes/iter/transform/callable.py
```diff
@@ -585,3 +591,187 @@ def __new__(
...
+                    has_next = False
+                    break
+
+            while len(futures_deque) > 0:
+                if has_next:
+                    try:
+                        futures_deque.append(executor.submit(self._apply_fn, next(itr)))
+                    except StopIteration:
+                        has_next = False
+                yield futures_deque.popleft().result()
```
ejguan, line 772: We need to handle `exception()` from future here.

SvenDS9: What should we do if an exception occurs? Skip the elment?

imo to not make it too complicated I think users should do exception handling inside their function.

ejguan: Just raise the exception.
```py
future = futures_deque.popleft()
exp = future.exception()
if exp is not None:
    raise exception
yield future.result()
```

ejguan: Actually, never mind it. `future.result()` would do this job. Resolving this comment

torchdata/datapipes/iter/transform/callable.py
SvenDS9: It should be higher as `max_workers` as not all tasks will take the same amount of time. With `max_workers` = `scheduled_tasks` some threads will sit idle. I think it depends on high big the variance of needed time for the given function is. As this is something that only the user knows we can either force them to choose or set a high enough default value.

torchdata/datapipes/iter/transform/callable.py
SvenDS9: Using my simple function from before:
```python
def wait_a_bit(x):
    time.sleep(random.random() * 3)
    return x * 10
```
with random.seed(0) before every test
first `random.random() * 3` is `2.533265554575144`

we have without sleep:
| threads | scheduled_tasks | time until first   |
|---------|-----------------|--------------------|
| 64      | 10              | 2.540492057800293  |
| 64      | 64              | 2.537590265274048  |
| 64      | 100             | 2.5425333976745605 |
| 64      | 500             | 2.540599822998047  |
| 64      | 1000            | 2.5444140434265137 |


| threads | scheduled_tasks | time until 5000    |
|---------|-----------------|--------------------|
| 64      | 64              | 207.91916251182556 |
| 64      | 128             | 121.00023198127747 |
| 64      | 256             | 120.81454801559448 |
| 64      | 1024            | 120.81783604621887 |
| 64      | 2048            | 120.80706715583801 |
| 64      | 4096            | 120.81476616859436 |

with `time.sleep(0)`:

| threads | scheduled_tasks | time until first   |
|---------|-----------------|--------------------|
| 64      | 10              | 2.538611888885498  |
| 64      | 64              | 2.539947032928467  |
| 64      | 100             | 2.5512337684631348 |
| 64      | 500             | 2.5357728004455566 |
| 64      | 1000            | 2.537018299102783  |

| threads | scheduled_tasks | time until 5000    |
|---------|-----------------|--------------------|
| 64      | 64              | 207.94583225250244 |
| 64      | 128             | 121.03096175193787 |
| 64      | 256             | 120.7996666431427  |
| 64      | 1024            | 120.8266670703888  |
| 64      | 2048            | 120.84870266914368 |
| 64      | 4096            | 120.82779574394226 |

Essentially no difference. 

One thing to note though is that when the `executor` is being closed he will first finish all scheduled tasks. The only way to prevent this is to not use `with` and explicitly call `shutdown` and cancel tasks manually or use a newly added parameter `cancel_futures` (new in 3.9). tbh I do think that this is non-issue as usually one will probably exhaust the previous dp anyway and not stop in the middle?

SvenDS9:
I have made two more comments (see above) but in principal I think the PR is ready for review.

torchdata/datapipes/iter/transform/callable.py
ejguan: If that's the case, should we turn default value of `scheduled_tasks` to 128?

torchdata/datapipes/iter/transform/callable.py
ejguan: Thanks for confirming `sleep(0)` gives no impact on the perf. Then, let's keep it as it is.

ejguan (approved):
Thank you. I think the implementation and test look good now. I only have the comment to do more benchmarking so that we can give users suggestion on the setting for `max_workers` and `scheduled_tasks`.

torchdata/datapipes/iter/transform/callable.py
SvenDS9: Setting a higher default value for `scheduled_tasks` is probably a good idea.  As the impact of both `workers` and `scheduled_tasks` probably (heavily) depends on the function and the software/hardware that is used I would refrain from making too concrete suggestions. 
Instead I think adding a more detailed comment about the impact of `scheduled_tasks` and encouraging users to try out different values for optimal performance should be sufficient.

SvenDS9: I have updated the default value and added a remark about the impact of `scheduled_tasks`. We already refer to https://docs.python.org/3/library/concurrent.futures.html#concurrent.futures.ThreadPoolExecutor in the note before. Let me know if that is still insufficient.

ejguan (approved):
Thank you so much! This is awesome!

NivekT (approved):

torchdata/datapipes/iter/transform/callable.py
ejguan: I am going to land this PR shortly. I just took a deeper look at the doc of `ThreadPoolExecutor`, it seems the default `ma_workers` is `min(32, os.cpu_count() + 4)`. Giving 128 scheduled tasks might not be ideal. But, at least it's not a blocking issue, I would like to land it and fix the `scheduled_tasks` when users give us more specific performance feedback on it.

SvenDS9: The default value is stilll up to change, yes. It's only necessary as the normal ThreadPool `map` would first iterate over the whole dp to submit all tasks. To avoid this we use `scheduled_tasks`, so it only affects how many tasks will be put in a queue (submitted).

The threads itself do not switch context until they finished their current task. Therefore the number of threads and `scheduled_tasks` only interact with each other if the variance of the time needed to complete the task is high. The main thread waits on the result of the thread that is working on the next (in correct order) result. Meanwhile the other threads can already work on other already submitted tasks. At least that's what I make of the source code of ThreadPoolExecutor, see: https://github.com/python/cpython/blob/1f7eb8fe6c1a566caf934a96c6fe2877b2447d5b/Lib/concurrent/futures/thread.py#L69

ejguan: I see what do you mean. The tasks are not executed until any worker become available.

Another point I want to make is the higher `scheduled_tasks`, the more memory is needed to load data from prior `DataPipe`. This is not a problem for the common use cases, as this DataPipe should be normally used to perform I/O operations, where input is just paths/urls.

PR #1053, opened 2023-02-27
Fix links in CONTRIBUTING.md

SvenDS9:
Per title

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -86,8 +86,8 @@ the checks automatically before every `git commit`, you can install them with `p
 
 When adding a new DataPipe, there are few things that need to be done to ensure it is working and documented properly.
 
-1. Naming - please following the naming convention as
-   [described here](https://github.com/pytorch/data/blob/main/docs/source/tutorial.rst#naming).
+1. Naming - please follow the naming convention as
+   [described here](https://pytorch.org/data/main/dp_tutorial.html#naming).
 2. Testing - please add unit tests to ensure that the DataPipe is functioning properly. Here are the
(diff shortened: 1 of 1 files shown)


ejguan (approved):
Thank you for actively helping TorchData

SvenDS9:
Noob question: In the installation guide `python setup.py develop` was recently replaced by `pip install -e .` but in the new Datapipe section it is still there. Should this also be replaced here?
In addition when running `setup.py` I get error: urllib3 2.0.0a3 is installed but urllib3<1.27,>=1.21.1 is required by {'requests'} but in the requirements file we have: urllib3 >= 1.25.

ejguan:
> Noob question: In the installation guide `python setup.py develop` was recently replaced by `pip install -e .` but in the new Datapipe section it is still there. Should this also be replaced here? In addition when running `setup.py` I get error: urllib3 2.0.0a3 is installed but urllib3<1.27,>=1.21.1 is required by {'requests'} but in the requirements file we have: urllib3 >= 1.25.

Let's update it to `pip install -e ,`.

I am not aware of the problem with `setup.py`. Can you pls open an issue for it?

SvenDS9:
I don't get the issue when using `pip install -e .` and if we don't intend to use `python setup.py develop` directly I think this is a non-issue.

ejguan:
> I don't get the issue when using `pip install -e .` and if we don't intend to use `python setup.py develop` directly I think this is a non-issue.

Kind `True` since easy install is deprecating. But, I personally think the majority of users are still using `python setup.py develop`. It might be good to figure this out if more users report the same problem.

PR #1062, opened 2023-03-01
Shuffled flatmap

### LINKED ISSUE #1043: Map-style equivalent of `flatmap` - ` FlatMapperMapDataPipe`
### 🚀 The feature

The equivalent of ` FlatMapperIterDataPipe` but that would take a `MapDataPipe` as input.
 
This new datapipe would probably need two function parameters as input instead of one: 
- `get_fn` that would return the i-th element of an item in the source datapipe;
- `len_fn` that would return the length of an item in the source datapipe.

See a sketch:
```python
from typing import Callable, Any
from itertools import accumulate

from torch.utils.data import MapDataPipe, functional_datapipe


def default_getitem(d, i):
    return d[i]

@functional_datapipe("flatmap")
class FlatMapperMapDataPipe(MapDataPipe):
    def __init__(self,
                 source_dp: MapDataPipe,
                 get_fn: Callable[[Any, int], Any] = default_getitem,
                 len_fn: Callable[Any, int] = len) -> None:
        super().__init__()
        self.source_dp = source_dp
        self.get_fn = get_fn
        self.len_fn = len_fn

    def _cumsum(self):
        return accumulate(self.len_fn(d) for d in self.source_dp)

    def __len__(self):
        return sum(self.len_fn(d) for d in self.source_dp)

    def _get_source_and_item(self, idx):
        last_size = 0
        for source, size in enumerate(self._cumsum()):
            if size > idx:
                return source, idx - last_size
            last_size = size
        raise IndexError(f'idx {idx} out of range')

    def __getitem__(self, idx):
        if not isinstance(idx, int):
            raise ValueError
        source, i = self._get_source_and_item(idx)
        return self.get_fn(self.source_dp[source], i)
```

### Motivation, pitch

I'm working with mne [raw](https://mne.tools/stable/generated/mne.io.Raw.html) and [epochs](https://mne.tools/stable/generated/mne.Epochs.html) objects to handle heavy EEG recordings.
This library allows you to load only small portions of the long recordings; i.e., you instantiate `epo = mne.Epochs(..., preload=False)` only once, then the different parts of the recording are loaded from disk only when you call `X = epo[i]`.

Using an iter-style flatmap is not the best option because you typically want to randomize the order of your training examples, and a map-style is the natural approach for this kind of object.

### Alternatives

Currently, you have to write a custom datapipe that computes the index of each item according to the length of the different items of the input datapipe. 

### Additional context

_No response_

PierreGtch:
Closes #1043

### Changes

- Adds new datapipe `ShuffledFlatMapper`/`shuffled_flatmap`

### DIFF
--- a/test/test_iterdatapipe.py
+++ b/test/test_iterdatapipe.py
@@ -941,6 +941,11 @@ class TestIterDataPipe(expecttest.TestCase):
         with self.assertRaisesRegex(TypeError, "length relies on the output of its function."):
             len(shuffled_flatmapped_dp)
 
+        # __len__ when no fn specified:
+        dp = IterableWrapper([[1, 2], [], [3], [4, 5, 6, [7, 8]]])
+        dp = dp.shuffled_flatmap()
+        self.assertEqual(len(dp), 7)
+
         # Tests with .set_shuffle(False)
         # In this case, the expected behavior is similar to flatmap
 
--- a/test/test_iterdatapipe.py
+++ b/test/test_iterdatapipe.py
@@ -887,6 +887,127 @@ class TestIterDataPipe(expecttest.TestCase):
         with self.assertRaisesRegex(TypeError, "length relies on the output of its function."):
             len(flatmapped_dp)
 
+    def test_shuffled_flatmap_iterdatapipe(self):
+        source_dp = IterableWrapper(list(range(20)))
+
+        def fn(e):
+            return [e, e * 10]
+
+        # Tests with buffer_size=1
+        # In this case, the expected behavior is similar to flatmap
+
+        shuffled_flatmapped_dp = source_dp.shuffled_flatmap(fn, buffer_size=1)
+        expected_list = list(itertools.chain(*[(e, e * 10) for e in source_dp]))
+
+        self.assertEqual(expected_list, list(shuffled_flatmapped_dp))
+
+        # Funtional Test: Specify input_col
+        tuple_source_dp = IterableWrapper([(d - 1, d, d + 1) for d in range(20)])
+
+        # Single input_col
+        input_col_1_dp = tuple_source_dp.shuffled_flatmap(fn, input_col=1, buffer_size=1)
+        self.assertEqual(expected_list, list(input_col_1_dp))
+
+        # Multiple input_col
+        def mul_fn(a, b):
+            return [a - b, b - a]
+
+        input_col_2_dp = tuple_source_dp.shuffled_flatmap(mul_fn, input_col=(0, 2), buffer_size=1)
+        self.assertEqual(list(itertools.chain(*[(-2, 2) for _ in range(20)])), list(input_col_2_dp))
+
+        # shuffled_flatmap with no fn specified
+        default_dp = tuple_source_dp.shuffled_flatmap(buffer_size=1)
+        self.assertEqual(list(itertools.chain(*[(n - 1, n, n + 1) for n in range(20)])), list(default_dp))
+
+        # shuffled_flatmap with no fn specified, multiple input_col
+        default_dp = tuple_source_dp.shuffled_flatmap(input_col=(0, 2), buffer_size=1)
+        self.assertEqual(list(itertools.chain(*[(n - 1, n + 1) for n in range(20)])), list(default_dp))
+
+        # shuffled_flatmap with no fn specified, some special input
+        tuple_source_dp = IterableWrapper([[1, 2, [3, 4]], [5, 6, [7, 8]]])
+        default_dp = tuple_source_dp.shuffled_flatmap(input_col=(0, 2), buffer_size=1)
+        self.assertEqual([1, [3, 4], 5, [7, 8]], list(default_dp))
+
+        # Reset Test: reset the DataPipe after reading part of it
+        n_elements_before_reset = 5
+        res_before_reset, res_after_reset = reset_after_n_next_calls(shuffled_flatmapped_dp, n_elements_before_reset)
+
+        self.assertEqual(expected_list[:n_elements_before_reset], res_before_reset)
+        self.assertEqual(expected_list, res_after_reset)
+
+        # __len__ Test: length should be len(source_dp)*len(fn->out_shape) which we can't know
+        with self.assertRaisesRegex(TypeError, "length relies on the output of its function."):
+            len(shuffled_flatmapped_dp)
+
+        # Tests with .set_shuffle(False)
+        # In this case, the expected behavior is similar to flatmap
+
+        shuffled_flatmapped_dp = source_dp.shuffled_flatmap(fn).set_shuffle(False)
+        expected_list = list(itertools.chain(*[(e, e * 10) for e in source_dp]))
+
+        self.assertEqual(expected_list, list(shuffled_flatmapped_dp))
+
+        # Funtional Test: Specify input_col
+        tuple_source_dp = IterableWrapper([(d - 1, d, d + 1) for d in range(20)])
+
+        # Single input_col
+        input_col_1_dp = tuple_source_dp.shuffled_flatmap(fn, input_col=1, buffer_size=1)
+        self.assertEqual(expected_list, list(input_col_1_dp))
+
+        # Multiple input_col
+        input_col_2_dp = tuple_source_dp.shuffled_flatmap(mul_fn, input_col=(0, 2)).set_shuffle(False)
+        self.assertEqual(list(itertools.chain(*[(-2, 2) for _ in range(20)])), list(input_col_2_dp))
+
+        # shuffled_flatmap with no fn specified
+        default_dp = tuple_source_dp.shuffled_flatmap().set_shuffle(False)
+        self.assertEqual(list(itertools.chain(*[(n - 1, n, n + 1) for n in range(20)])), list(default_dp))
+
+        # shuffled_flatmap with no fn specified, multiple input_col
+        default_dp = tuple_source_dp.shuffled_flatmap(input_col=(0, 2)).set_shuffle(False)
+        self.assertEqual(list(itertools.chain(*[(n - 1, n + 1) for n in range(20)])), list(default_dp))
+
+        # shuffled_flatmap with no fn specified, some special input
+        tuple_source_dp = IterableWrapper([[1, 2, [3, 4]], [5, 6, [7, 8]]])
+        default_dp = tuple_source_dp.shuffled_flatmap(input_col=(0, 2)).set_shuffle(False)
+        self.assertEqual([1, [3, 4], 5, [7, 8]], list(default_dp))
+
+        # Reset Test: reset the DataPipe after reading part of it
+        n_elements_before_reset = 5
+        res_before_reset, res_after_reset = reset_after_n_next_calls(shuffled_flatmapped_dp, n_elements_before_reset)
+
+        self.assertEqual(expected_list[:n_elements_before_reset], res_before_reset)
+        self.assertEqual(expected_list, res_after_reset)
+
+        # Other tests
+
+        # Test no empty buffers:
+        with self.assertRaises(AssertionError):
+            _ = source_dp.shuffled_flatmap(buffer_size=0)
+
+        # Functional Test: No seed
+        consecutive_tuple_source_dp = IterableWrapper([(d, d + 1, d + 2) for d in range(0, 21, 3)])
+        shuffled_flatmapped_dp = consecutive_tuple_source_dp.shuffled_flatmap()
+        self.assertEqual(set(range(21)), set(shuffled_flatmapped_dp))
+
+        # Functional Test: With global seed
+        torch.manual_seed(123)
+        shuffled_flatmapped_dp = tuple_source_dp.shuffled_flatmap()
+        res = list(shuffled_flatmapped_dp)
+        torch.manual_seed(123)
+        self.assertEqual(list(shuffled_flatmapped_dp), res)
+
+        # Functional Test: Set seed
+        shuffled_flatmapped_dp = tuple_source_dp.shuffled_flatmap().set_seed(123)
+        res = list(shuffled_flatmapped_dp)
+        shuffled_flatmapped_dp.set_seed(123)
+        self.assertEqual(list(shuffled_flatmapped_dp), res)
+
+        # Reset Test:
+        shuffled_flatmapped_dp = tuple_source_dp.shuffled_flatmap()
+        n_elements_before_reset = 5
+        res_before_reset, res_after_reset = reset_after_n_next_calls(shuffled_flatmapped_dp, n_elements_before_reset)
+        self.assertEqual(5, len(res_before_reset))
+
     def test_round_robin_demux_iterdatapipe(self):
         source_dp = IterableWrapper(list(range(23)))
         with self.assertRaisesRegex(ValueError, "Expected `num_instaces`"):
--- a/test/test_iterdatapipe.py
+++ b/test/test_iterdatapipe.py
@@ -915,6 +915,16 @@ class TestIterDataPipe(expecttest.TestCase):
         input_col_1_dp = tuple_source_dp.shuffled_flatmap(fn, input_col=1, buffer_size=1)
         self.assertEqual(expected_list, list(input_col_1_dp))
 
+        # With generator as fn
+        def gen_fn(e):
+            yield e
+            yield e * 10
+
+        shuffled_flatmapped_dp = source_dp.shuffled_flatmap(gen_fn, buffer_size=1)
+        expected_list = list(itertools.chain(*[(e, e * 10) for e in source_dp]))
+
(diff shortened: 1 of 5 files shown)


PierreGtch:
Hi @ejguan , I fixed the error

ejguan (commented):
Thanks for your effort. Just left a few comments below

torchdata/datapipes/iter/transform/callable.py
```diff
@@ -168,6 +171,140 @@ def __len__(self) -> int:
...
+        elif isinstance(self.input_col, (list, tuple)):
+            args = tuple(data[col] for col in self.input_col)
+            return self.fn(*args)  # type: ignore[misc]
+        else:
+            return self.fn(data[self.input_col])  # type: ignore[misc]
+
+    def _yield_one(self, buffer: List[Iterator[T_co]]) -> Iterator[T_co]:
+        idx = self._rng.randint(0, len(buffer) - 1) if self._enabled else 0
+        try:
+            yield next(buffer[idx])
```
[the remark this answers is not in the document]
PierreGtch: Don't you think that replacing the `next` by a random choice would overcomplicate the class?
To get a full shuffle, I think I would simply do:
```python
dp = IterableWrapper([[1, 2, 3, 4], 'abcd', 'ABCD'])
dp = dp.map(IterableWrapper)
dp = dp.map(Shuffler)
dp = dp.shuffled_flatmap()
list(dp)
# ['b', 3, 'c', 'd', 'C', 'A', 'a', 2, 'B', 'D', 4, 1]
```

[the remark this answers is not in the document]
PierreGtch: I used a `yield` to avoide a case distinction for when an iterator is empty or not in the `__iter__` function bellow and reduce the amount of code. But we can use `return` indeed

torchdata/datapipes/iter/transform/callable.py
```diff
@@ -168,6 +171,140 @@ def __len__(self) -> int:
...
+    """
+    datapipe: IterDataPipe[T_co]
+    fn: Optional[Callable]
+    buffer_size: int
+    _enabled: bool
+    _seed: Optional[int]
+    _rng: random.Random
+
+    def __init__(
+        self, datapipe: IterDataPipe, fn: Optional[Callable] = None, input_col=None, buffer_size: int = 10000
```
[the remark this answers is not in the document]
PierreGtch: > This default `buffer_size` might too big to me. If the output of `flatmap` is like 1k, the total buffer size becomes 10,000,000. Can we add two sizes of buffer? One is the size of input data, one is the total size of buffer `sum(len(buffer) for buffer in buffers)`

Yes, `buffer_size=10000` is probably too big. what about 1000?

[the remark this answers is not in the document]
PierreGtch: > Can we remove `input_col` for this operation for now? It makes it too complicated. And, we can easily replace the operation with `input_col` to `dp.drop(columns).map(map_fn)`.

Agree entirely, I kept `input_col` because it was present in `FlatMapper`. Tbh, I would also get rid of the `fn` argument because the "map" part can be achieved with a `Mapper` datapipe before. However, the way I flatten the inputs is closer to the behaviour of `FlatMapper` than of `Flattener`.

By the way, the names `FlatMapper` and `Flattener` can be confusing because they don't have the same purpose...

torchdata/datapipes/iter/transform/callable.py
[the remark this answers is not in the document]
ejguan: > ```python
> dp = IterableWrapper([[1, 2, 3, 4], 'abcd', 'ABCD'])
> dp = dp.map(IterableWrapper)
> dp = dp.map(Shuffler)
> dp = dp.shuffled_flatmap()
> ```

Emmm. This is not correct. This `shuffled_flatmap` is helped to shuffle across input object. We still need to shuffle the inner object for each input. This might work though
```py
dp = IterableWrapper([[1, 2, 3, 4], 'abcd', 'ABCD'])
dp = dp.in_batch_shuffle()
dp = dp.shuffled_flatmap()
```

torchdata/datapipes/iter/transform/callable.py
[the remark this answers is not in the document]
ejguan: > > This default `buffer_size` might too big to me. If the output of `flatmap` is like 1k, the total buffer size becomes 10,000,000. Can we add two sizes of buffer? One is the size of input data, one is the total size of buffer `sum(len(buffer) for buffer in buffers)`
> 
> Yes, `buffer_size=10000` is probably too big. What about 1000?

Let's do 100 for now? There is one metric like the time for the first batch/element, which is also important. Making buffer size big causes longer time to get the first batch.

Yea, since you have already implemented, let's keep it as it is. I need to take a deeper look at if removing `input_col` from `flatmap` and this operation is doable by combining with other DataPipes like `drop`.

torchdata/datapipes/iter/transform/callable.py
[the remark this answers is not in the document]
PierreGtch: I added an example of how to completely shuffle the input in the doctoring: https://github.com/PierreGtch/TorchData/commit/cf887655a150235428b5af7aa4b7a4469bbd1573 

Is that enough?

ejguan (commented):
Thanks. I have two more comments regarding the behavior of `shuffled_flatmap`

torchdata/datapipes/iter/transform/callable.py
```diff
@@ -6,10 +6,13 @@
 
 import asyncio
 import inspect
+import random
 import warnings
 
 from typing import Callable, Hashable, Iterator, List, Optional, Set, Sized, TypeVar, Union
 
+from torch import empty as torch_empty, int64 as torch_int64
```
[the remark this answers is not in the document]
PierreGtch: no particular reason

torchdata/datapipes/iter/__init__.py
```diff
@@ -71,6 +71,7 @@
     DropperIterDataPipe as Dropper,
     FlatMapperIterDataPipe as FlatMapper,
     FlattenIterDataPipe as Flattener,
+    ShuffledFlatMapperIterDataPipe as ShuffledFlatMapper,
```
SvenDS9, line 74: Could you please also add ShuffledFlatMapper to `__all__` (line 138) and `docs/source/torchdata.datapipes.iter.rst`?

PierreGtch: done https://github.com/PierreGtch/TorchData/commit/fb1735436debb96a750fbd841798204de247f81c

torchdata/datapipes/iter/transform/callable.py
```diff
@@ -168,6 +171,146 @@ def __len__(self) -> int:
...
+            return self.fn(*args)  # type: ignore[misc]
+        else:
+            return self.fn(data[self.input_col])  # type: ignore[misc]
+
+    def _yield_one(self, buffer: List[Iterator[T_co]]) -> Iterator[T_co]:
+        idx = self._rng.randint(0, len(buffer) - 1) if self._enabled else 0
+        try:
+            yield next(buffer[idx])
+        except StopIteration:
+            buffer.pop(idx)
```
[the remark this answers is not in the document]
PierreGtch: Yes I see, but it is not easy to ensure that only one random `idx` is generated per element yielded: when you finished iterating over the elements of the input datapipe (so you are only emptying the buffer, not adding new elements to it), the `idx` you sampled might not exist in the buffer after the `pop`(s).   

I changed the code so that we don't sample unnecessary `idx` while we are iterating over the input data pipe. Would this solution be ok: https://github.com/PierreGtch/TorchData/commit/97f2aeb2b5b709c441ed3337286acdd7cde5d389 

Otherwise, we can keep an additional buffer containing the next element of every iterator in the buffer to ensure that only one `idx` per item is sampled.

torchdata/datapipes/iter/transform/callable.py
```diff
@@ -168,6 +170,155 @@ def __len__(self) -> int:
...
+
+    Args:
+        datapipe: Source IterDataPipe
+        fn: the function to be applied to each element in the DataPipe, the output must be a Sequence
+        input_col: Index or indices of data which ``fn`` is applied, such as:
+
+            - ``None`` as default to apply ``fn`` to the data directly.
+            - Integer(s) is/are used for list/tuple.
+            - Key(s) is/are used for dict.
+        buffer_size: The buffer size for sampling the iterables (default to ``10000``)
```
ejguan, line 193: Can you please update the comment for the default value?

torchdata/datapipes/iter/transform/callable.py
```diff
@@ -168,6 +170,155 @@ def __len__(self) -> int:
...
+                        idx = self._rng.randint(0, self.buffer_size - 1)
+                    except StopIteration:
+                        self._buffer.pop(idx)
+                self._buffer.append(iter(self._apply_fn(x)))
+            while self._buffer:
+                try:
+                    idx = self._rng.randint(0, len(self._buffer) - 1)
+                    yield next(self._buffer[idx])
+                except StopIteration:
+                    self._buffer.pop(idx)
```
ejguan, line 280: We can simplify this part by directly using the result from `self._apply_fn(x)` rather than `iter`. We don;t need to handle `StopIteration` as we will be able to know the size of output.
WDYT?

torchdata/datapipes/iter/transform/callable.py
PierreGtch: done

torchdata/datapipes/iter/transform/callable.py
PierreGtch: I think it is nice to allow the `fn` to return iterables and not just sequences. This datapipe will be used only in very specific cases, otherwise, it can simply be replaced by `dp.flatmap().shuffle()`. So I would keep this freedom

ejguan: Then, why not do `list` for iterables so that no need to handle `StopIteration`

PierreGtch: If I'm not mistaken, a call to `list` would load all the items in the iterable, right?
Then, we would lose the whole interest of this datapipe, and it would be simpler to just use `dp.flatmap().shuffle()`.

ejguan: > If I'm not mistaken, a call to `list` would load all the items in the iterable, right?

Yeah. And, it would only load the output of each flatmap function, right?



> Then, we would lose the whole interest of this datapipe, and it would be simpler to just use `dp.flatmap().shuffle()`.

First, it's a syntax sugar, right? Second, this DataPipe provides more randomness as `dp.flatmap().shuffle()` would do buffered shuffle over the individual output from flatmap, and this DataPipe performs shuffle at the source.

ejguan: Actually, if we want to keep `randint` being called once per iteration, it should be doable by mapping the original `idx` from `[0, len(self._buffer) - 1]` to `[0, len(self._buffer) - 2]`

torchdata/datapipes/iter/transform/callable.py
```diff
@@ -170,6 +172,155 @@ def __len__(self) -> int:
...
+
+    Args:
+        datapipe: Source IterDataPipe
+        fn: the function to be applied to each element in the DataPipe, the output must be a Sequence
+        input_col: Index or indices of data which ``fn`` is applied, such as:
+
+            - ``None`` as default to apply ``fn`` to the data directly.
+            - Integer(s) is/are used for list/tuple.
+            - Key(s) is/are used for dict.
+        buffer_size: The buffer size for sampling the iterables (default to ``100``)
```
NivekT, line 195:
```suggestion
        buffer_size: the max number of iterables this DataPipe can hold at a time (default to ``100``)
```

NivekT: Question: Does 100 seem no? Or does an iterable take too much space?

torchdata/datapipes/iter/transform/callable.py
```diff
@@ -170,6 +172,155 @@ def __len__(self) -> int:
...
 
 
+@functional_datapipe("shuffled_flatmap")
+class ShuffledFlatMapperIterDataPipe(IterDataPipe):
+    r"""
+    Applies a function over each item from the source DataPipe,
+    then collects the iterables returned in a buffer,
+    then, at every iteration, chooses at random one of the iterables in the buffer
+    and yields one item from this iterable (functional name: ``shuffled_flatmap``).
+
```
NivekT, line 182:
```suggestion

    When the buffer is full, the DataPipe will begin to yield elements from iterables within the buffer.
    New iterables will be added to the buffer once the existing ones run out of elements.
```

torchdata/datapipes/iter/transform/callable.py
```diff
@@ -170,6 +172,155 @@ def __len__(self) -> int:
...
+                self._buffer.append(iter(self._apply_fn(x)))
+            while self._buffer:
+                try:
+                    idx = self._rng.randint(0, len(self._buffer) - 1)
+                    yield next(self._buffer[idx])
+                except StopIteration:
+                    self._buffer.pop(idx)
+
+    def __len__(self) -> int:
+        raise TypeError(f"{type(self).__name__}'s length relies on the output of its function.")
```
NivekT, line 285: I think we typically omit this method when it is not feasible to calculate the length, such that a `NotImplementedError` will be raised.

NivekT (commented):
Thanks for working on this! A few comments below:

torchdata/datapipes/iter/transform/callable.py
PierreGtch: > Actually, if we want to keep `randint` being called once per iteration, it should be doable by mapping the original `idx` from `[0, len(self._buffer) - 1]` to `[0, len(self._buffer) - 2]`

Is it fine if the sampling is not uniform? (for example with this mapping:`idx % len(self._buffer)`)

PierreGtch: About `list` vs `iter`, here s a concrete example:
```python
from torch.utils.data import MapDataPipe
from torch.utils.data.datapipes.iter import FileLister
from torch.utils.data.datapipes.map import Shuffler

import mne


def read_epochs(filename) -> mne.Epochs:
    return mne.read_epochs(filename, preload=False)


class EpochsWrapper(MapDataPipe):
    def __init__(self, epochs: mne.Epochs):
        super().__init__()
        self.epochs = epochs

    def __len__(self):
        return len(self.epochs)

    def __getitem__(self, item):
        return self.epochs.get_data(item=item)


# ~/data contains 100 files of 1GB each
root = '~/data/'

datapipe = FileLister(root) \
    .map(read_epochs) \
    .in_memory_cache() \
    .map(EpochsWrapper) \
    .map(Shuffler) \
    .shuffled_flatmap()
```
This would be similar to typical pipelines I would encounter.
In this example, we don't want to load in memory all the examples in each file because this would be 100GB.

If we used `list` instead of `iter`, the 100GB would be loaded in memory when filling the buffer right? (I am honestly unsure about this affirmation, this is not a rhetorical question, please help me if I missed something)

Or do you have another idea to shuffle this kind of data?

torchdata/datapipes/iter/transform/callable.py
PierreGtch: Actually, we can know the length sometimes, I will implement it

torchdata/datapipes/iter/transform/callable.py
ejguan: Okay, I see. It makes sense to keep it as a generator to prevent loading extra data in memory ahead of time within the buffer. 

BTW, I am not sure your pipeline is correct, though. `EpochsWrapper` is not a callable to be sent to `dp.map`.

So, let's keep it as `iter` to preserve the benefit of lazy loading within `flatmap_fn`.

> Is it fine if the sampling is not uniform? (for example with this mapping:`idx % len(self._buffer)`)

That's actually a good argument not doing that.

Emmm. Then, let's keep it as it is. It should be good as long as the operations are deterministic

test/test_iterdatapipe.py
```diff
@@ -889,6 +889,132 @@ def mul_fn(a, b):
         with self.assertRaisesRegex(TypeError, "length relies on the output of its function."):
             len(flatmapped_dp)
 
+    def test_shuffled_flatmap_iterdatapipe(self):
+        source_dp = IterableWrapper(list(range(20)))
+
+        def fn(e):
+            return [e, e * 10]
```
ejguan, line 896: Can we add one more test with generator function? I believe that's the reason we still keep it as `iter`.

ejguan (approved):
Overall LGTM with one comment on the test

test/test_iterdatapipe.py
PierreGtch: Done 👌

PR #901, opened 2022-11-16
Migrate docstrings to doctests

mabu-github:
### Changes
Migrate docstrings to doctest
- torchdata/dataloader2/adapter.py
- torchdata/datapipes/iter/load/*

Use PEP8 style in code examples, e.g. add newlines between defs:
```
  from torchdata.datapipes.iter import IterableWrapper


  def filepath_fn(name: str) -> str:
      return dir_path + name


  name_to_data = {"1.txt": b"DATA1", "2.txt": b"DATA2", "3.txt": b"DATA3"}
  source_dp = IterableWrapper(sorted(name_to_data.items()))
  fsspec_saver_dp = source_dp.save_by_fsspec(filepath_fn=filepath_fn, mode="wb")
  res_file_paths = list(fsspec_saver_dp)`
```

PR #1069, opened 2023-03-06
[Doc] Fix examples and migrate to sphinx doctest (iter/util)

SvenDS9:
Similar to #901 and as discussed in #850 this migrates examples in `iter/util` directory to sphinx doctest format.

As sphinx is also able to check normal doctest-style examples (which will be enabled once `doctest_test_doctest_blocks = ""` is deleted from `conf.py`), it is not necessary to convert all examples into the testcode/testoutput block format.
As some tests need to be skipped conditionally.

### Changes

- Fixes some faulty examples
- Migrates many examples in `iter/util` to testcode/testoutput block format
- conditionally skip testing some examples
- use functional form in examples

### DIFF
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -68,7 +68,18 @@ from torchdata.datapipes.iter import IterableWrapper, FileLister, FileOpener
 
 seperator_is_slash = os.sep == "/"
 
-faulty_test = True
+io_doctest = True
+
+try:
+    import torcharrow.dtypes as dt
+except ImportError:
+    dt = None
+
+try:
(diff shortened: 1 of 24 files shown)


SvenDS9:
Using the testcode/testoutput block format the output does look slightly different but imo this it's actually an improvement
grafik

SvenDS9:
I also noticed that many datapipes in `util` also handle loading so should probably be moved to `load`. 
Imho the division into `load`, `transform` and `util` is fine but is in contrast to the documentation, where we have 10(!) different types.

PR #1070, opened 2023-03-07
CI test: check if documentation is complete

### LINKED ISSUE #1047: Add a CI test to guarantee new DataPipe has online doc
### 🚀 The feature

Per title

Example:
I implemented `PinMemory` and related documents in #1014, but forgot to link it in https://github.com/pytorch/data/blob/main/docs/source/torchdata.datapipes.iter.rst

So, I have to open a separate [PR](https://github.com/pytorch/data/pull/1046) to add the inline doc to our website. It would be great if we can go through both [IterDataPipe](https://github.com/pytorch/data/blob/main/docs/source/torchdata.datapipes.iter.rst) and [MapDataPipe](https://github.com/pytorch/data/blob/main/docs/source/torchdata.datapipes.map.rst) to see if there is any missing DataPipe in the doc.

If it's a CI test, we can easily block a PR being landed without adding/linking the inline doc

### Motivation, pitch

Improve UX when they can't find the online doc for our DataPipe

### Alternatives

_No response_

### Additional context

_No response_

SvenDS9:
Fixes #1047 

Basic but should do the trick. Not sure about the Github/actions part (never used this before)

### Changes

- Add CI test that checks if all exported dps have documentation

### DIFF
--- a/.github/scripts/check_complete_doc.py
+++ b/.github/scripts/check_complete_doc.py
@@ -15,6 +15,7 @@ def collect_init_dps(init_file_location):
             if line.startswith("__all__ "):
                 while (line := init_file.readline()) != "" and (stripped_line := line.strip()).startswith('"'):
                     init_dps.add(stripped_line.replace(",", "").replace('"', ""))
+                break
     return init_dps
 
 
@@ -30,11 +31,9 @@ def collect_rst_dps(rst_file_location):
 
 
 def compare_sets(set_a, set_b, ignore_set=None):
-    set_a_copy = set_a.copy()
+    res = set_a.difference(set_b)
     if ignore_set is not None:
-        for elem in ignore_set:
-            set_a_copy.discard(elem)
-    res = set_a_copy.difference(set_b)
+        res.difference_update(ignore_set)
     return res
 
 
@@ -55,7 +54,7 @@ def main():
         dif_rst = compare_sets(rst_set, init_set)
 
         for elem in dif_init:
-            print(f"{elem} is missing from {rst_path}")
+            print(f"Please add {elem} to {rst_path}")
             exit_code = 1
         for elem in dif_rst:
             print(f"{elem} is present in {rst_path} but not in {init_path}")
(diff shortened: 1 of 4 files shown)


.github/scripts/check_complete_doc.py
```diff
@@ -0,0 +1,68 @@
...
+                    rst_dps.add(stripped_line)
+    return rst_dps
+
+
+def compare_sets(set_a, set_b, ignore_set=None):
+    set_a_copy = set_a.copy()
+    if ignore_set is not None:
+        for elem in ignore_set:
+            set_a_copy.discard(elem)
+    res = set_a_copy.difference(set_b)
```
ejguan, line 37: nit: Wht not:
```py
res = set_a.difference(set_b)
if ignore_set is not None:
    res = res.difference(ignore_set)
```

.github/scripts/check_complete_doc.py
```diff
@@ -0,0 +1,68 @@
...
+        rst_path = os.path.join(docs_source_folder, "torchdata.datapipes." + target + ".rst")
+
+        init_set = collect_init_dps(init_path)
+        rst_set = collect_rst_dps(rst_path)
+
+        dif_init = compare_sets(init_set, rst_set, ignore_set)
+        dif_rst = compare_sets(rst_set, init_set)
+
+        for elem in dif_init:
+            print(f"{elem} is missing from {rst_path}")
```
ejguan, line 58: Can we change the log to `Please add {elem} to {rst_path}`?

ejguan (commented):
Overall LGTM with two minor comments

mabu-github:
Imho using the abstract syntax tree would be a bit safer.

.github/scripts/check_complete_doc.py
SvenDS9: Much better, will do

ejguan:
Thank you @SvenDS9 

> Imho using the abstract syntax tree would be a bit safer.

Does ast work with non-compiled package? If so, it's definitely better. I think it's fine to keep it as it is as this workflow is not that complicated.

NivekT (approved):
Thank you so much!

ejguan:
The failed tests are caused by https://github.com/pytorch/pytorch/pull/95067

PR #1009, opened 2023-02-13
Officially graduate ProtypeMPRS to MPRS

ejguan:
Summary:
Fixes: https://github.com/pytorch/data/issues/932

- Convert all references from `ProtypeMPRS` to `MPRS`
- Remove usage of `MPRS`
- [ ] Update colab
- [ ] Add BC-breaking notes

Differential Revision: D43245136

PR #1074, opened 2023-03-08
HuggingFaceHubReader better align with default values of datasets.load_dataset()

SvenDS9:
### Changes

- Change default value of HuggingFaceHubReader parameters
- Remove legacy `source_datapipe` attribute

A quick search on Github reveals no usage of this dp outside torchdata.

### BC-breaking note

**0.6.0**: Previously `HuggingFaceHubReader("lhoestq/demo1")` called 
`datasets.load_dataset(path="lhoestq/demo1", streaming=True, split="train", revision="main")`

**PR**: This PR changes default arguments of `split` and `revision` to the default values of `datasets.load_dataset()`:  
Now `HuggingFaceHubReader("lhoestq/demo1")` calls 
`datasets.load_dataset(path="lhoestq/demo1", streaming=True)`

As the default value of `split` and `train` are both `None` this is equivalent to:
`datasets.load_dataset(path="lhoestq/demo1", streaming=True, split=None, revision=None)`

In some cases (depending on the dataset) this **will change the data retrieved from HuggingFace**.

`streaming` is still set to `True`.

### DIFF
--- a/torchdata/datapipes/iter/load/huggingface.py
+++ b/torchdata/datapipes/iter/load/huggingface.py
@@ -5,7 +5,6 @@
 # LICENSE file in the root directory of this source tree.
 
 
-import warnings
 from typing import Any, Iterator, Tuple
 
 from torchdata.datapipes.iter import IterDataPipe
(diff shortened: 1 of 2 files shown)


SvenDS9:
The failures in `torchvision` are unrelated.

torchdata/datapipes/iter/load/huggingface.py
```diff
@@ -62,8 +57,6 @@ class HuggingFaceHubReaderIterDataPipe(IterDataPipe[Tuple[str, StreamWrapper]]):
 
     """
 
-    source_datapipe: IterDataPipe[str]
```
NivekT, line 65: Interesting how this was never used but we didn't notice... 

Thanks for spotting this!

NivekT (approved):
LGTM with one comment!

We will need to add a backward compatibility breaking note to the description of this PR. We can keep it relatively brief.

Past examples:
- https://github.com/pytorch/data/pull/746
- https://github.com/pytorch/data/pull/1009

SvenDS9:
I have added a BC note.

PR #1076, opened 2023-03-11
Prevent overwrite

### LINKED ISSUE #886: Protecting users from overwriting special DataPipe methods
### 🚀 The feature

There are some built-in methods that exist across DataPipes (most commonly `reset`, and potentially `full_stop`, `resume` in the near future). Those methods are often used in ways that are hidden/not obvious to the users. Notably, `reset` is called when `iter(datapipe)` is called.

Currently, it is possible to define a new custom DataPipe with attributes using those names. While we cannot completely prohibit that, we should consider to rename those methods to different names (e.g. `_reset` or `__reset`) to decrease the probability of accidental overwrites of those names. Suggestions are welcomed

### Motivation, pitch

Protect users from overwriting special DataPipe methods

### Alternatives

Suggestions are welcomed

### Additional context

_No response_

carlosschmidtmuniz:
Fixes #886 

### Changes
Add `@final` decorators to reset and resume methods in datapipies. Mypy gives a soft hint when overriding these methods.

CLA is in progress.

In collaboration with @thomasdick

### DIFF
--- a/torchdata/dataloader2/communication/eventloop.py
+++ b/torchdata/dataloader2/communication/eventloop.py
@@ -9,7 +9,6 @@ import threading
 import time
 
 from itertools import zip_longest
-from typing import final
 
 import torch
 
@@ -56,7 +55,6 @@ class _ResetCounter:
             self._reached = True
         return self._reached
 
-    @final
     def reset(self) -> None:
         if self._reached:
             self._reached = False
--- a/torchdata/dataloader2/communication/eventloop.py
+++ b/torchdata/dataloader2/communication/eventloop.py
@@ -9,6 +9,7 @@ import threading
 import time
 
 from itertools import zip_longest
+from typing import final
 
 import torch
 
@@ -55,6 +56,7 @@ class _ResetCounter:
             self._reached = True
         return self._reached
 
+    @final
     def reset(self) -> None:
         if self._reached:
             self._reached = False
--- a/torchdata/dataloader2/communication/iter.py
+++ b/torchdata/dataloader2/communication/iter.py
@@ -267,6 +267,7 @@ class QueueWrapper(NonBlocking):
                 if NonBlocking.not_available_hook is not None:
                     NonBlocking.not_available_hook()
 
+    @final
     def resume(self):
         self.protocol.request_resume()
         while True:
--- a/torchdata/dataloader2/communication/iter.py
+++ b/torchdata/dataloader2/communication/iter.py
@@ -10,7 +10,7 @@ import warnings
 
 from collections import deque
 from functools import partial
-from typing import Callable, Deque, List, final
+from typing import Callable, Deque, List
 
 from torch.utils.data import IterDataPipe
 from torchdata._utils import ExceptionWrapper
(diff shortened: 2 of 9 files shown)


ejguan (commented):
I think PR looks good to me. Do you mind adding some tests to validate overriding is not permitted.

carlosschmidtmuniz:
> I think PR looks good to me. Do you mind adding some tests to validate overriding is not permitted.

As far as we know the `@final` notation does not prohibit overriding but only provides hints (only with mypy) that the function should not be overwritten.

ejguan:
> > I think PR looks good to me. Do you mind adding some tests to validate overriding is not permitted.
> 
> As far as we know the `@final` notation does not prohibit overriding but only provides hints (only with mypy) that the function should not be overwritten.

Emmm. You are right. There is no runtime check for `final` until python 3.11

torchdata/dataloader2/communication/iter.py
```diff
@@ -263,6 +263,7 @@ def pause(self):
                 if NonBlocking.not_available_hook is not None:
                     NonBlocking.not_available_hook()
 
+    @final
```
NivekT, line 266: nit: Don't think this is needed either? Since no one should be inheriting from this class (maybe developers/ourselves, but we are doing that)?

I wouldn't mind if you decide to keep this

torchdata/dataloader2/communication/iter.py
```diff
@@ -336,6 +337,7 @@ def __iter__(self):
                         self.datapipes[idx].protocol.request_next()
                     yield response.value
 
+    @final
```
NivekT, line 340: same as above

torchdata/dataloader2/dataloader2.py
```diff
@@ -95,6 +95,7 @@ def _pause(self) -> None:
         self.dataloader._pause()
         self.limit_counter = 0
 
+    @final
```
NivekT, line 98: nit: not a strong opinion, but I would guess if someone is inheriting from DataLoader2 to add things, they probably are intentionally rewriting some features

NivekT (commented):
Thanks for adding these. It would be nice if this can work without relying on a linter. I am unsure how many TorchData users use linter, but it is definitely a step in the right direction.

Some attributes can still be overwritten and ideally we should touch on that as well (but I can't think of a way).

torchdata/dataloader2/communication/eventloop.py
```diff
@@ -55,6 +56,7 @@ def is_reached(self) -> bool:
             self._reached = True
         return self._reached
 
+    @final
```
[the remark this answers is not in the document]
carlosschmidtmuniz: Removed it with the latest commit.

torchdata/dataloader2/dataloader2.py
carlosschmidtmuniz: I agree. Removed it with the latest commit

torchdata/dataloader2/communication/iter.py
carlosschmidtmuniz: In the end I decided to only keep the `@final` of the datapipes.

torchdata/dataloader2/communication/iter.py
carlosschmidtmuniz: see above :)

NivekT (approved):

ejguan (commented):
Could you pls fix lint Error? You can refer to https://github.com/pytorch/data/blob/main/CONTRIBUTING.md#code-style

PR #1086, opened 2023-03-15
Fix pin_memory_fn for NamedTuple

### LINKED ISSUE #1085: `pin_memory_fn` returns `None` when `NamedTuple` is provied
### 🐛 Describe the bug

Per title

Min-repro:
```py
from typing import NamedTuple
import torch
from torchdata.datapipes.utils import pin_memory_fn

class NamedTensors(NamedTuple):
    a: torch.Tensor
    b: torch.Tensor

data = NamedTensors(torch.tensor(1), torch.tensor(2))

print(pin_memory_fn(data))  # None
```

To fix it, we just need to add a `return` https://github.com/pytorch/data/blob/fea20d416cd40e23ee949e09f0f5a7858fcc3542/torchdata/datapipes/utils/pin_memory.py#L30

### Versions

main branch

ejguan:
Fixes #1085

Per title. And, even though I can add a test, this test won't be executed as we don't have a GPU CI machine yet.

I have tested on my local machine though

### DIFF
--- a/test/test_iterdatapipe.py
+++ b/test/test_iterdatapipe.py
@@ -13,7 +13,7 @@ import warnings
 
 from collections import defaultdict
 from functools import partial
-from typing import Dict
+from typing import Dict, NamedTuple
 
 import expecttest
 import torch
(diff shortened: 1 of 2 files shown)


NivekT (approved):
LGTM!

PR #1096, opened 2023-03-17
Properly cleanup processes and queues for MPRS and Fix `pause` for prefetch (#1075)

ejguan:
Summary:
Fixes issue about `MPRS.finalize` when `dataloader2.shutdown()` is called

### Changes

- DataLoader2 should always clean up `datapipe_iter` at shutdown
- Guard `MPRS` to finalize once
- Fix the problem of `ConnectionError` when DataLoader early exits
  - This is caused by `queue` is joined when main/worker/dispatching process exits. No more request/response can be passed across processes.
    - Consumer process shouldn't join the `req_queue` at exit to make sure producer process can still access the remaining request. And, consumer will close `req_queue` after clean up to prevent any further request sent to queue.
    - Produce process shouldn't join the `res_queue` at exit to make sure consumer process can still access response. And, producer will close `res_queue` after clean up to prevent any further response sent to queue.
       - Main (Consumer) <-> Worker (Producer)
       - Worker (Consumer) -> Dispatching (Producer)
- Fix `pause` API for DataLoader2
    - ~Invoke `pause` lazily until the `limit+1` iteration is reached to align with python's iterator behavior.~ (**Breaking internal tests**)
    - Make `prefetch.pause` blocking unless there might be potential racing issue. Main thread is paused but prefetch worker is still trying to fetch data from `iter`.
- Add tests to validate

Pull Request resolved: https://github.com/pytorch/data/pull/1075

Differential Revision: D44168655

Pulled By: ejguan

### DIFF
--- a/test/dataloader2/test_mprs.py
+++ b/test/dataloader2/test_mprs.py
@@ -9,7 +9,10 @@ import multiprocessing as mp
 import unittest
 from unittest import TestCase
 
-from torch.testing._internal.common_utils import instantiate_parametrized_tests, parametrize
+from torch.testing._internal.common_utils import instantiate_parametrized_tests, parametrize, subtest
+
+from torch.utils.data.datapipes.iter.sharding import SHARDING_PRIORITIES
+
 from torchdata.dataloader2 import DataLoader2, DataLoader2Iterator, MultiProcessingReadingService
 from torchdata.datapipes.iter import IterableWrapper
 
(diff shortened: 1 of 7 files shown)


ejguan:
This is the reland of #1075. To fix the internal test failures, I have to remove the optional `pause` change to `DataLoader2`.

NivekT (approved):

PR #1098, opened 2023-03-17
Fix DataLoader2 `seed = 0` bug and clean up unused codes

### LINKED ISSUE #1097: Potential bug in `DataLoader2` when `seed = 0`
### 🐛 Describe the bug

If you try to set the `seed = 0` as such:

```python
dp = IterableWrapper(range(20)).shuffle()
rs = MultiProcessingReadingService(num_workers=2)

dl = DataLoader2(datapipe=dp, reading_service=rs)
dl.seed(0)
```

The seed will not be set because this line returns `False`. I think the true intention is to skip over `None`, but `0` also gets skipped as it is currently implemented.

https://github.com/pytorch/data/blob/7d9747588173f38e3ce0d5c58ba3746dacb4438f/torchdata/dataloader2/dataloader2.py#L203

I suppose we can say `seed = 0` is invalid but it is probably better to check for `self._seed is not None` instead.

### Versions

main

NivekT:
* __->__ #1098

Fixes #1097

1. Fixes the bug in #1097
2. Minor clean up of unused code 

Differential Revision: [D44184683](https://our.internmc.facebook.com/intern/diff/D44184683)

### DIFF
--- a/torchdata/dataloader2/dataloader2.py
+++ b/torchdata/dataloader2/dataloader2.py
@@ -7,7 +7,6 @@
 
 import warnings
 
-from dataclasses import dataclass
 from typing import Any, Dict, Generic, Iterable, Iterator, Optional, TypeVar, Union
 
 from torchdata.dataloader2.adapter import Adapter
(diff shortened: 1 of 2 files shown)


NivekT:
@nivekt has imported this pull request.  If you are a Meta employee, you can view this diff [on Phabricator](https://www.internalfb.com/diff/D44184683).

ejguan (approved):
Thanks for digging out this issue and fix it.

PR #1100, opened 2023-03-20
Fix `worker_init_fn` to update DataPipe graph and move worker prefetch to the end of Worker pipeline

### LINKED ISSUE #1083: Return value of `worker_init_fn` is ignored by MPRS
### 🐛 Describe the bug

```
def abc(x):
    return -1

def worker_init(dp, worker_info):
    return dp.map(abc)
    
pipe = dp.iter.IterableWrapper(range(10))

rs = MultiProcessingReadingService(num_workers=1, worker_init_fn=worker_init)
dl = DataLoader2(pipe, reading_service=rs)

for x in dl:
    print(x)
```

Output:
```
0
1
2
3
4
5
6
7
8
9
```

Expected:
```
-1
-1
-1
-1
-1
-1
-1
-1
-1
-1
```

Notice that `worker_init_fn` MUST return a datapipe, and `process_init_fn` also accounts for that, but MPRS simply ignores it. The signature of the function heavily implies that `worker_init_fn` should be able to modify the datapipe in a per-worker manner; and arguably this would be a useful feature.

### Versions

main & torch nightly

ejguan:
Fixes #1083

### Changes

- Fix worker loop to update the DataPipe graph with `worker_init_fn`
  - Add corresponding Tests
- Since `worker_init_fn` function might attach new DataPipe to worker graph, guarantee `prefetch` in worker attached to the end of the pipeline in worker process

### DIFF
--- a/torchdata/dataloader2/communication/eventloop.py
+++ b/torchdata/dataloader2/communication/eventloop.py
@@ -65,6 +65,13 @@ def MultipleDataPipesToQueuesLoop(source_datapipes, req_queues, res_queues, proc
     r"""
     Set the appropriate pipes and protocol server type, and create a loop over multiple datapipes
     with the protocol server in a non-blocking manner.
+
+    Args:
+        source_datapipe: DataPipe being iterated in the dispatching process
+        req_queue: Multiprocessing queue providing requests from the worker process
+        res_queue: Multiprocessing queue sending results to the worker process
+        process_name: The name of process (used for logging and exception handling)
+        call_on_process_init: Not allowed by dispatching process for now.
     """
     assert call_on_process_init is None, "``MultipleDataPipesToQueuesLoop`` does not support call_on_process_init"
     num_loops = len(source_datapipes)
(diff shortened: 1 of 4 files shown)


torchdata/dataloader2/communication/eventloop.py
```diff
@@ -109,7 +109,7 @@ def DataPipeToQueuesLoop(source_datapipe, req_queue, res_queue, process_name, ca
     source_datapipe = extract_wrapper(source_datapipe)
 
     if call_on_process_init is not None:
-        call_on_process_init(source_datapipe)
+        source_datapipe = call_on_process_init(source_datapipe)
```
NivekT, line 112: Can we add some documentation or typing for `call_on_process_init` somewhere within the eventloop.py file?

torchdata/dataloader2/utils/worker.py
```diff
@@ -96,6 +97,9 @@ def process_init_fn(
         datapipe = custom_init_fn(datapipe, worker_info)
         assert isinstance(datapipe, (IterDataPipe, MapDataPipe))
 
+    if worker_prefetch_cnt > 0:
+        datapipe = datapipe.prefetch(worker_prefetch_cnt)
```
NivekT, line 101: IIUC, this is still applied whether or not the DataPipe is replicable but now it is after transformation done by the init function?

NivekT (approved):
LGTM! Small comments:

torchdata/dataloader2/communication/eventloop.py
ejguan: Do you mean adding inline doc for the arguments of `DataPipeToQueuesLoop`?

NivekT: Yes, or anywhere else that makes sense

ejguan: Sounds good

PR #1119, opened 2023-04-05
[DataLoader2] Adding support for naive checkpointing

NivekT:
* __->__ #1119

Differential Revision: [D44712802](https://our.internmc.facebook.com/intern/diff/D44712802)

### DIFF
--- a/torchdata/dataloader2/dataloader2.py
+++ b/torchdata/dataloader2/dataloader2.py
@@ -20,6 +20,7 @@ T_co = TypeVar("T_co", covariant=True)
 SERIALIZED_DATAPIPE_KEY_NAME = "serialized_datapipe"
 READING_SERVICE_STATE_KEY_NAME = "reading_service_state"
 RANDOMNESS_STATE_KEY_NAME = "randomness_state"
+NUM_PREV_YIELDED_BATCH_KEY_NAME = "num_of_previously_yielded_batches"
 
 
 class DataLoader2Iterator(Iterator[T_co]):
@@ -36,6 +37,7 @@ class DataLoader2Iterator(Iterator[T_co]):
         self.iterator_id = iterator_id
         self.limit_counter: Optional[int] = None
         self.limit_threshold: Optional[int] = None
+        self._batch_yielded: int = 0
 
     def __next__(self) -> T_co:
         if self.iterator_id == self.dataloader.valid_iterator_id:
@@ -47,6 +49,7 @@ class DataLoader2Iterator(Iterator[T_co]):
                     next_val = next(self.dataloader._datapipe_iter)  # type: ignore[arg-type]
                     if self.limit_threshold is not None:
                         self.limit_counter = self.limit_counter + 1  # type: ignore[operator]
+                    self._batch_yielded += 1
                     return next_val
             except PauseIteration:  # This can be used for raising `StopIteration` without `finalize_iteration`
                 raise StopIteration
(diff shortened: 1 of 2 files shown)


NivekT:
@nivekt has imported this pull request.  If you are a Meta employee, you can view this diff [on Phabricator](https://www.internalfb.com/diff/D44712802).

NivekT:
@nivekt has imported this pull request.  If you are a Meta employee, you can view this diff [on Phabricator](https://www.internalfb.com/diff/D44712802).

torchdata/dataloader2/dataloader2.py
```diff
@@ -348,6 +360,10 @@ def load_state_dict(self, state_dict: Dict[str, Any]) -> None:
             self._seed_generator = pickle.loads(randomness_state[2])
             self._initial_seed_generator = pickle.loads(randomness_state[3])
 
+        if NUM_PREV_YIELDED_BATCH_KEY_NAME in state_dict:
+            num_of_previously_yielded_batches = state_dict[NUM_PREV_YIELDED_BATCH_KEY_NAME]
+            self._num_prev_yielded_batches = num_of_previously_yielded_batches
```
ejguan, line 365: Can we combine them into a single line?

torchdata/dataloader2/dataloader2.py
```diff
@@ -365,6 +381,28 @@ def _restore_checkpoint_beginning_of_epoch(self) -> None:
         """
         self._seed_generator = self._initial_seed_generator
 
+    def _restore_naive_checkpoint(self, num_prev_yielded_batches: Optional[int] = None) -> DataLoader2Iterator[T_co]:
```
ejguan, line 384: When do we need to specify `num_prev_yielded_batches`?

I have a noob question: Why don't we make it automatically choosing the restoring option between naive or advanced based on if `NUM_PREV_YIELDED_BATCH_KEY_NAME` in the `state_dict`?

NivekT: > When do we need to specify `num_prev_yielded_batches`?

I see it as an option to override or to restore even if `self._num_prev_yielded_batches` has not been set. We can definitely take it out if you don't find it useful (and add it later if we see a need).

> automatically choose

I can imagine there are situations where users only want to restore the randomness state:
1) the model is only saved once per epoch (so you want DataLoader2 to be in sync)
2) maybe "naive" restoration is too slow and users just want to restore the randomness state, then do something custom

torchdata/dataloader2/dataloader2.py
```diff
@@ -222,7 +227,8 @@ def __iter__(self) -> DataLoader2Iterator[T_co]:
             self._reset_iter = False
 
         self.valid_iterator_id = 0 if self.valid_iterator_id is None else self.valid_iterator_id + 1
-        return DataLoader2Iterator(self, self.valid_iterator_id)
+        self._iterator = DataLoader2Iterator(self, self.valid_iterator_id)
```
ejguan, line 230: Can we try our best to prevent circular referencing?

torchdata/dataloader2/dataloader2.py
ejguan: > I see it as an option to override or to restore even if `self._num_prev_yielded_batches` has not been set.

If needed, users can always skipping number of iterations after doing `_restore_checkpoint_beginning_of_epoch`, right? I would like to keep the API minimum at first until a solid use case is required.


> * the model is only saved once per epoch (so you want DataLoader2 to be in sync)

In this case, we only need to save dataloader state once per epoch as well, right? If we want to support fault-tolerant, we need to make sure model has been stored at the same time.

> * maybe "naive" restoration is too slow and users just want to restore the randomness state, then do something custom

Can we list all the expected scenarios and corresponding API calls in the summary of the PR?

NivekT:

PR #1128, opened 2023-04-14
[NovaX] M1 Builds on Nova GHA

osalpekar:
As part of the move to Nova, this PR adds Wheels/Conda builds for M1 Macs for torchdata. It includes smoke-test support and the logic for uploading nightlies/release builds in the reusable workflows. If these builds run smoothly over the next 1-2 weeks, we can start deprecating the old release workflows. Until then, these Nova workflows do not upload any binaries to anaconda/pypi indices - they just run the build and verify with smoke tests. Would love any feedback from the data team about missing functionality/concerns/etc. regarding this move.

### DIFF
--- a/.github/workflows/build-conda-m1.yml
+++ b/.github/workflows/build-conda-m1.yml
@@ -23,19 +23,20 @@ jobs:
       test-infra-ref: main
   build:
     needs: generate-matrix
-    name: pytorch/vision
+    name: pytorch/data
     uses: pytorch/test-infra/.github/workflows/build_conda_macos.yml@main
     with:
-      conda-package-directory: packaging/torchvision
-      repository: pytorch/vision
+      conda-package-directory: packaging/torchdata
+      repository: pytorch/data
       ref: ""
       test-infra-repository: pytorch/test-infra
       test-infra-ref: main
       build-matrix: ${{ needs.generate-matrix.outputs.matrix }}
       pre-script: ""
       post-script: ""
-      package-name: torchvision
-      smoke-test-script: test/smoke_test.py
+      package-name: torchdata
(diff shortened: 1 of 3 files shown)


NivekT (approved):

NivekT:
Approved! Please import and land via Phabricator when it is ready

osalpekar:
@NivekT I see these errors at build time when `BUILD_S3=1` is set:

```
/Users/ec2-user/runner/_work/data/data/pytorch/data/third_party/aws-sdk-cpp/aws-cpp-sdk-core/source/utils/crypto/commoncrypto/CryptoImpl.cpp:668:26: error: 'CCCryptorGCMFinalize' is only available on macOS 10.13 or newer [-Werror,-Wunguarded-availability-new]
                status = CCCryptorGCMFinalize(m_decryptorHandle, m_tag.GetUnderlyingData(), tagLength);
                         ^~~~~~~~~~~~~~~~~~~~
```
Any ideas what this is? Looks like the macos-latest OS version from GitHub from the existing builds is also pre-10.13 so perhaps I am missing some config/env var here?

ejguan (commented):
It seems that it complains about [macOS version](https://github.com/pytorch/data/actions/runs/4703264613/jobs/8341547098?pr=1128#step:10:685). Should we specify something like [this](https://github.com/pytorch/data/blob/f2bfd3db8a70ea3ae1bef55188b4bd478436a7aa/packaging/pkg_helpers.bash#L111)?

NivekT (approved):
Do you mind fixing the linting? Thanks!

ejguan (commented):
Wondering why only the binary is only built for python3.8?

ejguan (approved):
Thank you for this effort!

PR #1139, opened 2023-04-21
Implement InProcessReadingService

### LINKED ISSUE #616: [TODO]  Warn and recommend usage of InProcessReadingService when `num_workers == 0` for  `MultiProcessingReadingService`
https://github.com/pytorch/data/blob/2f29adba451e1b87f1c0c654557d9dd98673fdd8/./torchdata/dataloader2/reading_service.py#L148
cc @VitalyFedyunin

### LINKED ISSUE #720: Add `SingleProcessReadingService` for `DataLoader2` by default
### 🚀 The feature

Add `SingleProcessReadingService` to adapt the graph for the sake of:
- Shuffle seed setting per epoch
- Set shuffle/sharding
- etc.

Pros:
This would prevent making `DataLoader2` more complicated. The reading service should take charge of all adaptation of the graph of `DataPipes`.

Cons:
By adding `SingleProcessReadingService`, we need mixed ReadingServices for the most of cases.

### Motivation, pitch

To align the behavior of `DataLoader2` with or without `MPRS`. 

### Alternatives

_No response_

### Additional context

_No response_

### LINKED ISSUE #1107: Dataloader2 does not invoke `set_seed` if used without ReadingService
### 🐛 Describe the bug

See title: My custom datapipe with `set_seed` does not get invoked when using `Dataloader2` (and calling seed(123)). This makes it impossible to fixate seeds when one does not want to use multiprocessing.

This is due to https://github.com/pytorch/data/blob/main/torchdata/dataloader2/dataloader2.py#L210

### Versions

e78ab6c9ec94f05f0a350ced7fe571f6863c20ec

ejguan:
Fixes #1107 
Fixes #720
Fixes #616

### Changes

- Implement `InProcessReadingService` (Willing to take any suggestion on naming)
  - Control shuffle and sharding (noop)
  - Add support to pause/resume/limit
- ~Make `InProcessReadingService` as the default `reading_service` to `DataLoader2`.~
  - ~Then, `reading_service` always has a value, and remove the logic of `reading_service` is None.~
- Modify `MultiProcessingReadingService`
  - When `num_workers=0`, raise a warning

### DIFF
--- a/torchdata/dataloader2/dataloader2.py
+++ b/torchdata/dataloader2/dataloader2.py
@@ -22,8 +22,8 @@ from torchdata.dataloader2.random import SeedGenerator
 from torchdata.dataloader2.random.seed_generator import _UINT64_UPPER_BOUND
 from torchdata.dataloader2.reading_service import (
     CheckpointableReadingServiceInterface,
+    InProcessReadingService,
     ReadingServiceInterface,
-    SingleProcessingReadingService,
 )
 
 T_co = TypeVar("T_co", covariant=True)
@@ -150,7 +150,7 @@ class DataLoader2(Generic[T_co]):
             right after the creation of the DataLoader.
         datapipe_adapter_fn (``Iterable[Adapter]`` or ``Adapter``, optional): ``Adapter`` function(s) that
             will be applied to the DataPipe (default: ``None``).
-        reading_service (ReadingServiceInterface): defines how ``DataLoader2`` should execute operations over
+        reading_service (ReadingServiceInterface, optional): defines how ``DataLoader2`` should execute operations over
             the ``DataPipe``, e.g. multiprocessing/distributed (default: ``None``). A deepcopy of this will be
             created during initialization, allowing the ReadingService to be re-used in a different
             ``DataLoader2`` without sharing states.
--- a/torchdata/dataloader2/dataloader2.py
+++ b/torchdata/dataloader2/dataloader2.py
@@ -20,11 +20,7 @@ from torchdata.dataloader2.graph._serialization import (
 )
 from torchdata.dataloader2.random import SeedGenerator
 from torchdata.dataloader2.random.seed_generator import _UINT64_UPPER_BOUND
-from torchdata.dataloader2.reading_service import (
-    CheckpointableReadingServiceInterface,
-    InProcessReadingService,
-    ReadingServiceInterface,
-)
+from torchdata.dataloader2.reading_service import CheckpointableReadingServiceInterface, ReadingServiceInterface
 
 T_co = TypeVar("T_co", covariant=True)
 SERIALIZED_DATAPIPE_KEY_NAME = "serialized_datapipe"
@@ -182,7 +178,6 @@ class DataLoader2(Generic[T_co]):
             self.datapipe_adapter_fns = datapipe_adapter_fn
         else:
             self.datapipe_adapter_fns = [datapipe_adapter_fn]
-
         self.reading_service = clone(reading_service)
         self.reading_service_state: Optional[bytes] = None  # is not `None` when `load_state_dict` is called
         self._terminated: bool = False
--- a/torchdata/dataloader2/dataloader2.py
+++ b/torchdata/dataloader2/dataloader2.py
@@ -20,7 +20,11 @@ from torchdata.dataloader2.graph._serialization import (
 )
 from torchdata.dataloader2.random import SeedGenerator
 from torchdata.dataloader2.random.seed_generator import _UINT64_UPPER_BOUND
-from torchdata.dataloader2.reading_service import CheckpointableReadingServiceInterface, ReadingServiceInterface
+from torchdata.dataloader2.reading_service import (
+    CheckpointableReadingServiceInterface,
+    ReadingServiceInterface,
+    SingleProcessingReadingService,
+)
 
 T_co = TypeVar("T_co", covariant=True)
 SERIALIZED_DATAPIPE_KEY_NAME = "serialized_datapipe"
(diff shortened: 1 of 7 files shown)


torchdata/dataloader2/reading_service.py
```diff
@@ -381,28 +447,21 @@ def _pause(self):
...
-        if self.main_prefetch_cnt > 0 and self.num_workers > 0:
-            self._main_prefetch_datapipe.resume()  # type: ignore[union-attr]
+        assert self._end_datapipe is not None
+
+        self._worker_consumer_datapipe.request_resume()  # type: ignore[union-attr]
+        dp_list = list_dps(traverse_dps(self._end_datapipe))
+        # Reversed order
+        for dp in dp_list[::-1]:
+            if hasattr(dp, "resume") and callable(dp.resume):
+                dp.resume()
```
NivekT, line 464: Do we need to duplicate this logic when it also exists within `iter.py` where it checks `elif isinstance(request, communication.messages.ResumeRequest):`? Similar with `pause`. 

I need more time to think but perhaps there is a way for it to only exist in one place?

torchdata/dataloader2/reading_service.py
```diff
@@ -146,6 +146,73 @@ def __new__(cls, *args, **kwargs):
         return MultiProcessingReadingService(*args, **kwargs)
 
 
+class SingleProcessingReadingService(ReadingServiceInterface):
```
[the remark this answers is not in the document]
ejguan: Sure, let's do `InProcessReadingService`

torchdata/dataloader2/dataloader2.py
```diff
@@ -141,10 +145,11 @@ class DataLoader2(Generic[T_co]):
             right after the creation of the DataLoader.
         datapipe_adapter_fn (``Iterable[Adapter]`` or ``Adapter``, optional): ``Adapter`` function(s) that
             will be applied to the DataPipe (default: ``None``).
-        reading_service (ReadingServiceInterface, optional): defines how ``DataLoader2`` should execute operations over
+        reading_service (ReadingServiceInterface): defines how ``DataLoader2`` should execute operations over
```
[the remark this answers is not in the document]
ejguan: You are right.

torchdata/dataloader2/reading_service.py
```diff
@@ -186,6 +252,20 @@ class MultiProcessingReadingService(ReadingServiceInterface):
...
+    def __new__(
+        cls,
+        num_workers: int = 0,
+        multiprocessing_context: Optional[str] = None,
+        worker_prefetch_cnt: int = 10,
+        main_prefetch_cnt: int = 10,
+        worker_init_fn: Optional[Callable[[DataPipe, WorkerInfo], DataPipe]] = None,
+        worker_reset_fn: Optional[Callable[[DataPipe, WorkerInfo, SeedGenerator], DataPipe]] = None,
+    ):
+        if num_workers == 0:
```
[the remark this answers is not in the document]
ejguan: Using `num_workers=1` would create a single worker process that is dedicated to loading data. And, the main process would work on the model training. The use case might be preventing the data part drain the CPU resource.

torchdata/dataloader2/reading_service.py
```diff
@@ -146,6 +146,73 @@ def __new__(cls, *args, **kwargs):
...
+            as the expected arguments.
+    """
+    _init_fn: Optional[Callable[[DataPipe, WorkerInfo], DataPipe]]
+    _reset_fn: Optional[Callable[[DataPipe, WorkerInfo, SeedGenerator], DataPipe]]
+    _end_datapipe: Optional[DataPipe]
+
+    def __init__(
+        self,
+        init_fn: Optional[Callable[[DataPipe, WorkerInfo], DataPipe]] = None,
+        reset_fn: Optional[Callable[[DataPipe, WorkerInfo, SeedGenerator], DataPipe]] = None,
```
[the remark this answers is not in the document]
ejguan: Maybe adding one is better. But, it will be BC-breaking as previously default `DataLoader2` won't do any prefetching. So, I guess I will keep that a separate small PR.

torchdata/dataloader2/reading_service.py
ejguan: We do because the logic in `iter` is called either in worker process or dispatching process. But, this part is called in the main process for the rest of `DataPipes` remaining in the main process.

NivekT: In that case, we need something similar for `_pause` right above as well?

ejguan: I think `_pause` has already done that.

torchdata/dataloader2/reading_service.py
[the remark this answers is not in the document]
NivekT: Maybe we can add the option but leave the default as no prefetch?

torchdata/dataloader2/reading_service.py
```diff
@@ -188,13 +258,14 @@ class MultiProcessingReadingService(ReadingServiceInterface):
...
         self,
-        num_workers: int = 0,
+        num_workers: int,
         multiprocessing_context: Optional[str] = None,
         worker_prefetch_cnt: int = 10,
         main_prefetch_cnt: int = 10,
         worker_init_fn: Optional[Callable[[DataPipe, WorkerInfo], DataPipe]] = None,
         worker_reset_fn: Optional[Callable[[DataPipe, WorkerInfo, SeedGenerator], DataPipe]] = None,
     ) -> None:
+        assert num_workers > 0, "Please use `InProcessReadingService` for num_workers=0"
```
ejguan, line 268: I am making a BC-breaking change here. `MPRS` won't accept `num_workers=0` anymore.

In previous commits, I have been trying to add a `__new__` to return `InProcessReadingService` when `num_workers=0`. However, it doesn't work well with `pickling` due to clone and mp.

NivekT (commented):
Thanks! LGTM! Let's also add it to this documentation table here:

https://pytorch.org/data/beta/dataloader2.html#readingservice

NivekT (approved):

torchdata/dataloader2/reading_service.py
```diff
@@ -381,28 +440,21 @@ def _pause(self):
...
-        if self.main_prefetch_cnt > 0 and self.num_workers > 0:
-            self._main_prefetch_datapipe.resume()  # type: ignore[union-attr]
+        assert self._end_datapipe is not None
+
+        self._worker_consumer_datapipe.request_resume()  # type: ignore[union-attr]
+        dp_list = list_dps(traverse_dps(self._end_datapipe))
+        # Reversed order
+        for dp in dp_list[::-1]:
+            if hasattr(dp, "resume") and callable(dp.resume):
+                dp.resume()
```
ejguan, line 457: I need to skip `QueueWrapper` here as well.

PR #1146, opened 2023-05-01
[DataLoader2] Handle `MapDataPipe` by converting to `IterDataPipe` internally by default

### LINKED ISSUE #1143: AttributeError: 'SequenceWrapperMapDataPipe' object has no attribute 'prefetch . when using MultiProcessingReadingService
### 🐛 Describe the bug

```python
from torchdata.datapipes.map import SequenceWrapper
from torchdata.dataloader2 import DataLoader2
from torchdata.dataloader2 import MultiProcessingReadingService


class SimpleDataset:
    def __len__(self):
        return 1000

    def __getitem__(self, i):
        if i >= len(self): raise StopIteration()
        return i


if __name__ == '__main__':
    ds = SimpleDataset()
    warp_ds = SequenceWrapper(ds)

    rs = MultiProcessingReadingService(num_workers=2)
    # no error
    # dl = DataLoader2(warp_ds)
    # raise error
    dl = DataLoader2(warp_ds, reading_service=rs)

    for x in dl:
        print(x)

```

### Versions

torchdata nightly

NivekT:
* __->__ #1146

Fixes #1143

I believe since `DataLoader2` would always call `iter(map_data_pipe)` at some point anyway, it should make no different to perform the conversion with `to_iter_datapipe()` within `__init__`.

This will also allow prefetchers to be attached to the DataPipe.

Please let me know if this may be BC-breaking.

Differential Revision: [D45452068](https://our.internmc.facebook.com/intern/diff/D45452068)

### DIFF
--- a/torchdata/dataloader2/dataloader2.py
+++ b/torchdata/dataloader2/dataloader2.py
@@ -11,7 +11,13 @@ from typing import Any, Dict, Generic, Iterable, Iterator, Optional, TypeVar, Un
 
 from torchdata.dataloader2.adapter import Adapter
 from torchdata.dataloader2.error import PauseIteration
-from torchdata.dataloader2.graph._serialization import clone, DataPipe, deserialize_datapipe, serialize_datapipe
+from torchdata.dataloader2.graph._serialization import (
+    clone,
+    DataPipe,
+    deserialize_datapipe,
+    MapDataPipe,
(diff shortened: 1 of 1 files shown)


NivekT:
@ejguan Let me know what you think of this.

ejguan (approved):
LGTM! Thanks

PR #1147, opened 2023-05-02
Enable miniepoch for MultiProcessingReadingService

ejguan:
Summary:
Add `pause`, `resume`, `limit` API to `MultiProcessingReadingService` to support miniepoch API.
- Adding a function as the input to make sure it working with internal ReadingService to control miniepoch within the worker/dispatching process

All tests should be passing in `test/dataloader2/test_mrps.py`

Differential Revision: D45289736

### DIFF
--- a/torchdata/dataloader2/reading_service.py
+++ b/torchdata/dataloader2/reading_service.py
@@ -366,12 +366,20 @@ class MultiProcessingReadingService(ReadingServiceInterface):
         self._worker_processes = []
         self._dispatch_process = None
 
-    def _pause(self):
+    def _pause(
+        self, pause_fn: Optional[Callable[[DataPipe], DataPipe]] = None
+    ) -> Optional[Callable[[DataPipe], DataPipe]]:
         """
         Pauses DataPipes' activities such as prefetching within main/worker/dispatching processes,
         in order to collect state.
         """
+        if self.num_workers == 0:
+            raise RuntimeError(
+                "If you would like to use `pause` with `MultiProcessingReadingService`, "
+                "please use more than 0 worker."
+            )
         assert self._end_datapipe is not None
+        # Call pause for DataPipes in the main process (e.g. prefetch, fullsync)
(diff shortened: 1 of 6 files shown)


torchdata/dataloader2/reading_service.py
```diff
@@ -381,35 +388,44 @@ def _pause(self):
...
+        dp_list = list_dps(traverse_dps(self._end_datapipe))
+        for dp in dp_list[::-1]:
+            # TODO: Combine QueueWrapper and _IterateQueueDataPipes,
+            #       and attach resume method. Then, no need to call
+            #       self._worker_consumer_datapipe.request_resume()
+            if isinstance(dp, communication.iter.QueueWrapper):
+                continue
+            if hasattr(dp, "resume") and callable(dp.resume):
+                dp.resume()
+        return None
```
NivekT, line 418: What DataPipes do we need to `resume`/`pause` here aside from the `prefetch` for main loop? Dispatching process (seems to be handled elsewhere)? `FullSync` (unlikely)?

optional nit: Maybe add a comment before `assert` to indicate the purpose?

NivekT (commented):
With this PR, is `pause`/`resume` full functional with dispatching process? Or is something still missing?

NivekT (approved):

torchdata/dataloader2/reading_service.py
ejguan: > With this PR, is pause/resume full functional with dispatching process? Or is something still missing?

Yes, it is. I might add one more test to validate it.

> What DataPipes do we need to `resume`/`pause` here aside from the `prefetch` for main loop? Dispatching process (seems to be handled elsewhere)? `FullSync` (unlikely)?

Two possible DataPipes. One is fullsync and prefetch, another case is some user-defined non-replicable datapipe with `resume`/`pause` implemented (in the main process)

> optional nit: Maybe add a comment before `assert` to indicate the purpose?

Sure, I can do that.

PR #1152, opened 2023-05-05
[Release 0.6.1] Enable official release

ejguan:
Per title


ejguan (commented):
Actually, this is not needed any more

PR #1153, opened 2023-05-05
[DataPipe] Fix FullSync shutdown hanging issue while paused

NivekT:
* __->__ #1153

Before this PR, the executor within FullSync fails to shutdown if it were currently paused. This PR allows shutdown without submitting additional jobs.

Differential Revision: [D45610885](https://our.internmc.facebook.com/intern/diff/D45610885)

### DIFF
--- a/test/test_distributed.py
+++ b/test/test_distributed.py
@@ -149,6 +149,14 @@ class DistributedTest(TestCase):
         except Exception as e:
             assert isinstance(e, PrefetchTimeoutError)
 
+        # Test that reset/shutdown does not hang while paused
+        dp3 = dp.fullsync()
+        it = iter(dp3)
+        next(it)
+        dp3.pause()
+        it2 = iter(dp3)  # Reset
+        next(it2)
+
         _finalize_distributed_queue(rank, q)
 
     @world_size_parametrize
--- a/torchdata/datapipes/iter/util/distributed.py
+++ b/torchdata/datapipes/iter/util/distributed.py
@@ -114,6 +114,7 @@ class _PrefetchExecutor:
         return data
 
     def shutdown(self):
+        self._paused = False
         self._executor.shutdown(wait=True)
 
     def pause(self):
--- a/torchdata/datapipes/iter/util/distributed.py
+++ b/torchdata/datapipes/iter/util/distributed.py
@@ -69,9 +69,11 @@ class _PrefetchExecutor:
         self._executor = ThreadPoolExecutor(max_workers=1)
         self._futures: Deque[Future] = deque()
         self._lock = threading.RLock()
+        # `_end_flag` indicates the end of epoch or an exception has been raised,
+        # with the exception being handled by `callback_fn`
         self._end_flag: bool = False
         self._paused: bool = False
-        self._is_shutdown: bool = False
+        self._is_shutdown: bool = False  # indicates if `_executor` has been shutdown by `shutdown` method
         self._idx = 0
         for _ in range(prefetch_size):
             with self._lock:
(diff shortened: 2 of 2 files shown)


test/test_distributed.py
```diff
@@ -149,6 +149,14 @@ def _test_fullsync(rank, world_size, backend, q):
         except Exception as e:
             assert isinstance(e, PrefetchTimeoutError)
 
+        # Test that reset/shutdown does not hang while paused
+        dp3 = dp.fullsync()
+        it = iter(dp3)
+        next(it)
+        dp3.pause()
+        it2 = iter(dp3)  # Reset
+        next(it2)
```
NivekT, line 158: Does this look reasonable? I also created a local test script to test `_PrefetchExector` and confirmed that it doesn't shutdown (it hangs) when it is paused:

```python
from torchdata.datapipes.iter.util.distributed import _PrefetchExecutor
import time

it = iter(range(10))
e = _PrefetchExecutor(it, 1)
print(e.fetch_next())
e.pause()
print("pause is called")
e.get_next()  # custom method to call fetch_next()
time.sleep(0.5)
print("shutdown is called")
e.shutdown()
```

NivekT:
@nivekt has imported this pull request.  If you are a Meta employee, you can view this diff [on Phabricator](https://www.internalfb.com/diff/D45610885).

test/test_distributed.py
```diff
@@ -149,6 +149,14 @@ def _test_fullsync(rank, world_size, backend, q):
...
+
```
ejguan, line 159: Is this test failing without the patch?

test/test_distributed.py
ejguan: Sounds good

test/test_distributed.py
ejguan: Can you add a test for `DataLoader2` with fullsync?

NivekT: Yea, it fails without the patch.

NivekT:
@nivekt has imported this pull request.  If you are a Meta employee, you can view this diff [on Phabricator](https://www.internalfb.com/diff/D45610885).

torchdata/datapipes/iter/util/distributed.py
```diff
@@ -93,7 +94,7 @@ def _done_callback_fn(self, index: int, f: Future):
         if f.exception():
             with self._lock:
                 self._end_flag = True
-        if self.callback_fn is not None:
+        if self.callback_fn is not None and not self._is_shutdown:
```
NivekT, line 97: Without this, it may attempts to submit the `callback_fn` even when the `_executor` has been shut down.

cc: @ejguan

ejguan: It seems reasonable.
Then, as the dev note, can you pls add some comments about the difference between `end_flag` and `is_shutdown`? Like `end_flag` indicates the end of epoch or exception,  and callback function will be executed to handle the exception.

test/test_distributed.py
NivekT: I can't really add a `DataLoader2` test because `DistributedRS` currently doesn't support `pause`....

I will have to add that separately. Let me know if I should land this as it is or add that on top of this.

NivekT:
@nivekt has imported this pull request.  If you are a Meta employee, you can view this diff [on Phabricator](https://www.internalfb.com/diff/D45610885).

ejguan (approved):

NivekT:
@nivekt has imported this pull request.  If you are a Meta employee, you can view this diff [on Phabricator](https://www.internalfb.com/diff/D45610885).

PR #1158, opened 2023-05-12
Fix pin_memory to support single-element batch

### LINKED ISSUE #1157: default pin_memory_fn with string Sequences and batch_size 1
### 🐛 Describe the bug

Pinning collated batches of size 1 of strings yields the string in a list instead of the string itself.
E.g. ['hello'] -> ['h', 'e', 'l', 'l', 'o']

```python

dp = IterableWrapper(["hello", "world"])
dp = dp.batch(1)
dp = dp.collate()
dp = dp.pin_memory()

print(next(iter(dp)))
```
```
['h', 'e', 'l', 'l', 'o']
```

Seems to be because in the `pin_memory_fn` on line 30 `type(data)(*pinned_data)` calls list on the string itself. For batch sizes > 1 this leads to an error which just returns `pinned_data`.
Should this line be  `type(data)(pinned_data)`?

### Versions

torchdata.__version__=='0.7.0a0+cacf355'

ejguan:
Fixes #1157

### DIFF
--- a/test/test_iterdatapipe.py
+++ b/test/test_iterdatapipe.py
@@ -1687,6 +1687,10 @@ class TestIterDataPipe(expecttest.TestCase):
         )
         self.assertTrue(all(d0.is_pinned() and d1.is_pinned() for batch in dp for d0, d1 in batch))
 
+        # Single str
+        dp = IterableWrapper(["hello", "world"]).batch(1).collate().pin_memory()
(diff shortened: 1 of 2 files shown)


NivekT (approved):
LGTM! Just to confirm, I don't think this will change the result for non-single element case. Correct?

ejguan:
> LGTM! Just to confirm, I don't think this will change the result for non-single element case. Correct?

Yes

PR #1160, opened 2023-05-13
Prevent `in_memory_cache` from yielding from `source_dp` when it's fully cached

### LINKED ISSUE #1159: InMemoryCache dp will yield from source dp once even when data is fully cached
### 🐛 Describe the bug

The `InMemoryCacheHolderIterDataPipe` will try yielding from `source_dp` once even when the data from the source is fully cached. This can cause unexpected IO for users. Adding a condition check before enumerating `source_dp` should fix it.

### Versions

Torch==2.0.1, TorchData==0.6.1

hirayaku:
Fixes #1159.


ejguan (approved):
Thank you. LGTM

hirayaku:
Hi, CLA is already signed and the status is updated. Thank you.

PR #1168, opened 2023-05-18
[DataLoader2] Adding pause/resume to DistributedRS

NivekT:
* __->__ #1168

Differential Revision: [D46029182](https://our.internmc.facebook.com/intern/diff/D46029182)

### DIFF
--- a/torchdata/dataloader2/reading_service.py
+++ b/torchdata/dataloader2/reading_service.py
@@ -371,7 +371,7 @@ class MultiProcessingReadingService(ReadingServiceInterface):
         Pauses DataPipes' activities such as prefetching within main/worker/dispatching processes,
         in order to collect state.
         """
-        assert self._end_datapipe is not None
+        assert self._end_datapipe is not None, "datapipe must not be `None` when `pause` is called by MultiProcessingRS"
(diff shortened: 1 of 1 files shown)


NivekT:
@ejguan There are a few things we should talk through:

1. I am taking the approach of iterating through the `DataPipes` that exist in the `ReadingService` and calling `pause/resume` on all of them. Let me know what you think.
2. `MultiprocessingRS`'s `_resume` (line 392) - I have a feeling that this should mostly mirror `_pause` in that it should create a `dp_list` and traverse through to call `resume` on each of the DataPipe. This will cover the case when there are multiple DPs in the main process that needs to be `resumed`. However, it is not implemented as such here. I think you mentioned why this is not done, but if you remember why, we should document it here.
    - Depending to the answer of this, the `_resume` implementation of `DistributedRS` may change.
3. `graph = ...` on line 302 goes unused. Should that be saved to `end_datapipe` in some ways or should we mark that as intentional unused?

NivekT:
@nivekt has imported this pull request.  If you are a Meta employee, you can view this diff [on Phabricator](https://www.internalfb.com/diff/D46029182).

NivekT:

PR #1170, opened 2023-05-23
Enable miniepoch for MultiProcessingReadingService

ejguan:
Summary:
A retry of https://github.com/pytorch/data/pull/1147
- Add pause, resume, limit API to MultiProcessingReadingService to support miniepoch API.
- When limit is requested, `IterQueueWrapper` will make sure only `limit` number of requests are sent to worker process and dispatching proccess.
- Consolidate a few API/messages
  - reset and reset_epoch is merged as a single message
  - Process reset function is shared for both worker process and dispatching process
- For pause/resume/limit/reset_epoch, add a counter to sync between loops to make sure the DataPipe graph is in-place modified once.
- Remove unused thread eventloop

Differential Revision: D45666912

### DIFF
--- a/torchdata/dataloader2/communication/eventloop.py
+++ b/torchdata/dataloader2/communication/eventloop.py
@@ -4,11 +4,10 @@
 # This source code is licensed under the BSD-style license found in the
 # LICENSE file in the root directory of this source tree.
 
-import pickle
-import threading
 import time
 
 from itertools import zip_longest
+from typing import Dict, List
 
 import torch
 
@@ -32,36 +31,47 @@ __all__ = [
     "DataPipeToQueuesLoop",
     "CreateProcessForDataPipeline",
     "CreateProcessForMultipleDataPipelines",
-    "CreateThreadForDataPipeline",
 ]
 
 
-class _ResetCounter:
+class _RequestCounter:
+    r"""
+    _RequestCounter is used to synchronize between eventloops within the dispatching
+    process. It guarantees to only handle the limit/pause/reset_epoch/resume request
+    util all loops have received the same message.
+    """
     exp_cnt: int
-    cnt: int
-    _reached: bool
+    _keys: List[str] = ["limit", "pause", "reset_epoch", "resume"]
+    _cnt: Dict[str, int]
+    _reached: Dict[str, bool]
 
     def __init__(self, exp_cnt: int):
         self.exp_cnt = exp_cnt
-        self.cnt = 0
-        self._reached = False
+        self._cnt = {k: 0 for k in self._keys}
+        self._reached = {k: False for k in self._keys}
 
-    def increment(self) -> None:
-        self.cnt += 1
-        assert self.cnt <= self.exp_cnt
+    def increment(self, key: str) -> None:
+        assert key in self._reached
+        self._cnt[key] += 1
+        assert self._cnt[key] <= self.exp_cnt
+        if self._cnt[key] == self.exp_cnt:
+            self._reached[key] = True
 
-    def is_reached(self) -> bool:
-        if self.cnt == self.exp_cnt:
-            self._reached = True
-        return self._reached
+    def is_reached(self, key: str) -> bool:
+        assert key in self._reached
+        return self._reached[key]
 
-    def reset(self) -> None:
-        if self._reached:
-            self._reached = False
-            self.cnt = 0
+    def reset(self, key: str) -> None:
+        assert key in self._reached and self._reached[key]
+        assert self._cnt[key] >= 1
+        self._cnt[key] -= 1
+        if self._cnt[key] == 0:
+            self._reached[key] = False
 
 
-def MultipleDataPipesToQueuesLoop(source_datapipes, req_queues, res_queues, process_name, call_on_process_init=None):
+def MultipleDataPipesToQueuesLoop(
+    source_datapipes, req_queues, res_queues, process_name, worker_info, call_on_process_init=None, custom_reset_fn=None
+):
     r"""
     Set the appropriate pipes and protocol server type, and create a loop over multiple datapipes
     with the protocol server in a non-blocking manner.
@@ -71,7 +81,9 @@ def MultipleDataPipesToQueuesLoop(source_datapipes, req_queues, res_queues, proc
         req_queue: Multiprocessing queue providing requests from the worker process
         res_queue: Multiprocessing queue sending results to the worker process
         process_name: The name of process (used for logging and exception handling)
+        worker_info: Worker information (worker id and number of workers)
         call_on_process_init: Not allowed by dispatching process for now.
+        custom_reset_fn: Optional callable function to reset the DataPipe.
     """
     assert call_on_process_init is None, "``MultipleDataPipesToQueuesLoop`` does not support call_on_process_init"
     num_loops = len(source_datapipes)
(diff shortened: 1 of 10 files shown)


torchdata/dataloader2/reading_service.py
```diff
@@ -527,3 +544,27 @@ def restore(self, datapipe, serialized_state: bytes) -> DataPipe:
...
                 warnings.warn(f"{rs} doesn't support `restore` from state, skipping...")
         return datapipe
+
+    def _pause(
+        self, pause_fn: Optional[Callable[[DataPipe], DataPipe]] = None
+    ) -> Optional[Callable[[DataPipe], DataPipe]]:
+        for rs in self.reading_services:
+            if hasattr(rs, "_pause"):
+                pause_fn = rs._pause(pause_fn)
+        return pause_fn
```
NivekT, line 554: I think we need an explanation of why this needs to accept an initial `pause_fn`? I assume it must be passed in from `DataLoader2._pause()`? And then the return object `pause_fn` will be used for the subsequent `rs._pause()`?

Ditto for `_resume` and `_limit`.

torchdata/dataloader2/reading_service.py
```diff
@@ -321,11 +328,9 @@ def initialize_iteration(
                 # Stop prefetching first
                 self._main_prefetch_datapipe.reset()  # type: ignore[union-attr]
             # Send the shared seed to subprocesses
-            call_on_epoch_reset = partial(
-                process_reset_fn, iter_reset_fn=iter_reset_fn, custom_reset_fn=self.worker_reset_fn
-            )
             assert self._worker_consumer_datapipe is not None
-            self._worker_consumer_datapipe.reset_epoch(call_on_epoch_reset, seed_generator)
+            print("[Erjia] Call reset epoch in main process")
```
NivekT, line 332: Remove print

torchdata/dataloader2/reading_service.py
```diff
@@ -303,6 +309,7 @@ def initialize(self, datapipe: DataPipe) -> DataPipe:
             end_datapipe = datapipe  # type: ignore[assignment]
 
         self._end_datapipe = end_datapipe
+        print(f"[Erjia] Main process graph: {traverse_dps(end_datapipe)}")
```
NivekT, line 312: Remove print

torchdata/dataloader2/reading_service.py
```diff
@@ -366,50 +371,62 @@ def finalize(self) -> None:
         self._worker_processes = []
         self._dispatch_process = None
 
-    def _pause(self):
+    def _pause(
+        self, pause_fn: Optional[Callable[[DataPipe], DataPipe]] = None
+    ) -> Optional[Callable[[DataPipe], DataPipe]]:
```
NivekT, line 376: Given `DataLoader2` doesn't invoke these methods with `pause_fn`, it is not obvious to users how it is intended to be used. Perhaps we should add some documentations

torchdata/dataloader2/dataloader2.py
```diff
@@ -377,25 +377,31 @@ def _restore_checkpoint_beginning_of_epoch(self) -> None:
...
         self._seed_generator = self._initial_seed_generator
 
-    def _pause(self):
+    def _pause(self) -> None:
         if hasattr(self.reading_service, "_pause"):
             self._is_paused = True
-            self.reading_service._pause()
+            pause_fn = self.reading_service._pause()
+            if pause_fn is not None:
+                self.datapipe = pause_fn(self.datapipe)
```
NivekT, line 385: Is the intention here to apply `pause` on the DataPipe in the main process as well?

torchdata/dataloader2/dataloader2.py
```diff
@@ -377,25 +377,31 @@ def _restore_checkpoint_beginning_of_epoch(self) -> None:
...
                 self._is_paused = False
         else:
             warnings.warn("ReadingService doesn't support `resume`.")
 
     def _limit(self, num_batches: Optional[int]) -> None:
         if hasattr(self.reading_service, "_limit"):
-            self.reading_service._limit(num_batches)
+            limit_fn = self.reading_service._limit(num_batches)
+            if limit_fn is not None:
+                self.datapipe = limit_fn(self.datapipe, num_batches)
```
NivekT, line 405: Depending on the `limit_fn`, the `limit` on the DataPipe and `DataLoader2Iterator` can be out of sync. One example is if there exists a `filter` in the pipeline. Correct?

torchdata/dataloader2/communication/eventloop.py
```diff
@@ -32,36 +31,42 @@
     "DataPipeToQueuesLoop",
     "CreateProcessForDataPipeline",
     "CreateProcessForMultipleDataPipelines",
-    "CreateThreadForDataPipeline",
 ]
 
 
-class _ResetCounter:
+class _RequestCounter:
```
NivekT, line 37: nit: I think it would be nice to have a short documentation about the usage of this class.

torchdata/dataloader2/communication/iter.py
```diff
@@ -149,53 +160,117 @@ def DataPipeBehindQueues(
...
+                    dp.request_limit(num_batches, limit_fn, worker_num_batches)
+                if limit_fn is not None:
+                    # Set limit to the DataPipe graph in worker/dispatching process
+                    source_datapipe = limit_fn(source_datapipe, worker_num_batches)
+            if request_counter is not None:
+                request_counter.reset("limit")
+                while request_counter.is_reached("limit"):
+                    yield True
+            protocol.response_limit()
+            yield True  # Returns control
```
NivekT, line 216: nit: I wonder if we can have a helper function to handles all the overlapping logic among `limit/pause/resume` with more inline docs.

NivekT (commented):
A few questions/notes. Overall, it does look good but our codebase is getting very complicated.

torchdata/dataloader2/reading_service.py
ejguan: Good catch.

torchdata/dataloader2/reading_service.py
ejguan: Sorry, what do you mean? `DataLoader2` does invoke the returned `pause_fn`.

torchdata/dataloader2/dataloader2.py
ejguan: Yes. Let's use an example of single DppRS. The returned `pause_fn` should do the traverse/search/pause to `DppDataPipe` in the main process.

torchdata/dataloader2/dataloader2.py
ejguan: Correct. I will add a note in `limit` that miniepoch doesn't support `filter` operation when dispatching process gets involved. It is still fine without dispatching process (only workers).

torchdata/dataloader2/communication/eventloop.py
ejguan: Sure

torchdata/dataloader2/communication/iter.py
ejguan: Yeah, I can do that.

torchdata/dataloader2/reading_service.py
NivekT: I mean this line in `DataLoader2` never pass in a `pause_fn`.

https://github.com/pytorch/data/blob/c8a52da300fa0d29b8603aeed475621897d794b3/torchdata/dataloader2/dataloader2.py#L383

ejguan: Yeah. In theory, users should not realize a `pause_fn` can be provided. This is should be tied to developers who want to develop their own ReadingService.

torchdata/dataloader2/dataloader2.py
ejguan: And, this can be removed if we can make sure all users are using SequentialRS with InProcessRS rather than directly using DppRS.

torchdata/dataloader2/reading_service.py
ejguan: And, the general reason that there is an initial pause_fn is to have an aligned API across `_pause`.

NivekT (approved):
LGTM! Thanks!

PR #1188, opened 2023-06-17
Simpler file chunking

adamjstewart:
Now that Python 3.8 is the minimum supported version, we can use the new [assignment expression](https://docs.python.org/3/whatsnew/3.8.html#assignment-expression) syntax to simplify how we read files chunk-by-chunk.

### DIFF
--- a/test/_utils/_common_utils_for_test.py
+++ b/test/_utils/_common_utils_for_test.py
@@ -87,9 +87,7 @@ def check_hash_fn(filepath, expected_hash, hash_type="md5"):
         raise ValueError("Invalid hash_type requested, should be one of {}".format(["sha256", "md5"]))
 
     with open(filepath, "rb") as f:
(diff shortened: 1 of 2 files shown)


adamjstewart:
Ping @ejguan

adamjstewart:
@pmeier can you review since you reviewed https://github.com/pytorch/vision/pull/7673 too?

pmeier (approved):
Thanks @adamjstewart, LGTM if CI is green. Although I can review, I don't have merge rights here. So we need to wait for the maintainers.

adamjstewart:
Ping ping @ejguan

pmeier:
Sorry to be the bearer of the bad news here, but it seems this PR won't go anywhere in the near future: #1196

adamjstewart:
Thanks for the heads up!

PR #1195, opened 2023-07-17
correct SampleMultiplexer docstring example

timothee1996:
### Changes

- correct docstring of SampleMultiplexerDataPipe. The examples show a result of multiplexing 2 datapipes of length 5 but the two datapipes are initialized with 10 elements each


gokulavasan:
looks good to me!

gokulavasan (approved):

PR #1199, opened 2023-08-14
Add aarch64 to TorchData CICD

xncqr:
creating your pull request.

Fixes #{issue number}

### Changes
* Adding aarch64 workflow to CICD.
* Setting setup-miniconda to `false` as conda is installed from "Set linux aarch64 CI"
* Updated `post_build_script_linux.sh` to support different CPU arch types by finding the arch from there the script is executed.
* Remove torchvision from domain_ci testing as the newer versions of torchvision do not use torchdata.

### DIFF
--- a/.github/workflows/build_wheels_aarch64_linux.yml
+++ b/.github/workflows/build_wheels_aarch64_linux.yml
@@ -1,4 +1,4 @@
-name: Build Linux Wheels
+name: Build AARCH64 Linux Wheels
 
 on:
   pull_request:
--- a/.github/workflows/build_wheels_aarch64_linux.yml
+++ b/.github/workflows/build_wheels_aarch64_linux.yml
@@ -41,6 +41,7 @@ jobs:
       package-name: ${{ matrix.package-name }}
       env-var-script: packaging/env-var-script.txt
       architecture: aarch64
+      setup-miniconda: false
       # Using "development" as trigger event so these binaries are not uploaded to official channels yet
       trigger-event: development
     secrets:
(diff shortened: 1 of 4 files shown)


atalman (approved):
LGTM we can merge if CI is green

msaroufim:
Btw the torchvision tests are expected to fail because they've dropped their use of torchdata so the only job that needs to be green is the BUILD AARCH64

xncqr:
> Btw the torchvision tests are expected to fail because they've dropped their use of torchdata so the only job that needs to be green is the BUILD AARCH64
> 
> You can however fix the torchvision tests if you pin to an older version https://github.com/pytorch/data/blob/main/.github/workflows/domain_ci.yml#L13

Since this is dropped for torchvision, do we even need the test anymore?

msaroufim:
yeah probably not, feel free to delete and I can retrigger CI

msaroufim:
idk what's going on with the linter but try running precommit and see if that unblocks

xncqr:
@msaroufim think we are good this time. Want to give it another try?

msaroufim (approved):

msaroufim:
Err not sure what to do here there's way too many failing tests that I'm not sure this is right to merge - wdyt @atalman ?

xncqr:
7 of the 8 were `Error: The operation was canceled.` we know why they were canceled? the MyPy looks like an internal error, but sure what to think about that one.

xncqr:
am closing this as the repo is a little, on pause.

PR #1206, opened 2023-10-03
Use GitHub M1 runners

huydhn:
To build and test conda and wheel for release.


atalman:
The workflow passes but it does not looks like it builds correct wheel. We are looking for something like this:
```
torchdata-0.7.0-cp311-cp311-macosx_11_0_arm64.whl 
``

huydhn:
Closing this as this is outdate. The build workflow in data has been cleaned up, and the label is call macos-m1-stable now.

PR #1210, opened 2023-10-16
build: Just set torch>=2.0

seemethere:
This project is in maintenance mode so set it free from having a hard dependency on a single version of PyTorch


abhinavarora (approved):
LGTM! Thank you for making this change!

gdippolito:
Would it be possible to tag a release with this commit so users can install torchdata with Pytorch 2.1? 🙏 
Thanks

PR #1213, opened 2023-11-01
Relax torchdata torch conda requirement to >= 2.0

huydhn:
This is similar to https://github.com/pytorch/data/pull/1210, but relax the requirement for conda build instead of wheel.  For example, `linux-64/torchdata-0.7.0-py39.tar.bz2` conda packages on https://anaconda.org/pytorch-test/torchdata/files currently has the following deps `openssl >=3.0.11,<4.0a0, python >=3.9,<3.10.0a0, pytorch 2.1.0, requests, urllib3 >=1.25, zlib >=1.2.13,<1.3.0a0` where [pytorch](https://anaconda.org/pytorch-test/pytorch) is fixed at `2.1.0`

PR #1214, opened 2023-11-01
[Release-only] Bump minor version to 0.7.1

huydhn:
Although the package has paused new releases, I think we need this final bump because conda package metadata (https://anaconda.org/pytorch-test/torchdata/files) won't be updated otherwise. And we probably want to include https://github.com/pytorch/data/pull/1213 to relax `pytorch>=2.0` similar to https://github.com/pytorch/data/pull/1210


seemethere (approved):
LGTM

huydhn:
Merged into release branch https://github.com/pytorch/data/commits/release/0.7

PR #1227, opened 2024-02-26
Data loader directly use loaded state dict when save right after load

MiiiraK:
Summary: If no iterator is created in the middle of load_state_dict and state_dict calls, we should be able to directly return the original state dict without triggering reading service because the states are unchanged

Reviewed By: xunnanxu

Differential Revision: D54102267


ejguan (approved):

gokulavasan:
PR already merged in the main branch - https://github.com/pytorch/data/commit/39b3497517d5850c600c6b18e73cf5a346a539bf

PR #1232, opened 2024-03-25
Set mypy version to match pytorch core

gokulavasan:
Here is the link to mypy version on pytorch core - https://github.com/pytorch/pytorch/blob/198927170d34efaeb77e1a772d96ad3615a21cba/.ci/docker/requirements-ci.txt#L88

There are mypy failures on the main branch. Trying to update the mypy version before attempting to fix them.

### Changes

- Update the mypy version


gokulavasan:
merged to master, closing PR

PR #1234, opened 2024-03-26
Fix two types of issues raised by mypy - implicit optional issues and arg-type mismatch

gokulavasan:
Title. Here is the mypy run failing with 24 errors - https://github.com/pytorch/data/actions/runs/8428471909/job/23081016202.

With these two changes, the mypy errors are now down to 8 - https://github.com/pytorch/data/actions/runs/8428898766/job/23082258705?pr=1234

### Changes

- Add optional where it was implicit earlier
- Change ignore type to arg-type

### DIFF
--- a/torchdata/datapipes/iter/util/rows2columnar.py
+++ b/torchdata/datapipes/iter/util/rows2columnar.py
@@ -5,7 +5,7 @@
 # LICENSE file in the root directory of this source tree.
 
 from collections import defaultdict
-from typing import Dict, Iterator, List, Union
+from typing import Dict, Iterator, List, Optional, Union
 
 from torchdata.datapipes import functional_datapipe
 from torchdata.datapipes.iter import IterDataPipe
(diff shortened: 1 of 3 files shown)


ejguan (commented):
Do you plan to fix all mypy error? If so, you can run `mypy --config=mypy.ini` to check them locally.

gokulavasan:
@ejguan Yes I am trying to fix all the errors. Got it down from 24 to 8. When I run the local command that you provided above, I am getting 11 errors. I will try and get to the remaining 8 in a separate PR as they warrant a bit more debugging

andrewkho:
LGTM but noob q: where can we see that the errors decreased?

torchdata/datapipes/iter/util/rows2columnar.py
```diff
@@ -50,7 +50,9 @@ class Rows2ColumnarIterDataPipe(IterDataPipe[Dict]):
     """
     column_names: List[str]
 
-    def __init__(self, source_datapipe: IterDataPipe[List[Union[Dict, List]]], column_names: List[str] = None) -> None:
+    def __init__(
+        self, source_datapipe: IterDataPipe[List[Union[Dict, List]]], column_names: Optional[List[str]] = None
+    ) -> None:
```
andrewkho, line 55: Another n00b q: what formatter are you using in torchdata vs torch.utils.data? Is linting part of CI on the github side?

gokulavasan: Formatter being used - black - https://github.com/pytorch/data/blob/4725a01ce38c646b6273acacb9df895b3d9c3f66/.pre-commit-config.yaml#L28-L34.

linting part of CI is on the github side. I think there is an explicit exclusion of it in phabricator.

ejguan (approved):

PR #1245, opened 2024-04-25
Stateful dataloader README

andrewkho:
creating your pull request.

Fixes #{issue number}

### Changes
Adds README.md for torchdata.stateful_dataloader

Testplan:
Checked rendering and links in the branch preview on github:
image

-
-

### DIFF
--- a/torchdata/stateful_dataloader/README.md
+++ b/torchdata/stateful_dataloader/README.md
@@ -1,8 +1,9 @@
 # StatefulDataLoader
 
-[**Saving and Loading State**](#saving-and-loading-state) | [**Custom State: Map-Style**](#saving-custom-state-with) |
-[**Custom State: Iterable-Style**](#saving-custom-state-with-1) | [**Install guide**](#installation) |
-[**Beta Usage and Feedback**](#beta-usage-and-feedback) | [**License**](#license)
+[**Saving and Loading State**](#saving-and-loading-state) |
+[**Custom State: Map-Style**](#saving-custom-state-with-map-style-datasets) |
+[**Custom State: Iterable-Style**](#saving-custom-state-with-iterable-style-datasets) |
+[**Install guide**](#installation) | [**Beta Usage and Feedback**](#beta-usage-and-feedback) | [**License**](#license)
 
 `StatefulDataLoader` is a drop-in replacement for `torch.utils.data.DataLoader` which offers
 `state_dict/load_state_dict` methods for handling mid-epoch checkpointing which operate on the previous/next iterator
--- a/torchdata/stateful_dataloader/README.md
+++ b/torchdata/stateful_dataloader/README.md
@@ -21,7 +21,7 @@ handles aggregation and distribution of state across multiprocess workers (but n
 Using pip:
 
 ```bash
-pip install --pre torchdata --extra-index-url https://download.pytorch.org/whl/nightly/cpu
+pip install --pre torchdata --index-url https://download.pytorch.org/whl/nightly/cpu
 ```
 
 Using conda:
@@ -65,6 +65,9 @@ import torch
 import torch.utils.data
 from torchdata.stateful_dataloader import StatefulDataLoader
 
+# If you are using the default RandomSampler and BatchSampler in torch.utils.data
+# they are patched when you import torchdata.stateful_dataloader so that defining
+# a custom sampler here is unnecessary
 class MySampler(torch.utils.data.Sampler[int]):
   def __init__(self, high: int, seed: int, limit: int):
     self.seed, self.high, self.limit = seed, high, limit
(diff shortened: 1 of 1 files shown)


andrewkho:
It seems that CI gets skipped when the last action is by facebook-github-bot

torchdata/stateful_dataloader/README.md
```diff
@@ -0,0 +1,215 @@
...
+your dataset has worker-specific state (eg RNG transform state) you can add `state_dict/load_state_dict` methods to your
+dataset.
+
+```py
+from typing import *
+import torch
+import torch.utils.data
+from torchdata.stateful_dataloader import StatefulDataLoader
+
+class MySampler(torch.utils.data.Sampler[int]):
```
gokulavasan, line 68: Should we mentioned that RandomSampler and BatchSampler has been patched so that it should work out of the box for users who use the default samplers. On a related note, should we also consider making DistributedSampler 'Stateful`?

torchdata/stateful_dataloader/README.md
```diff
@@ -0,0 +1,215 @@
...
+    noise = x.item()
+    return idx + noise
+
+  def load_state_dict(self, state_dict):
+    torch.set_rng_state(state_dict["rng"])
+
+  def state_dict(self):
+    return {"rng": torch.get_rng_state()}
+
+for num_workers in [0, 2]:
```
gokulavasan, line 109: Maybe add a comment here that it works for both SingleProcess and MultiProcess dataloading setups?

torchdata/stateful_dataloader/README.md
```diff
@@ -0,0 +1,215 @@
...
+[tensor([3.7412, 1.2438], dtype=torch.float64), tensor([4.4807, 4.0036], dtype=torch.float64)]
+"""
+```
+
+## Saving Custom State with Iterable-Style Datasets
+
+Tracking iteration order with Iterable-style datasets requires state from each worker-level instance of the dataset to
+be captured. You can define `state_dict/load_state_dict` methods on your dataset which capture worker-level state.
+`StatefulDataLoader` will handle aggregation across workers and distribution back to the workers. Calling
+`load_state_dict` requires `StatefulDataLoader` to have same `num_workers` as those of the provided `state_dict`.
```
gokulavasan, line 142: `num_workers` need to be same even for map-style datasets correct? 

Separate discussions: 
i) should we consider relaxing that if map dataset doesn't have state?
ii) assert statement https://github.com/pytorch/data/blob/main/torchdata/stateful_dataloader/stateful_dataloader.py#L1211 can have an error message to make this check explicit for the user

torchdata/stateful_dataloader/README.md
```diff
@@ -0,0 +1,215 @@
...
+
+  def __iter__(self):
+    worker_info = torch.utils.data.get_worker_info()
+    if worker_info is not None:
+      worker_id = worker_info.id
+      num_workers = worker_info.num_workers
+    else:
+      worker_id = 0
+      num_workers = 1
+    self.g.manual_seed(self.seed)
```
gokulavasan, line 165: Do we need to set the seed to be same for all workers? Can the one set by the worker loop in worker.py be used?

torchdata/stateful_dataloader/README.md
```diff
@@ -0,0 +1,215 @@
...
+Output:
+num_workers=0
+[tensor([ 2, 10]), tensor([3, 1]), tensor([11,  6])]
+[tensor([ 2, 10]), tensor([3, 1]), tensor([11,  6])]
+num_workers=2
+[tensor([ 4, 10]), tensor([ 3, 11]), tensor([1, 6])]
+[tensor([ 4, 10]), tensor([ 3, 11]), tensor([1, 6])]
+"""
+```
+
```
gokulavasan, line 207: [This can be done in a separate PR too] Can we also list the corner cases where it doesn't work as expected? Basically this case where user has to pass in an explicit generator for it work correctly - https://github.com/pytorch/data/blob/main/test/stateful_dataloader/test_state_dict.py#L686.

gokulavasan (changes requested):
Few comments, looks good!

gokulavasan (changes requested):
Few comments, looks good otherwise!

torchdata/stateful_dataloader/README.md
andrewkho: num_workers can be different for map-style if they only care about iteration order, I can clarify that

torchdata/stateful_dataloader/README.md
andrewkho: Because of how this is set up (same array on all workers, iterate through based on worker_id) this is correct

torchdata/stateful_dataloader/README.md
andrewkho: Yes that's a good idea. If we could detect and raise assertion errors it would be even better

PR #1246, opened 2024-04-28
Don't patch _worker_loop

andrewkho:
Summary: We originally patched the _worker_loop function in torch.utils.data._utils.worker.py in order to make sure `torch.utils.data.get_worker_info()` worked correctly, but at some point this must have become unnecessary. This currently causes issues where importing StatefulDataLoader will break other usages of `torch.utils.data.DataLoader` with multiprocessing because the two _worker_loop's have different APIs. In this PR we drop the patching and add a small test that confirms that there is no conflict when using both non-stateful and StatefulDataLoader.

Test Plan: New unit test which used to fail but now passes. 

Fixes #{issue number}

### Changes

-
-

### DIFF
--- a/test/stateful_dataloader/test_state_dict.py
+++ b/test/stateful_dataloader/test_state_dict.py
@@ -10,7 +10,7 @@ import unittest
 from typing import Iterator
 
 import torch
-
+import torch.utils.data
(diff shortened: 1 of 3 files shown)


andrewkho:
The MacOS tests seem to be failing intermittently / inconsistently during Cleanup phase (not during testing)

PR #1248, opened 2024-04-30
add timing info to ci, and update tests for mac to use forkserver

### LINKED ISSUE #1247: [StatefulDataLoader] macOS tests are too slow
### 🐛 Describe the bug

test_state_dict is very slow on macOS (and slows down CI), likely because of macOS default multiprocessing_context being spawn instead of fork. The StatefulDataLoader tests on macOS take ~1.5 hours, vs 10 minutes on Linux and Windows. 

Example of test-runtimes on my local mac:

image

We should a) update CI to log test times, b) for macOS, drop some of the tests. Each test_mp* test runs 6x, and if we have coverage from Linux + Win then we probably don't need all of them for mac

### Versions

Nightly

andrewkho:
Fixes #1247 

### Changes
* Adjusts pytest CI command to show timing info
* Set all multiprocessing_context = "forkserver" for macOS, as default spawn is too slow 
-
-

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -86,7 +86,7 @@ jobs:
       - name: Run DataPipes tests with pytest
         if: ${{ ! contains(github.event.pull_request.labels.*.name, 'ciflow/slow') }}
         run:
-          pytest --no-header -v test --ignore=test/test_period.py --ignore=test/test_text_examples.py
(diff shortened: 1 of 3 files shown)


gokulavasan:
lg!

gokulavasan (approved):

PR #1249, opened 2024-05-01
Update actions/setup-python from v4 -> v5

andrewkho:
MacOS continues to fail during intermittently during shutdown step. Before creating an issue in actions/setup-python let's try upgrading this to V5 

Fixes #{issue number}

### Changes

-
-


andrewkho:
I've created an issue with actions/setup-python here: https://github.com/actions/setup-python/issues/857

PR #1250, opened 2024-05-06
update worker keys and add test

andrewkho:
Users [internal] are occasionally using json.dumps to serialize the state_dict when they know the state_dicts are JSON serializable. Ignoring the fact this won't work in general, this fails because json spec does not allow int properties (string required) which we're using for _worker_states in the state dict. This PR changes the format of the state_dict to use strings instead, ie `_worker_states: {0: ..., 1: ...}` -> `_worker_states: {"worker_0": ..., "worker_1": ...}`

### Changes

* Adds json serde test which fails before fix
* Sets worker state keys to explicitly be strings for zero ambiguity

-
-

### DIFF
--- a/test/stateful_dataloader/test_state_dict.py
+++ b/test/stateful_dataloader/test_state_dict.py
@@ -4,8 +4,8 @@
 # This source code is licensed under the BSD-style license found in the
 # LICENSE file in the root directory of this source tree.
 
-import copy
 import itertools
+import json
 import unittest
 from typing import Iterator
 
@@ -16,9 +16,10 @@ from torchdata.stateful_dataloader import Stateful, StatefulDataLoader
 
 
 class DummyIterator(Iterator, Stateful):
-    def __init__(self, samples, shuffle):
+    def __init__(self, samples, shuffle, include_generator):
         self.samples = samples
         self.shuffle = shuffle
+        self.include_generator = include_generator
         self.size = len(self.samples)
         self.i = 0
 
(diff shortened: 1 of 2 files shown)


test/stateful_dataloader/test_state_dict.py
```diff
@@ -841,5 +847,49 @@ def test_two_dataloaders(self) -> None:
         self.assertEqual(data, exp)
 
 
+class TestJsonSerDe(unittest.TestCase):
+    def _run_test(self, num_workers):
+        interrupt = 4
+        dataset = DummyIterableDataset([0, 100, 37], shuffle=False, include_generator=False)
```
gokulavasan, line 853: Do we need test case for map style with worker state too?

gokulavasan (changes requested):
Changes look good, one question about testing

gokulavasan (changes requested):
Changes look good, one question about testing

test/stateful_dataloader/test_state_dict.py
andrewkho: For our existing dummy tests, this wouldn't make a difference since non-RNG state only exists for the sampler, which shouldn't fall in to this category. I could add it if you feel strongly

gokulavasan: @andrewkho Hmm maybe I misunderstood something here - why does the state being RNG vs non-RNG matter for this change? For example, DummyMapDataset which has state (though RNG one) should be considered as worker state isn't it?

andrewkho: @gokulavasan it's for trivial reasons: rng state is not json-serializable so we can't include it in this test :( but you know what, I just had another look and realized that the sampler also does have json-serializable state, no RNG, so I'll add that test too

gokulavasan (approved):

PR #1251, opened 2024-05-06
fix an edge case bug where state is incomplete if request checkpoint after resume

andrewkho:
StatefulDataLoader may return incomplete checkpoints if a state_dict is requested soon after a resume.

Consider a multiprocess dataloader with num_workers = 4. When we resume this dataloader, assuming we make it through 4 batches before the next state_dict is requested, we should have fresh snapshots for all 4 workers again. However in the case where we only make it through say 2, then we are missing state for the other 2 workers. 

Test plan:
Wrote a new unit-test to capture this, which fails with below error before the fix:
```
        worker_states = [None] * self._num_workers
        if next_iter_state is not None:
            assert (
                self._SNAPSHOT in next_iter_state
            ), f"State doesn't contain key '{self._SNAPSHOT}' expected for multiprocess dataloader"
            wstates = next_iter_state[self._SNAPSHOT].get(self._WORKER_SNAPSHOTS, {})
>           assert set(range(len(wstates))) == set(wstates.keys()), (len(wstates), wstates.keys())
E           AssertionError: (2, dict_keys([3, 0]))
```

### Changes
* Fixes edge-case bug where StatefulDataLoader.state_dict() may return an incomplete state after a resume if a request is made before num_workers steps.

-
-

### DIFF
--- a/test/stateful_dataloader/test_state_dict.py
+++ b/test/stateful_dataloader/test_state_dict.py
@@ -841,5 +841,49 @@ class TestConcurrentDataLoaders(unittest.TestCase):
         self.assertEqual(data, exp)
 
 
+class TestFailAfterResume(unittest.TestCase):
+    def test_fail_after_resume(self) -> None:
+        num_workers = 4
+        interrupt = 11  # because of round robin, this should stop after worker 2
+        dataset = DummyIterableDataset([25, 25, 25, 25], shuffle=True)
+
+        dl = StatefulDataLoader(
+            dataset=dataset,
+            num_workers=num_workers,
+            batch_size=4,
+            collate_fn=identity,
+            persistent_workers=True,
+            multiprocessing_context="forkserver" if IS_MACOS else None,
+        )
+        it = iter(dl)
+        for _ in range(interrupt):
+            next(it)
+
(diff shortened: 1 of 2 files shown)


test/stateful_dataloader/test_state_dict.py
```diff
@@ -841,5 +841,49 @@ def test_two_dataloaders(self) -> None:
...
+        for _ in range(interrupt):
+            next(it)
+
+        state_dict = dl.state_dict()
+        for _ in range(2):
+            next(it)
+        exp = list(it)
+
+        dl.load_state_dict(state_dict)
+        # new iter after load_state_dict, try to fail fast
```
gokulavasan, line 868: Can you expand on fail fast means here?

gokulavasan (approved):

andrewkho:
The CI did run, it's visible on the checks page. Not sure if we can fix this, the facebook-github-bot CLA Signed message seems to kick off a new set of actions, showing the tests as skipped. 

For the tests themselves, MacOS still failing during post-test cleanup so will ignore that for now as we have an open issue https://github.com/actions/setup-python/issues/857

PR #1252, opened 2024-05-08
Add an additional test for fast-state-dict request problem when snapshot_every_n_steps > 1

andrewkho:
#1251 <- See this PR for more context on the bug. Adds an additional test to check that this still works when snapshot_every_n_steps > 1

### Changes
* New unit test 
-
-


gokulavasan (approved):
Looks like this test expands the coverage/testing with snapshot every n steps to be non-default value

andrewkho:
MacOS tests are failing during shutdown still, it's intermittent which is really annoying

PR #1254, opened 2024-05-08
remove label filters for test ci

andrewkho:
Our tests are not being shown on PR creation because something in the filters disables them when facebook-github-bot adds the CLA Signed tag. Just disable the filters to match pytorch/pytorch repo setup

### Changes
- Remove label filters on CI 
- Bump actions/setup-python from v4 -> v5
- Bump actions/checkout from v3 -> v4

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -15,9 +15,7 @@ on:
 
 jobs:
   test:
-    if:
-      ${{ github.repository_owner == 'pytorch' && (github.event.action != 'labeled' ||
-      startsWith(github.event.label.name, 'ciflow')) }}
+    if: ${{ github.repository_owner == 'pytorch' }}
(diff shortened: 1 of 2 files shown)


gokulavasan:
@huydhn Tagging you for review as you might have some context on why this was done this way? Or do you think it is good to remove.

andrewkho:
Also bumping versions here instead of here: #1249

huydhn (approved):
This looks fine to me.  From what I see, this seems to be a way that torchdata uses to manually control what is run on the PR.  Take `ci.yml` as an example, it says that the workflow would run when a PR is opened, synchronize (updated), reopened or when someone labels it with `ciflow/something`.  I don't have the context on why it was setup as such, but it's better to get it up and runner than racking our brain trying to figure out why :)

andrewkho:
Thanks @huydhn for weighing in!

PR #1255, opened 2024-05-09
[do not review] check if test_state_dict or test_dataloader is the culprit

andrewkho:
Since the CI failure cannot repro locally, pushing some commits to see what is causing the cleanup failures 

-
-


andrewkho:
Closing because fixed in https://github.com/pytorch/data/pull/1257

PR #1257, opened 2024-05-10
Shard StatefulDataLoader tests

### LINKED ISSUE #1256: MacOS state_dict tests in CI are failing during shutdown
### 🐛 Describe the bug

MacOS tests of StatefulDataLoader CI action fail intermittently during shutdown. on Mac it also takes a lot longer than both windows and ubuntu to shut down (10 minutes vs 1s). I'm not sure what causes Github Actions to mark the test as failed, but 
Created an issue here on actions/setup-python but still no response: https://github.com/actions/setup-python/issues/857 

Although we still get positive signals from the test, it shows up as an X in Github 

### Versions

Nightly / main branch in CI,

andrewkho:
Fixes #1256 

### Changes
* Rename test_state_dict tests so they are easy to shard in CI Actions
-
-


andrewkho:
Closing because sharding in https://github.com/pytorch/data/pull/1257 should fix the issue

andrewkho:
whoops closed the wrong PR

PR #1258, opened 2024-05-23
save iterator state once instead of storing it again in dataset state

gokulavasan:
Summary:

Test Plan:

Reviewers:

Subscribers:

Tasks:

Tags:

creating your pull request.

Fixes #{issue number}

### Changes

-
-


torchdata/stateful_dataloader/worker.py
```diff
@@ -213,10 +213,13 @@ def _worker_loop(
                                 _DATASET_ITER_STATE: try_to_serialize(fetcher.dataset_iter),  # type: ignore[union-attr]
                                 _FETCHER_ENDED: fetcher.ended,  # type: ignore[union-attr]
                             }
+                            # Pick up any user-defined dataset state if it is not the iterator as it is already captured in fetcher_state's dataset_iter_state
+                            dataset_state = try_to_serialize(dataset) if fetcher.dataset_iter is not dataset else None  # type: ignore[union-attr]
```
andrewkho, line 217: nice

andrewkho: Question: when they are the same, why choose iterable over dataset's state?

andrewkho (approved):
LGTM

torchdata/stateful_dataloader/worker.py
gokulavasan: dataset state works too honestly, I just picked one. I can switch it to dataset state as it is one level higher in the nesting dict hierarchy?

andrewkho: I can't quite articulate why but it seems like dataset state seems more correct, I can't think of a breaking example yet though

PR #1260, opened 2024-05-23
Update state_dict() request at time 0 to include worker state

andrewkho:
creating your pull request.

When state_dict is requested before an iterator is created, we currently return an empty dict. This does not reflect a true initial state of the workers, and may break assumptions made by DCP.

This update will guarantee that anytime state_dict is requested, you will get a "full" state_dict that includes main state, and all worker initial states, including dataset/iterator states. We achieve this by including an ack signal from the workers which includes an initial state_dict. The main process now waits for these acks and stores the initial states for any subsequent state_dict requests, as well as for resetting the iterator.

Fixes #{issue number}

### Changes
* Changes the way state_dict is handled before iterator is generated, and before any iteration begins. 
-
-

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -76,6 +76,7 @@ jobs:
       - name: Install test requirements
         run: pip3 install -r test/requirements.txt
       - name: Test documentation examples
+        if: matrix.os != 'windows-latest'
         run: |
           cd ./docs
           pip3 install -r requirements.txt
--- a/.github/workflows/stateful_dataloader_ci.yml
+++ b/.github/workflows/stateful_dataloader_ci.yml
@@ -72,15 +72,9 @@ jobs:
         run: |
           pip3 install .
         env:
-          BUILD_S3: 1
+          BUILD_S3: 0
       - name: Install test requirements
         run: pip3 install -r test/requirements.txt
-      - name: Test documentation examples
-        run: |
-          cd ./docs
-          pip3 install -r requirements.txt
-          make doctest
-          cd ..
       - name: Run StatefulDataLoader tests with pytest - dataloader
         if: ${{ ! contains(github.event.pull_request.labels.*.name, 'ciflow/slow') }}
         run: pytest --durations=0 --no-header -v test/stateful_dataloader/test_dataloader.py
--- a/test/stateful_dataloader/test_state_dict.py
+++ b/test/stateful_dataloader/test_state_dict.py
@@ -1228,7 +1228,7 @@ class TestInitialState_shard0(TestCase):
             next(it)
 
     def test_load_then_state(self):
-        for pw, interrupt in itertools.product([False, True], [2, 9]):
+        for pw in itertools.product([False, True]):
             num_workers = 4
             dataset = DummyMapDataset(100, shuffle=True)
             dl = StatefulDataLoader(
--- a/test/stateful_dataloader/test_state_dict.py
+++ b/test/stateful_dataloader/test_state_dict.py
@@ -103,15 +103,15 @@ class DummyIteratorIterableDataset(torch.utils.data.IterableDataset, Iterator, S
         return sample
 
     def state_dict(self):
-        sd = {"i": self.i}
+        sd = {"nest": {"i": self.i}}
         if self.include_generator:
-            sd["g"] = torch.get_rng_state()
+            sd["nest"]["g"] = torch.get_rng_state()
         return sd
 
     def load_state_dict(self, state_dict):
-        self.i = state_dict["i"]
+        self.i = state_dict["nest"]["i"]
         if self.include_generator:
-            torch.set_rng_state(state_dict["g"])
+            torch.set_rng_state(state_dict["nest"]["g"])
 
 
 class DummyIterableDataset(torch.utils.data.IterableDataset):
@@ -1266,6 +1266,111 @@ class TestDatasetIteratorStateDuplication_shard0(TestCase):
             else:
                 self.assertEqual(state_dict["dataset_state"], None)
                 self.assertTrue(state_dict["fetcher_state"]["dataset_iter_state"])
+
+
+class PeriodicStateIterableDataset(torch.utils.data.IterableDataset):
+    def __init__(self):
+        self.state = {
+            0: 0,
+            1: 0,
+            2: 0,
+            3: 0,
+        }
+        self.step = 0
+        self.limit = 100
+
+    def __iter__(self):
+        for _ in range(self.step, self.limit):
+            self.step += 1
+            y = self.state[self.step % len(self.state)]
+            self.state[self.step % len(self.state)] = 1 - y
+
+            yield self.state[self.step % len(self.state)]
+
+    def state_dict(self):
+        return {
+            "state": self.state,
+            "step": self.step,
+        }
+
+    def load_state_dict(self, state):
+        self.state = state["state"]
+        self.step = state["step"]
+
+
+class TestFastStateDictRequestRoundRobin_shard3(TestCase):
+    def _run_test(self, snapshot_every_n_steps, interrupt):
+        num_workers = 4
+        dataset = PeriodicStateIterableDataset()
+
+        dl = StatefulDataLoader(
+            dataset=dataset,
+            num_workers=num_workers,
+            batch_size=1,
+            collate_fn=identity,
+            persistent_workers=True,
+            multiprocessing_context="forkserver" if IS_MACOS else None,
+            snapshot_every_n_steps=snapshot_every_n_steps,
+        )
+        it = iter(dl)
+        for _ in range(interrupt):
+            next(it)
+
+        state_dict = dl.state_dict()
+        for _ in range(2):
+            next(it)
+        exp_state_dict = dl.state_dict()
+        exp = list(it)
+
+        dl.load_state_dict(state_dict)
+        # new iter after load_state_dict, ask for state dict before num_workers batches
+        # are yielded to ensure old worker states are stored properly
+        it = iter(dl)
+        for _ in range(2):
+            next(it)
+
+        state_dict2 = dl.state_dict()
+
+        dataset = PeriodicStateIterableDataset()
+        dl = StatefulDataLoader(
+            dataset=dataset,
+            num_workers=num_workers,
+            batch_size=1,
+            collate_fn=identity,
+            persistent_workers=True,
+            multiprocessing_context="forkserver" if IS_MACOS else None,
+            snapshot_every_n_steps=snapshot_every_n_steps,
+        )
+        dl.load_state_dict(state_dict2)
+        data = list(dl)
+
+        self.assertEqual(data, exp)
+
+        dataset = PeriodicStateIterableDataset()
+        dl = StatefulDataLoader(
+            dataset=dataset,
+            num_workers=num_workers,
+            batch_size=1,
+            collate_fn=identity,
+            persistent_workers=True,
+            multiprocessing_context="forkserver" if IS_MACOS else None,
+            snapshot_every_n_steps=snapshot_every_n_steps,
+        )
+        dl.load_state_dict(state_dict)
+        it = iter(dl)
+        for _ in range(2):
+            next(it)
+
+        state_dict3 = dl.state_dict()
+        self.assertEqual(state_dict3, exp_state_dict)
+
+    def test_fast_state_dict_request(self) -> None:
+        # these test settings will trigger a failure if
+        # the worker/main incremental state_dicts are out of sync during initialization
+        self._run_test(1, 15)
+
+    def test_fast_state_dict_request_skip_steps(self) -> None:
+        self._run_test(17, 19)
 
 
 if __name__ == "__main__":
--- a/test/stateful_dataloader/test_state_dict.py
+++ b/test/stateful_dataloader/test_state_dict.py
@@ -986,6 +986,14 @@ class TestJsonSerDe_shard3(TestCase):
         self._run_test_map(3)
 
 
+class ErrorDataset(torch.utils.data.Dataset):
+    def __getitem__(self, index: int):
+        raise ValueError("Iteration error")
+
+    def __len__(self):
+        return 10
+
+
 class TestInitialState_shard0(TestCase):
     def test_initial_state(self):
         for pw in [False, True]:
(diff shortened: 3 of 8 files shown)


gokulavasan (approved):

torchdata/stateful_dataloader/__init__.py
```diff
@@ -4,19 +4,8 @@
...
-    _FETCHER_ENDED,
-    _FETCHER_STATE,
-    _flatten,
-    _IncrementalState,
-    _IncrementalWorkerState,
-    _Tombstone,
-    _unflatten,
-    _WORKER_ID,
-)
+from .sampler import BatchSampler, RandomSampler
```
andrewkho, line 7: I pulled this out here because of the way we inherit from torch.utils.data.DataLoader, we need these to be patched in order for state_dict to work with them, and importing here is sufficient. Side effects should be negligble, and should get tested by test_dataloader.py tests

gokulavasan (changes requested):
One question about understanding the reason for sending AckRequest even when the dataloader knows the worker states

test/stateful_dataloader/test_state_dict.py
```diff
@@ -39,15 +39,15 @@ def __next__(self):
         return sample
 
     def state_dict(self):
-        sd = {"i": self.i}
+        sd = {"nest": {"i": self.i}}
```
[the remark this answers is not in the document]
andrewkho: Yep I wanted to throw these in as additional tests against the incremental state_dict changes

test/stateful_dataloader/test_state_dict.py
```diff
@@ -1054,13 +1053,145 @@ def test_json_serde_multi_process_map(self):
...
+                collate_fn=identity,
+                multiprocessing_context="forkserver" if IS_MACOS else None,
+            )
+            dl.load_state_dict(state)
+            state0 = dl.state_dict()
+            self.assertEqual(state0, state)
+            data = list(dl)
+            self.assertEqual(data, exp)
+
+    def test_init_error(self):
```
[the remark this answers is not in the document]
andrewkho: Since we're messing with the worker initialization flow, I want to make sure that the init_error and regular errors are still being handled the same way

torchdata/stateful_dataloader/stateful_dataloader.py
```diff
@@ -232,27 +235,31 @@ def __iter__(self) -> "_BaseDataLoaderIter":
...
-            return {}
-        else:
-            return self._iterator.state_dict()
+            self._iterator = self._get_iterator()
+            self._initial_iter_for_state_dict = True
+        return self._iterator.state_dict()
 
     def load_state_dict(self, state_dict: Dict[str, Any]) -> None:
         self._iterator = None
+        self._initial_iter_for_state_dict = False
```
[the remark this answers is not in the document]
andrewkho: Actually for your example, I think I'd expect dl_state to be == state, which should be the existing behaviour, and yes setting it False here should ensure that's the case.

the _initial_iter_for_state_dict has a comment in the `__init__` above describing it's usage, but it's basically a micro-optimization to prevent re-creating an iterator if one was created solely for state_dict and persisten_workers = False, we know it was never returned so we can just return it without worrying about reset logic

torchdata/stateful_dataloader/stateful_dataloader.py
```diff
@@ -861,9 +868,28 @@ def __init__(self, loader, next_iter_state):
         _utils.signal_handling._set_worker_pids(id(self), tuple(w.pid for w in self._workers))  # type: ignore[misc]
         _utils.signal_handling._set_SIGCHLD_handler()
         self._worker_pids_set = True
-        self._snapshot, self._worker_snapshots, self._main_snapshots = {}, {}, collections.deque()  # type: ignore[var-annotated]
+        self._snapshot, self._main_snapshots = {}, collections.deque()  # type: ignore[var-annotated]
+        self._worker_snapshots: Dict[str, _IncrementalWorkerState] = {}
```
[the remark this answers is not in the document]
andrewkho: AckStartup always sends back an initial state, I think this is needed for this scenario, though code may have moved around since I last visited this:

```
dl = DataLoader(..., persistent_workers=True)
dl.load_state_dict({step: 2})
batches = list(dl)
batches2 = list(dl)
assert len(batches2) > batches
```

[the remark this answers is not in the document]
andrewkho: For comments, I left this in the worker.py, I'll tag you in there. Let me know if you think it would be better here, but at the very least I can add a reference here to worker.py

torchdata/stateful_dataloader/worker.py
```diff
@@ -111,16 +120,23 @@ def _worker_loop(
...
             if init_fn is not None:
                 init_fn(worker_id)
 
             fetcher = _DatasetKind.create_fetcher(dataset_kind, dataset, auto_collation, collate_fn, drop_last)
 
+            initial_state = _make_state_dict(worker_id, dataset_kind, fetcher, dataset)
+            # The main-process instantiates its IncrementalWorkerState with worker_state if present
+            # else initial_state. We can't always start with initial_state because of the case
+            # where we load a state_dict, call next for n < num_worker steps (ie do not receive updates)
+            # and then request another state_dict.
```
andrewkho, line 137: @gokulavasan comment on IncrementalWorkerState here

.github/workflows/stateful_dataloader_ci.yml
```diff
@@ -72,15 +72,9 @@ jobs:
...
-          BUILD_S3: 1
+          BUILD_S3: 0
       - name: Install test requirements
         run: pip3 install -r test/requirements.txt
-      - name: Test documentation examples
-        run: |
-          cd ./docs
-          pip3 install -r requirements.txt
-          make doctest
-          cd ..
```
andrewkho, line 83: we're running identical tests in ci.yml (datapipes tests)

.github/workflows/ci.yml
```diff
@@ -76,6 +76,7 @@ jobs:
       - name: Install test requirements
         run: pip3 install -r test/requirements.txt
       - name: Test documentation examples
+        if: matrix.os != 'windows-latest'
```
andrewkho, line 79: Temporarily disabling doctest for windows as the s3FileLister test is failing with no debuggable error, it should be unrelated to this diff and will not invest in fixing since it's deprecated

torchdata/stateful_dataloader/stateful_dataloader.py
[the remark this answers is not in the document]
gokulavasan: > Actually for your example, I think I'd expect dl_state to be == state, which should be the existing behaviour, and yes setting it False here should ensure that's the case.

I did a quick test: https://pastebin.com/PAxtEdTx and it looks like calling state_dict right after load_state_dict yields an empty dict.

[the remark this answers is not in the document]
andrewkho: sorry by existing behaviour, I meant what's in the code here, not what's in main @gokulavasan

torchdata/stateful_dataloader/stateful_dataloader.py
```diff
@@ -861,18 +870,30 @@ def __init__(self, loader, next_iter_state):
...
+        # before n >= num_workers steps are taken.
+        self._worker_snapshots: Dict[str, _IncrementalWorkerState] = {}
         self._reset(loader, first_iter=True, prime_prefetch=next_iter_state is None)
 
         # Try to restore main state
         if next_iter_state is not None:
             self._restore_main_state(next_iter_state[self._SNAPSHOT][self._MAIN_SNAPSHOT])
             self._num_yielded = next_iter_state[self._SNAPSHOT][self._SNAPSHOT_STEP]
 
-            # Back-fill the worker snapshots before starting, in case of failure before a full cycle
```
andrewkho, line 874: I moved this in to reset call

torchdata/stateful_dataloader/stateful_dataloader.py
```diff
@@ -940,17 +962,41 @@ def _reset(self, loader, first_iter=False, prime_prefetch=True):
         self._workers_status = [True for i in range(self._num_workers)]
         # Reset the worker queue cycle so it resumes next epoch at worker 0
         self._worker_queue_idx_cycle = itertools.cycle(range(self._num_workers))
-        # We resume the prefetching in case it was enabled
-        if not first_iter:
+        worker_states = {}
+        if first_iter:
```
andrewkho, line 966: All logic for getting "initial" states from workers is here, for first iter and for resumes

torchdata/stateful_dataloader/worker.py
```diff
@@ -111,16 +120,20 @@ def _worker_loop(
...
-
+        initial_state = None
         try:
             if init_fn is not None:
                 init_fn(worker_id)
 
             fetcher = _DatasetKind.create_fetcher(dataset_kind, dataset, auto_collation, collate_fn, drop_last)
 
+            # See NOTE [ Incremental worker state ]
+            initial_state = worker_state or _make_state_dict(worker_id, dataset_kind, fetcher, dataset)
```
andrewkho, line 134: I'm keeping the flow identical for now between worker_state == None and != None to reduce the differences in code paths between the two cases, and easier to reason about that main and worker start with the same state. We can optimize this in a subsequent diff by not sending worker_state back if we just got it from the main proc, which will be a trivial change, but IMO makes it a little harder to follow

torchdata/stateful_dataloader/stateful_dataloader.py
```diff
@@ -909,16 +930,17 @@ def __init__(self, loader, next_iter_state):
...
+        self._worker_snapshots = {key: _IncrementalWorkerState(state) for key, state in worker_states.items()}
         self._main_snapshots = collections.deque()
         self._last_yielded_worker_id = self._num_workers - 1
         self._update_snapshot(
             snapshot_step=0,
             last_yielded_worker_id=self._num_workers - 1,
             num_workers=self._num_workers,
-            main_snapshot=self._main_state_0,
-            worker_snapshots={},
+            main_snapshot=self._get_main_state(),
```
gokulavasan, line 942: nice catch!

torchdata/stateful_dataloader/stateful_dataloader.py
```diff
@@ -861,18 +870,30 @@ def __init__(self, loader, next_iter_state):
...
             self._num_yielded = next_iter_state[self._SNAPSHOT][self._SNAPSHOT_STEP]
 
-            # Back-fill the worker snapshots before starting, in case of failure before a full cycle
-            self._worker_snapshots = {key: _IncrementalWorkerState(state) for key, state in worker_states.items()}
+            self._update_snapshot(
+                snapshot_step=next_iter_state[self._SNAPSHOT][self._SNAPSHOT_STEP],
+                last_yielded_worker_id=next_iter_state[self._SNAPSHOT][self._LAST_YIELDED_WORKER_ID],
+                num_workers=self._num_workers,
+                main_snapshot=next_iter_state[self._SNAPSHOT][self._MAIN_SNAPSHOT],
+                worker_snapshots=self._worker_snapshots,
```
gokulavasan, line 895: where is the worker states extracted from the next_iter_state?

torchdata/stateful_dataloader/stateful_dataloader.py
gokulavasan: To make sure, I understand the outcome of the changes - for the case when load_state_dict is called before iter(dl), ie:

```
dl = StatefulDataLoader(...)
dl.load_state_dict(state_from_somewhere)
it = iter(dl)
print(next(it))
```

In this case, does AckStartup get sent out explicitly to get the initial worker state?

torchdata/stateful_dataloader/stateful_dataloader.py
andrewkho: They're initialized in `_reset`

torchdata/stateful_dataloader/stateful_dataloader.py
```diff
@@ -909,16 +930,17 @@ def __init__(self, loader, next_iter_state):
                 next(self)
             self._finished = next_iter_state[_ITERATOR_FINISHED]
 
-    def _reset_state_vars(self):
-        self._worker_snapshots = {}
+    def _reset_state_vars(self, worker_states):
+        assert len(worker_states) == self._num_workers, (self._num_workers, worker_states)
+        self._worker_snapshots = {key: _IncrementalWorkerState(state) for key, state in worker_states.items()}
```
andrewkho, line 935: This is where the worker snapshots are initialized now, worker_states will either contain t0 worker state or whatever is being loaded

torchdata/stateful_dataloader/stateful_dataloader.py
andrewkho: Yes, AckStartup is always sent (My preference is to have fewer logic branches to keep the code understandable as much as possible

torchdata/stateful_dataloader/worker.py
```diff
@@ -111,16 +120,22 @@ def _worker_loop(
...
+            incremental_worker_state = _IncrementalWorkerState(initial_state)
+            if initial_state is worker_state:
+                initial_state = None
```
andrewkho, line 137: This stops us from passing worker_state back to main process

torchdata/stateful_dataloader/stateful_dataloader.py
gokulavasan: So AckStartup is sent only once at the beginning of the first epoch. When iter(dl) is called subsequently, AckStartup is not sent, is that correct?

test/stateful_dataloader/test_state_dict.py
```diff
@@ -1054,13 +1089,186 @@ def test_json_serde_multi_process_map(self):
...
+        for pw, interrupt in itertools.product([False, True], [2, 9]):
+            num_workers = 4
+            dataset = DummyMapDataset(100, shuffle=True)
+            dl = StatefulDataLoader(
+                dataset=dataset,
+                num_workers=num_workers,
+                persistent_workers=pw,
+                collate_fn=identity,
+                multiprocessing_context="forkserver" if IS_MACOS else None,
+                pin_memory=TEST_CUDA,
```
gokulavasan, line 1146: For this explicit pin memory setting, is the pin_memory=True tested on CI with a gpu host?

gokulavasan:
Two questions, rest lgtm! Also lint check is failing

test/stateful_dataloader/test_state_dict.py
andrewkho: Yep

torchdata/stateful_dataloader/stateful_dataloader.py
andrewkho: It's sent on _reset as well

gokulavasan (approved):

PR #1264, opened 2024-05-30
Implement incremental state transfer from worker to dataloader process #1262

gokulavasan:
Summary: Transfer of full state from worker to main process can be expensive if the state is very large. In order to alleviate the cost of state transfer, maintain a flattened state map in the worker process and send only the key/value that have changed to the main dataloader process where the full state is reconstructed with the changes received from the workers.

Test Plan: Unit tests

Reviewers:

Subscribers:

Tasks:

Tags:Please read through our [contribution guide](https://github.com/pytorch/data/blob/main/CONTRIBUTING.md) prior to
creating your pull request.

Fixes #{issue number}

### Changes

-
-

### DIFF
--- a/test/stateful_dataloader/test_state_dict.py
+++ b/test/stateful_dataloader/test_state_dict.py
@@ -512,8 +512,6 @@ class TestStatefulDataLoaderGeneratorNoState_shard2(TestStatefulDataLoaderIterab
         for _ in range(interrupt):
             batches.append(next(it))
         state_dict = dl.state_dict()
-        print("Got sd: ", state_dict)
-
         self.assertEqual(batches, exp[:interrupt])
 
         # Restore new instance from state
--- a/test/stateful_dataloader/test_state_dict.py
+++ b/test/stateful_dataloader/test_state_dict.py
@@ -7,6 +7,7 @@
 import itertools
 import json
 import unittest
+from copy import deepcopy
 from typing import Iterator
 
 import torch
@@ -181,7 +182,7 @@ class DynamicStateIterableDataset(torch.utils.data.IterableDataset):
     def state_dict(self):
         state = {"i": self.i}
         for a in range(self.i):
-            state[str(a)] = a
+            state[str(a)] = {a: list(range(a))}
             state[f"t{str(a)}"] = torch.tensor(a, dtype=torch.int8)
         return state
 
@@ -1064,40 +1065,48 @@ class TestStatefulDataLoaderIterable2_shard3(TestStatefulDataLoaderIterable_shar
 class TestDynamicStateIterableDataset(TestCase):
     def test(self):
         dataset = DynamicStateIterableDataset(list(range(100)))
-        for num_workers in [2]:
-            dl = StatefulDataLoader(
-                dataset=dataset,
-                num_workers=num_workers,
-                collate_fn=identity,
-                multiprocessing_context="forkserver" if IS_MACOS and num_workers else None,
-            )
-            it = iter(dl)
-            # Fetch at least one batch from each worker
-            for _ in range((num_workers + 1) * 2):
-                next(it)
-            state_dict = dl.state_dict()
-            worker_state = state_dict["_snapshot"]["_worker_snapshots"]["worker_0"]["fetcher_state"][
-                "dataset_iter_state"
-            ]
-            self.assertEqual(len(worker_state), 7)
+        num_workers = 2
+        dl = StatefulDataLoader(
+            dataset=dataset,
+            num_workers=num_workers,
+            collate_fn=identity,
+            multiprocessing_context="forkserver" if IS_MACOS and num_workers else None,
+        )
+        it = iter(dl)
+        # Fetch at least one batch from each worker
+        for _ in range((num_workers + 1) * 2):
+            next(it)
+        state_dict = dl.state_dict()
+        worker_state = state_dict["_snapshot"]["_worker_snapshots"]["worker_0"]["fetcher_state"]["dataset_iter_state"]
+        self.assertEqual(len(worker_state), 7)
+        deep_copy_state_dict = deepcopy(state_dict)
 
-            dl = StatefulDataLoader(
-                dataset=dataset,
-                num_workers=num_workers,
-                collate_fn=identity,
-                multiprocessing_context="forkserver" if IS_MACOS and num_workers else None,
-            )
-            dl.load_state_dict(state_dict)
-            it = iter(dl)
-            exp = []
-            for _ in range(2):
-                exp.extend(next(it))
-            state_dict = dl.state_dict()
-            self.assertEqual(exp, [3, 3])
-            worker_state = state_dict["_snapshot"]["_worker_snapshots"]["worker_0"]["fetcher_state"][
-                "dataset_iter_state"
-            ]
-            self.assertEqual(len(worker_state), 9)
+        # Iterate a few more steps and ensure earlier state_dict hasn't changed
+        for _ in range(num_workers * 2):
+            next(it)
+        next_state_dict = dl.state_dict()
+        self.assertEqual(state_dict, deep_copy_state_dict)
+        self.assertFalse(state_dict == next_state_dict)
+        worker_state = next_state_dict["_snapshot"]["_worker_snapshots"]["worker_0"]["fetcher_state"][
+            "dataset_iter_state"
+        ]
+        self.assertEqual(len(worker_state), 11)
+
+        dl = StatefulDataLoader(
+            dataset=dataset,
+            num_workers=num_workers,
+            collate_fn=identity,
+            multiprocessing_context="forkserver" if IS_MACOS and num_workers else None,
+        )
+        dl.load_state_dict(state_dict)
+        it = iter(dl)
+        exp = []
+        for _ in range(num_workers):
+            exp.extend(next(it))
+        state_dict = dl.state_dict()
+        self.assertEqual(exp, [3, 3])
+        worker_state = state_dict["_snapshot"]["_worker_snapshots"]["worker_0"]["fetcher_state"]["dataset_iter_state"]
+        self.assertEqual(len(worker_state), 9)
 
 
 class TestDatasetIteratorStateDuplication_shard3(TestCase):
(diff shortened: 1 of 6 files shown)


torchdata/stateful_dataloader/incremental_state.py
```diff
@@ -0,0 +1,178 @@
...
+_WORKER_ID = "worker_id"
+_FETCHER_STATE = "fetcher_state"
+_FETCHER_ENDED = "fetcher_ended"
+_DATASET_STATE = "dataset_state"
+_DATASET_ITER_STATE = "dataset_iter_state"
+
+
+class Flattener:
+    @classmethod
+    def flatten(cls, nested_data: Optional[Dict[str, Any]], prefix: str = "", separator: str = "/"):
```
andrewkho, line 19: Can we just make these plain functions?

torchdata/stateful_dataloader/incremental_state.py
```diff
@@ -0,0 +1,178 @@
...
+        if nested_data is None:
+            return None
+        data = {}
+        if isinstance(nested_data, dict):
+            for key, value in nested_data.items():
+                flat = cls.flatten(
+                    value,
+                    prefix=f"{prefix}{separator}{key}" if len(prefix) > 0 else key,
+                    separator=separator,
```
andrewkho, line 28: Would this break if users have keys with `/`?

andrewkho: Another way to do this that might be more robust is to pass this in as expanding tuples, which are also immutable:

```
prefix: Tuple[...]
flatten(value, prefix=prefix + (key,)
```

andrewkho: There's also an assumption here that keys are a) strings, b) immutable

andrewkho: (responding to myself) I think if it's usable as a key in a dictionary, it should be immutable, see: https://docs.python.org/3/reference/datamodel.html#object.__hash__

torchdata/stateful_dataloader/incremental_state.py
```diff
@@ -0,0 +1,178 @@
...
+
+        for key in all_keys:
+            if self.flat_state is None or key not in self.flat_state:
+                # New key, retain it
+                delta_flat_state[key] = new_flat_state[key]
+                continue
+
+            if key not in new_flat_state:
+                # Key deletion, put in a tombstone
+                delta_flat_state[key] = DeletionTombStone()
```
andrewkho, line 84: do we need all other nested keys to also have TombStone?

torchdata/stateful_dataloader/incremental_state.py
```diff
@@ -0,0 +1,178 @@
...
+                continue
+
+            prev_value, new_value = self.flat_state[key], new_flat_state[key]
+            try:
+                if isinstance(prev_value, torch.Tensor) and isinstance(new_value, torch.Tensor):
+                    if torch.allclose(prev_value, new_value):
```
andrewkho, line 90: I wonder if we should actually use down-to-the-bit equality here, since we do actually want to pick up any changes, even if noise

torchdata/stateful_dataloader/incremental_state.py
```diff
@@ -0,0 +1,178 @@
...
+        return delta_flat_state
+
+    def apply_delta(self, flat_delta_state: Dict[str, Any]) -> None:
+        for key, update in flat_delta_state.items():
+            if self.flat_state is None:
+                self.flat_state = {}
+
+            if isinstance(update, DeletionTombStone):
+                # Remove key if present in the state
+                self.flat_state.pop(key, None)
```
andrewkho, line 109: Oh I see, I think this will drop the entire branch so no need to manually delete everything. Wondering how that case actually plays out though

torchdata/stateful_dataloader/incremental_state.py
```diff
@@ -0,0 +1,178 @@
...
+                _FETCHER_ENDED: self._fetcher_ended,
+            }
+        return incr_state_dict
+
+    def apply_delta(self, delta_state_dict: Dict[str, Any]) -> None:
+        self._worker_id = delta_state_dict[_WORKER_ID]
+        if ds_state := delta_state_dict.get(_DATASET_STATE, None):
+            self._incr_dataset_state.apply_delta(ds_state)
+
+        if fetcher_state := delta_state_dict.get(_FETCHER_STATE, None):
```
andrewkho, line 160: note: walrus requires Python >= 3.8, not sure if we are supporting older python versions

torchdata/stateful_dataloader/incremental_state.py
```diff
@@ -0,0 +1,178 @@
...
+            self._fetcher_ended = fetcher_state[_FETCHER_ENDED]
+            if iter_state := fetcher_state.get(_DATASET_ITER_STATE, None):
+                self._incr_fetcher_iter_state.apply_delta(iter_state)
+
+    def get_state(self) -> Dict[str, Any]:
```
andrewkho, line 165: Meta question: shoudl we occasionally send the full state / hash of the true state so that main process can throw an exception if things start to go out of sync?

andrewkho (changes requested):
This looks really good, thanks Gokul! Some minor comments/update requests.

torchdata/stateful_dataloader/incremental_state.py
```diff
@@ -0,0 +1,178 @@
...
+_DATASET_ITER_STATE = "dataset_iter_state"
+
+
+class Flattener:
+    @classmethod
+    def flatten(cls, nested_data: Optional[Dict[str, Any]], prefix: str = "", separator: str = "/"):
+        if nested_data is None:
+            return None
+        data = {}
+        if isinstance(nested_data, dict):
```
andrewkho, line 23: One thought: how much complexity would it add to support dataclasses or plain python objects?

andrewkho: https://stackoverflow.com/questions/60713703/what-is-a-better-way-to-iterate-over-dataclass-keys-and-values Here is an example of iterating over dataclass fields

torchdata/stateful_dataloader/incremental_state.py
gokulavasan: pytorch core recommendation seems to be 3.8+ - https://pytorch.org/get-started/locally/#linux-python. Also our CI runs only for 3.8-3.10. So I think this is alright?

torchdata/stateful_dataloader/incremental_state.py
gokulavasan: For dataclasses as state, one option users have is to use dataclasses asdict method to convert it to state_dict anyway. Thus handling nested dicts covers decent ground

torchdata/stateful_dataloader/incremental_state.py
gokulavasan: That's a good point, I will switch to use torch.equal

torchdata/stateful_dataloader/incremental_state.py
gokulavasan: Since the passed in state is flattened first in this function, all those nested keys won't be present and so they will all get the Tombstone

torchdata/stateful_dataloader/incremental_state.py
gokulavasan: Can you elaborate on this a bit more - entire branch as in the child nodes in a parent dict? The state in this function is always maintained as a flat_dict. The delta state generated so contains flat keys. So deletion iterates through all the flat keys and deletes it from the flat_state. 

Are you suggesting that we don't have to delete all these keys instead just have a marked for the top level key deletion and then delete all the child keys based on it?

torchdata/stateful_dataloader/incremental_state.py
gokulavasan: Is there guarantee that the object will be hashable? Open to ideas here which can be implemented in subsequent PRs

gokulavasan:
@andrewkho Thanks for the review! Addressed/responded to the comments, please take a look again, thanks!

torchdata/stateful_dataloader/incremental_state.py
andrewkho: makes sense, just wanted to flag so we know explicitly we have 3.8+ only features

torchdata/stateful_dataloader/incremental_state.py
andrewkho: I assume we can do an MD5 of any bitwise representation so eventually at some level it must be hashable, but not in the __hash__ sense

andrewkho: yes subsequent PR definitely makes sense

test/stateful_dataloader/test_state_dict.py
```diff
@@ -481,6 +512,7 @@ def _run_and_checkpoint(self, num_workers, batch_size, pw, interrupt, every_n_st
         for _ in range(interrupt):
             batches.append(next(it))
         state_dict = dl.state_dict()
+        print("Got sd: ", state_dict)
```
andrewkho, line 515:
```suggestion
```

torchdata/stateful_dataloader/incremental_state.py
```diff
@@ -0,0 +1,169 @@
...
+import torch
+
+_WORKER_ID = "worker_id"
+_FETCHER_STATE = "fetcher_state"
+_FETCHER_ENDED = "fetcher_ended"
+_DATASET_STATE = "dataset_state"
+_DATASET_ITER_STATE = "dataset_iter_state"
+
+
+def _flatten(nested_data, key_lineage=()):
```
andrewkho, line 17: nit: type hints

torchdata/stateful_dataloader/incremental_state.py
```diff
@@ -0,0 +1,169 @@
...
+    if nested_data is None:
+        return None
+    data = {}
+    if isinstance(nested_data, dict):
+        for key, value in nested_data.items():
+            flat = _flatten(value, key_lineage + (key,))
+            data.update(flat)
```
andrewkho, line 24: I still think we could add dataclass support but let's do it in a subsequent diff. I am unsure if there is significant overhead of dataclasses.asdict, since it appears they're making copies, eg

andrewkho (approved):
LGTM!

PR #1266, opened 2024-06-04
Incremental state: fix flatten None handling

gokulavasan:
Flatten could return None but result of flatten was recursively used for updating the flatten map. This caused an error when the value of the dict is None. Handling that case now and also added unit tests to capture that scenario

Fixes #{issue number}

### Changes

-
-

### DIFF
--- a/test/stateful_dataloader/test_incremental_state.py
+++ b/test/stateful_dataloader/test_incremental_state.py
@@ -27,24 +27,34 @@ from torchdata.stateful_dataloader import (
 class TestFlattener(TestCase):
     def test(self):
         test_dict_pairs = [
-            (None, 0),
-            ({}, 0),
+            (None, 1),
+            ({}, 1),
             ({"kt": torch.rand(2, 2)}, 1),
             ({"k1": "v1"}, 1),
             ({"k2": ["v1", "v2"]}, 1),
             ({"k1": {"k2": "v2"}}, 1),
             ({"k1": {"k2": "v2"}, "k2": {"k2": "v2"}}, 2),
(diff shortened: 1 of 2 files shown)


torchdata/stateful_dataloader/incremental_state.py
```diff
@@ -16,7 +16,7 @@
 
 
 def _flatten(nested_data: Any, key_lineage: Tuple = ()):
-    if nested_data is None:
+    if nested_data is None and len(key_lineage) == 0:
         return None
```
andrewkho, line 20: Is there a way to have better guarantees on the return type? ie we always return a dict so we know that eg subsequent calls to items() are always valid

gokulavasan: Addressed this in the updated commit by making flatten always return Dict. Also added unit tests to cover such cases

andrewkho (approved):
Thank you for the quick fix!

test/stateful_dataloader/test_incremental_state.py
```diff
@@ -27,24 +27,34 @@
...
+            ({"k1": {"k2": None}}, 1),
+            ("abcd", 1),
+            ({"k1": {"k2": {}, "k3": None}}, 2),
         ]
 
         for kv, flat_key_count in test_dict_pairs:
             flat_dict = _flatten(kv)
             nest_dict = _unflatten(flat_dict)
+            print(flat_dict)
+            print(nest_dict)
```
andrewkho, line 47: We should delete these, let's do it after merge to land the fix sooner

PR #1272, opened 2024-06-12
Handle case where state is a tensor

gokulavasan:
Fixes the case where tensor is returned as the state instead of a dict. This was failing because the tensor object was being used as a bool check for the walrus operator and that failed. Instead just do regular fetch and not None comparison.

Added unit tests in all the layers to cover this

Fixes #{1271}

### Changes

-
-

### DIFF
--- a/test/stateful_dataloader/test_state_dict.py
+++ b/test/stateful_dataloader/test_state_dict.py
@@ -8,6 +8,8 @@ import itertools
 import json
 import unittest
 from copy import deepcopy
+
+from enum import Enum
 from typing import Iterator
 
 import torch
@@ -424,6 +426,32 @@ class GeneratorIterable(torch.utils.data.IterableDataset):
     def load_state_dict(self, state):
         self.resume = state["i"]
         self.epoch = state["epoch"]
+
+
+class StateType(Enum):
+    INTEGER = (1,)
+    STRING = (2,)
+    TENSOR = (3,)
+
+
+class TensorStateIterableDataset(torch.utils.data.IterableDataset, Stateful):
+    def __init__(self, len, state_type):
+        self.len = len
+        self.state_type = state_type
+
+    def __iter__(self):
+        return iter(range(self.len))
+
+    def load_state_dict(self, state):
+        pass
+
+    def state_dict(self):
+        if self.state_type == StateType.INTEGER:
+            return self.len
+        if self.state_type == StateType.STRING:
+            return str(self.len)
+        if self.state_type == StateType.TENSOR:
+            return torch.rand(self.len)
 
 
 class GeneratorIterableNoState(torch.utils.data.IterableDataset):
@@ -1448,6 +1476,22 @@ class TestFastStateDictRequestRoundRobin_shard3(TestCase):
 
     def test_fast_state_dict_request_skip_steps(self) -> None:
         self._run_test(17, 19)
+
+
+class TestNonDictState_shard0(TestCase):
+    def test(self):
+        for state_type in StateType:
+            ds = TensorStateIterableDataset(5, state_type)
+            dl = StatefulDataLoader(
+                dataset=ds,
+                num_workers=2,
+                collate_fn=identity,
+                multiprocessing_context=("forkserver" if IS_MACOS and num_workers else None),
+            )
+            it = iter(dl)
+            next(it)
+            sd = dl.state_dict()
+            self.assertTrue(sd)
 
 
 class TestMultiEpochState_shard0(TestCase):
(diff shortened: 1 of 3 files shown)


gokulavasan:
Fix is in this PR - https://github.com/pytorch/data/pull/1272

test/stateful_dataloader/test_state_dict.py
```diff
@@ -426,6 +428,32 @@ def load_state_dict(self, state):
         self.epoch = state["epoch"]
 
 
+class StateType(Enum):
+    INTEGER = (1,)
+    STRING = (2,)
+    TENSOR = (3,)
```
andrewkho, line 434: I don't really care because this is just a test, but generally for python just use a string, I find this boilerplate just adds friction and doesn't usually help

torchdata/stateful_dataloader/incremental_state.py
```diff
@@ -120,7 +120,8 @@ def __init__(self, initial_worker_state_dict: Optional[Dict[str, Any]]):
         if initial_worker_state_dict:
             self._worker_id = initial_worker_state_dict[_WORKER_ID]
             dataset_state = initial_worker_state_dict.get(_DATASET_STATE, None)
-            if fetcher_state := initial_worker_state_dict.get(_FETCHER_STATE, None):
+            fetcher_state = initial_worker_state_dict.get(_FETCHER_STATE, None)
+            if fetcher_state is not None:
```
andrewkho, line 124: Why this change?

test/stateful_dataloader/test_state_dict.py
```diff
@@ -1450,6 +1478,23 @@ def test_fast_state_dict_request_skip_steps(self) -> None:
...
+                dl = StatefulDataLoader(
+                    dataset=ds,
+                    num_workers=num_workers,
+                    collate_fn=identity,
+                    multiprocessing_context=("forkserver" if IS_MACOS and num_workers else None),
+                )
+                it = iter(dl)
+                next(it)
+                sd = dl.state_dict()
+                self.assertTrue(sd)
```
andrewkho, line 1495: what is this testing?

test/stateful_dataloader/test_state_dict.py
andrewkho: Oh I see from summary now

torchdata/stateful_dataloader/incremental_state.py
andrewkho: Oh I see from summary now

test/stateful_dataloader/test_state_dict.py
andrewkho:
```suggestion
```

test/stateful_dataloader/test_state_dict.py
```diff
@@ -426,6 +428,32 @@ def load_state_dict(self, state):
...
+    def load_state_dict(self, state):
+        pass
+
+    def state_dict(self):
+        if self.state_type == StateType.INTEGER:
+            return self.len
+        if self.state_type == StateType.STRING:
+            return str(self.len)
+        if self.state_type == StateType.TENSOR:
+            return torch.rand(self.len)
```
andrewkho, line 454:
```suggestion
        return self.len
```

test/stateful_dataloader/test_state_dict.py
```diff
@@ -426,6 +428,32 @@ def load_state_dict(self, state):
...
+class StateType(Enum):
+    INTEGER = (1,)
+    STRING = (2,)
+    TENSOR = (3,)
+
+
+class TensorStateIterableDataset(torch.utils.data.IterableDataset, Stateful):
+    def __init__(self, len, state_type):
+        self.len = len
+        self.state_type = state_type
```
andrewkho, line 440:
```suggestion
    def __init__(self, len):
        self.len = len
```

test/stateful_dataloader/test_state_dict.py
```diff
@@ -1450,6 +1478,23 @@ def test_fast_state_dict_request_skip_steps(self) -> None:
         self._run_test(17, 19)
 
 
+class TestNonDictState_shard0(TestCase):
+    def test(self):
+        for state_type in StateType:
```
andrewkho, line 1483:
```suggestion
        for ctr in [int, str, torch.rand]:
```

test/stateful_dataloader/test_state_dict.py
```diff
@@ -1450,6 +1478,23 @@ def test_fast_state_dict_request_skip_steps(self) -> None:
...
+            for num_workers in [0, 2]:
+                ds = TensorStateIterableDataset(5, state_type)
```
andrewkho, line 1485:
```suggestion
                ds = TensorStateIterableDataset(ctr(5))
```

andrewkho (approved):
LGTM please simplify the test per instructions before landing and maybe beef up the assert statement a bit

PR #1273, opened 2024-06-12
fix for iter called twice

### LINKED ISSUE #1267: iter(dataset) is called twice for certain cases of state restore of IterableDataset
### 🐛 Describe the bug

Consider this code:
```
class DatasetStateIterable(torch.utils.data.IterableDataset, Stateful):
    def __init__(self, length):
        self.length = length

    def __iter__(self):
        print("Calling iter")
        return iter(list(range(self.length)))

    def state_dict(self):
        return {"key": "value"}

        pass

class TestSimple(TestCase):
    def test(self):
        dataset = DatasetStateIterable(100)
        dl = StatefulDataLoader(
            dataset=dataset,
        )
        it = iter(dl)
        print(next(it))
        state = dl.state_dict()
        print(state)
        del dl
        print("Creating a new dl")
        dl2 = StatefulDataLoader(
            dataset=dataset,
        )
        dl2.load_state_dict(state)
        it2 = iter(dl2)
        print(next(it))
        print(dl2.state_dict())
        self.assertTrue(False)
```

The above code calls iter on the dataset twice as seen below after the state restore. The output of the above looks as follows. Note that this also happens when num_workers > 0 due to recreation here (https://github.com/pytorch/data/blob/82918dd12281e69d39fa80bfeabfc02b1f93561c/torchdata/stateful_dataloader/worker.py#L122-L134):

```
Calling iter
tensor([0])
{'_index_sampler_state': ...}
Creating a new dl
Calling iter
Calling iter
tensor([1])
{'_index_sampler_state': ...}
```

### Versions

As of commit 82918dd

andrewkho:
creating your pull request.

Fixes #1267 

### Changes

-
-

### DIFF
--- a/test/stateful_dataloader/test_state_dict.py
+++ b/test/stateful_dataloader/test_state_dict.py
@@ -1491,5 +1491,64 @@ class TestMultiEpochState_shard0(TestCase):
         self._run(True, 4)
 
 
+class CountIterCalls(torch.utils.data.IterableDataset):
+    def __init__(self, length):
+        self.length = length
+        self.iter_calls = 0
+
+    def __iter__(self):
+        self.iter_calls += 1
+        return iter(list(range(self.length)))
+
+    def state_dict(self):
+        return {"iter_calls": self.iter_calls}
+
+    def load_state_dict(self, state_dict):
+        pass
+
+
+class TestSingleIterCalled_shard0(TestCase):
+    def _get_iter_calls(self, state):
+        if "dataset_state" in state:
+            w_states = [state]
+        else:
+            w_states = list(state["_snapshot"]["_worker_snapshots"].values())
+
+        return [x["dataset_state"]["iter_calls"] for x in w_states]
+
+    def _run_test(self, num_workers):
+        dataset = CountIterCalls(100)
+        dl = StatefulDataLoader(
+            dataset=dataset,
+            num_workers=num_workers,
+            multiprocessing_context=("forkserver" if IS_MACOS and num_workers else None),
+        )
+        iter(dl)
+        state = dl.state_dict()
+        if num_workers == 0:
+            self.assertEqual(dataset.iter_calls, 1)
+        else:
+            self.assertEqual(self._get_iter_calls(state), [1] * num_workers)
+        dl2 = StatefulDataLoader(
+            dataset=dataset,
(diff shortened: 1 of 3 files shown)


test/stateful_dataloader/test_state_dict.py
```diff
@@ -1491,5 +1491,64 @@ def test_pw(self):
...
+            num_workers=num_workers,
+            multiprocessing_context=("forkserver" if IS_MACOS and num_workers else None),
+        )
+        dl2.load_state_dict(state)
+        iter(dl2)
+        state2 = dl.state_dict()
+        if num_workers == 0:
+            self.assertEqual(dataset.iter_calls, 2)
+        else:
+            print(state2)
```
gokulavasan, line 1543: nit: remove; we have to put in a linter to catch print statements in unit tests

test/stateful_dataloader/test_state_dict.py
```diff
@@ -1491,5 +1491,64 @@ def test_pw(self):
...
+            self.assertEqual(self._get_iter_calls(state2), [1] * num_workers)
```
gokulavasan, line 1544: Hm this remains 1 in the multiprocess case because we are not getting the state0 back to main process when we load from a state?

test/stateful_dataloader/test_state_dict.py
```diff
@@ -1491,5 +1491,64 @@ def test_pw(self):
         self._run(True, 4)
 
 
+class CountIterCalls(torch.utils.data.IterableDataset):
```
gokulavasan, line 1494: Would it be possible to add another iterable dataset which returns itself as the iterator (similar to DummyIteratorIterableDataset) so that flow is also tested?

gokulavasan (approved):
Request to expand testing to cover another case and a question about state0 optimization

test/stateful_dataloader/test_state_dict.py
andrewkho: I suspect this is because of the change to not send worker state back after initialization

andrewkho: Discussing offline, my suggestion is we always send worker state back as this is a correctness issue

gokulavasan:
lg!

PR #1274, opened 2024-06-12
Add huggingface regression test

andrewkho:
Test against HF datasets main to catch regressions against https://github.com/huggingface/datasets/pull/6658 

### Changes
* adds HF test
* datasets req in test/requirements to pull from github main
-
-


gokulavasan (approved):
nice! lg!

PR #1278, opened 2024-06-18
Adding state parser utility that can be used for retrieving and modifying worker states

gokulavasan:
Summary: Allows users to get the worker states out of the state dict. It can be used if users want to modify the state offline. Starts off with very preliminary set of methods aimed at supporting usecases where number of workers need to change. 

Test Plan: Unit tests

Reviewers:

Subscribers:

Tasks:

Tags:

creating your pull request.

Fixes #{issue number}

### Changes

-
-

### DIFF
--- a/torchdata/stateful_dataloader/state_parser.py
+++ b/torchdata/stateful_dataloader/state_parser.py
@@ -1,0 +1,74 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
+# All rights reserved.
+#
+# This source code is licensed under the BSD-style license found in the
+# LICENSE file in the root directory of this source tree.
+import logging
+from typing import Any, Dict, Union
+
+logger = logging.getLogger(__name__)
+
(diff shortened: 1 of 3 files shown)


torchdata/stateful_dataloader/state_parser.py
```diff
@@ -0,0 +1,74 @@
...
+#
+# This source code is licensed under the BSD-style license found in the
+# LICENSE file in the root directory of this source tree.
+import logging
+from typing import Any, Dict, Union
+
+logger = logging.getLogger(__name__)
+
+
+class StateParserUtil:
```
andrewkho, line 12: can we keep these as functions? I'm not sure of the benefit of having a class here

andrewkho (changes requested):
I think pure functions are easier to reason about and harder to screw up, let me know if there's other reasons you want to keep them in a class

PR #1279, opened 2024-06-21
Save state in dataset_iter_state when dataset is also an iterator

gokulavasan:
Context:

For iterable datasets, state can be defined both at dataset level and also in the iterator. In the state_dict that is vended, this dataset_state is stored in dataset_state and iterator state is stored in fetcher_state::dataset_iter_state.

For a dataset which also acts as an iterator, initially the identical state was saved in both places which was wasteful. This was fixed in https://github.com/pytorch/data/pull/1258 where the state was saved only once in fetcher_state::dataset_iter_state.

But this was swapped in https://github.com/pytorch/data/pull/1273.

This is being swapped back to its original state to make sure the state variables can be initialized even in the iter method (which is the only place where it can access variables such as [WorkerInfo](https://github.com/pytorch/pytorch/blob/9d1b65b569a304032776375c24aa6babbf576c30/torch/utils/data/_utils/worker.py#L66)). 

Fixes #{issue number}

### Changes

-
-

### DIFF
--- a/test/stateful_dataloader/test_state_dict.py
+++ b/test/stateful_dataloader/test_state_dict.py
@@ -1503,21 +1503,22 @@ class CountIterCalls(torch.utils.data.IterableDataset):
         return iter(list(range(self.length)))
 
     def state_dict(self):
-        # iter_calls in the state is purely to track number of items iter was invoked since dataset initialization
         return {"iter_calls": self.iter_calls}
 
     def load_state_dict(self, state_dict):
-        pass
+        self.iter_calls = state_dict["iter_calls"]
+        
 
 class CountIterCallsIter(torch.utils.data.IterableDataset):
     def __init__(self, length):
         self.length = length
-        self.iter_calls = 0
+        self.local_iter_calls = 0
+        self.prev_state_iter_calls = 0
         self.items = []
 
     def __iter__(self):
         self.items = list(range(self.length))
-        self.iter_calls += 1
+        self.local_iter_calls += 1
         return self
 
     def __next__(self):
@@ -1528,9 +1529,10 @@ class CountIterCallsIter(torch.utils.data.IterableDataset):
 
     def state_dict(self):
         # iter_calls in the state is purely to track number of items iter was invoked since dataset initialization
-        return {"iter_calls": self.iter_calls, "items": deepcopy(self.items)}
+        return {"iter_calls": self.local_iter_calls + self.prev_state_iter_calls, "items": deepcopy(self.items)}
 
     def load_state_dict(self, state_dict):
+        self.prev_state_iter_calls = state_dict["iter_calls"]
         self.items = state_dict["items"]
 
 
@@ -1580,13 +1582,13 @@ class TestSingleIterCalled_shard0(TestCase):
         it = iter(dl2)
         state2 = dl2.state_dict()
         # Ensure that iter is called only once per worker even when dataloader resumes from a state
-        self.assertEqual(self._get_iter_calls(state2), [1] * max(1, num_workers))
+        self.assertEqual(self._get_iter_calls(state2), [3] * max(1, num_workers))
 
         for _ in range(10):
             next(it)
         state = dl2.state_dict()
         # Ensure that iter has not been invoked again
-        self.assertEqual(self._get_iter_calls(state2), [1] * max(1, num_workers))
+        self.assertEqual(self._get_iter_calls(state2), [3] * max(1, num_workers))
 
 
     def test_inline(self):
--- a/test/stateful_dataloader/test_state_dict.py
+++ b/test/stateful_dataloader/test_state_dict.py
@@ -1503,11 +1503,11 @@ class CountIterCalls(torch.utils.data.IterableDataset):
         return iter(list(range(self.length)))
 
     def state_dict(self):
+        # iter_calls in the state is purely to track number of items iter was invoked since dataset initialization
         return {"iter_calls": self.iter_calls}
 
     def load_state_dict(self, state_dict):
-        self.iter_calls = state_dict["iter_calls"]
-
+        pass
 
 class CountIterCallsIter(torch.utils.data.IterableDataset):
     def __init__(self, length):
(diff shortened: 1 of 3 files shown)


test/stateful_dataloader/test_state_dict.py
```diff
@@ -1529,7 +1530,8 @@ def state_dict(self):
         return {"iter_calls": self.iter_calls, "items": deepcopy(self.items)}
 
     def load_state_dict(self, state_dict):
-        self.iter_calls = state_dict["iter_calls"]
+        # sequence of calls for this : iter is called first and then load_state_dict is called and thus we don't want state to override calls to iter that has happened before load_state_dict was called
+        self.iter_calls += state_dict["iter_calls"]
```
andrewkho, line 1534: This makes load_state_dict non-idempotent, so eg if load_state_dict is called multiple times it will result in incorrect behaviour

gokulavasan: @andrewkho Updated the unit test to check specifically the iter_calls alone and not restore it value which is unnecessary

test/stateful_dataloader/test_state_dict.py
```diff
@@ -1511,25 +1512,26 @@ def load_state_dict(self, state_dict):
...
             raise StopIteration
 
     def state_dict(self):
-        return {"iter_calls": self.iter_calls, "items": deepcopy(self.items)}
+        return {"iter_calls": self.local_iter_calls + self.prev_state_iter_calls, "items": deepcopy(self.items)}
 
     def load_state_dict(self, state_dict):
-        self.iter_calls = state_dict["iter_calls"]
+        self.prev_state_iter_calls = state_dict["iter_calls"]
         self.items = state_dict["items"]
```
andrewkho, line 1535: Let's revert this, except delete the `self.iter_calls` in load_state_dict and just use it to track calls in the test

test/stateful_dataloader/test_state_dict.py
```diff
@@ -1540,26 +1542,52 @@ def _get_iter_calls(self, state):
...
             w_states = list(state["_snapshot"]["_worker_snapshots"].values())
 
-        return [x["dataset_state"]["iter_calls"] for x in w_states]
+        if w_states[0]["dataset_state"] is not None:
+            return [x["dataset_state"]["iter_calls"] for x in w_states]
+        return [x["fetcher_state"]["dataset_iter_state"]["iter_calls"] for x in w_states]
 
     def _run_test(self, num_workers, dataset):
+        # Need to make a copy here as iter calls is tracked and its count is stored in dataset state which persists across calls for single process runs.
+        dataset_copy = deepcopy(dataset) if num_workers == 0 else dataset
```
andrewkho, line 1551: Let's get the test working without this

andrewkho, line 1551:
```suggestion
        # Need to make a copy here as iter calls is tracked and its count is stored in dataset state which persists across calls for single process runs.
        dataset_copy = deepcopy(dataset)
```
Let's not have the test behave differently for different num_workers as much as possible, a test that is modified just to pass is not really a good test and I'd prefer we delete it

test/stateful_dataloader/test_state_dict.py
```diff
@@ -1540,26 +1542,52 @@ def _get_iter_calls(self, state):
...
         dl = StatefulDataLoader(
             dataset=dataset,
             num_workers=num_workers,
             multiprocessing_context=("forkserver" if IS_MACOS and num_workers else None),
+            persistent_workers=True if num_workers else False,
```
andrewkho, line 1556: The test should work for both PW = True and False

test/stateful_dataloader/test_state_dict.py
```diff
@@ -1540,26 +1542,52 @@ def _get_iter_calls(self, state):
...
-        return [x["dataset_state"]["iter_calls"] for x in w_states]
+        if w_states[0]["dataset_state"] is not None:
+            return [x["dataset_state"]["iter_calls"] for x in w_states]
+        return [x["fetcher_state"]["dataset_iter_state"]["iter_calls"] for x in w_states]
 
     def _run_test(self, num_workers, dataset):
+        # Need to make a copy here as iter calls is tracked and its count is stored in dataset state which persists across calls for single process runs.
+        dataset_copy = deepcopy(dataset) if num_workers == 0 else dataset
         dl = StatefulDataLoader(
             dataset=dataset,
```
andrewkho, line 1553: I think this should be dataset_copy too?

test/stateful_dataloader/test_state_dict.py
```diff
@@ -1540,26 +1542,52 @@ def _get_iter_calls(self, state):
...
         state2 = dl2.state_dict()
-        self.assertEqual(self._get_iter_calls(state2), [2] * max(1, num_workers))
+        # Ensure that iter is called only once per worker even when dataloader resumes from a state
+        self.assertEqual(self._get_iter_calls(state2), [3] * max(1, num_workers))
+
+        for _ in range(10):
+            next(it)
+        state = dl2.state_dict()
+        # Ensure that iter has not been invoked again
+        self.assertEqual(self._get_iter_calls(state2), [3] * max(1, num_workers))
```
andrewkho, line 1590: Let's change this test to just assert that __iter__ is only called once for new dataloaders with and without state-resumes, but not enforce the actual value of state["iter_calls"]

gokulavasan:
@andrewkho Addressed comments, please take a look. The only comment I couldn't resolve was around persistent workers being True. It has to be set to True if we want to match the behavior of single process and multiprocess test cases in the TestSingleIterCalled_shard0 unit test. This is because self.iter_calls is local dataset variable.

In case of single process, the modified value persists beyond the lifetime of the first dataloader created.

In case of multiprocess without persistent workers, the value is modified only the copy of that variable in the multiprocessing worker.

Let me know how you want to me to address it - either we can switch to persistent workers=True and keep the expected iter_calls value same for single process and multiprocess, or pass in different expected values for those two scenarios.

andrewkho (approved):
gogogo

PR #1282, opened 2024-07-03
Define and use _T_co locally instead of importing from torch.utils.data DataLoader

gokulavasan:
StatefulDataLoader import failed in torchtitan here - https://github.com/pytorch/torchtitan/actions/runs/9784768290/job/27016395174. 

The error was T_co import from dataloader failed. The failure happened because of this change - https://github.com/pytorch/pytorch/pull/129885 which renamed T_co to _T_co. 

Instead of import T_co from dataloader, define and use it locally within StatefulDataLoader

### DIFF
--- a/torchdata/stateful_dataloader/stateful_dataloader.py
+++ b/torchdata/stateful_dataloader/stateful_dataloader.py
@@ -25,7 +25,7 @@ import logging
 import queue
 import threading
 
-from typing import Any, Dict, Iterable, List, Optional, Union
+from typing import Any, Dict, Iterable, List, Optional, TypeVar, Union
 
 import torch
 import torch.multiprocessing as multiprocessing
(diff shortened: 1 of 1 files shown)


tianyu-l (approved):
Thanks for the quick fix!

andrewkho (approved):
Looks good, this works so let's land it, but one thought: this will generally happen because of our dependency on Core. This change fixes it for all builds which is great, but in general when release branch is cut, we should be testing it against the matching Core release branch. Hopefully we push quickly towards the 7/24 PyTorch Core release and we can start pointing people to pinned versions, and fixes like this won't be so urgent.

torchdata/stateful_dataloader/stateful_dataloader.py
```diff
@@ -183,7 +184,7 @@ class StatefulDataLoader(DataLoader[T_co]):
 
     def __init__(
         self,
-        dataset: Dataset[T_co],
+        dataset: Dataset[_T_co],
```
andrewkho, line 187: Will this have any issues for mypy or pyre for downstream users?

PR #1288, opened 2024-07-18
Bring back windows binary builds for release 0.8.0 (#1286)

atalman:


andrewkho:
I'm seeing Numpy warnings:
```
If you are a user of the module, the easiest solution will be to

downgrade to 'numpy<2' or try to upgrade the affected module.

We expect that some modules will need time to support NumPy 2.
```

and errors like this in windows:
`RuntimeError: Could not infer dtype of numpy.int64 `

Do we need to pin numpy < 2?

gokulavasan (approved):
lg!

Issue #1196, opened 2023-07-17
Future of torchdata and dataloading


laurencer:
### 📚 The doc issue

As of July 2023, we have paused active development on TorchData and have paused new releases. We have learnt a lot from building it and hearing from users, but also believe we need to re-evaluate the technical design and approach given how much the industry has changed since we began the project. During the rest of 2023 we will be re-evaluating our plans in this space.

We want to hear from users on their use-cases and the pain-points they have (with data loading in general or `torchdata` specifically). Please reply on this issue to help inform our future roadmap.

erip:
Thanks for the update, @laurencer. Does this mean `torchdata` as a domain library or the entire concept of datapipes/`DataLoader2` is under review?

laurencer:
The short answer is we need to look at both. More holistically there's lots of benefits with datapipes & DataloaderV2, however we've seen some limitations in a few use-cases which indicate we may need to tweak them a bit (or they're not the one-stop solution that we were hoping for). Overall the data loading space is really important and we hear about a lot of pain-points so we want to make sure we get the core abstractions right.

andrew-bydlon:
I too would like to hear what limitations you are referencing. 

If it is performance oriented, I believe there's an argument. You could make something compatible with a compiled framework, especially something like torch.compile. Someone recently talked about speedups loading tar files with rust as an example. 

My question is what you recommend as an alternative platform to torchdata for flexible and fast dataloading in pytorch?

nairbv:
Does the design re-evaluation apply only to TorchData, or also to the portion of the datapipes API that was upstreamed to PyTorch core?

https://github.com/pytorch/pytorch/tree/main/torch/utils/data/datapipes

BlueskyFR:
IMO the first thing that comes to mind is TensorFlow's `tf.data.Dataset` API, which is super cool to use. It is deeply integrated in the framework and with Keras, and the different operations you "pipe"/chain together can be fused at runtime so that the input pipeline is more optimized.
However there are still some solutions such as [NVIDIA DALI](https://github.com/NVIDIA/DALI) that is waaay faster if it applies to your use case.

I don't know TorchData in much details but I'd say building a nice-looking pipeline is easy, but making a pipeline optimized for high performance while still making it look cool is the real challenge.
By cool-looking I mean nice/easy to read as code, but also easily extensible, like users could share "databricks" together or something.

I hope this helps!

BlueskyFR:
[...]

PR #1289, opened 2024-07-20
add __len__ to get dp length

kaiyuyue:
```python
dp = ...
dp = dp.load_from_tar(length=ds_length)  # now __len__ will return self.length
dp = dp.webdataset()  # inherit self.length in the __len__

print(len(dp))  # able to get dp length
```


andrewkho:
Hi @kaiyuyue, please see https://github.com/pytorch/data/issues/1196 as we won't be supporting dataloader2 nor datapipes in torchdata moving forward

PR #1290, opened 2024-07-22
Fix linux wheels and conda on condition

atalman:
See following: https://github.com/pytorch/data/blob/main/.github/workflows/build-wheels-m1.yml


gokulavasan (approved):
thank you so much! looks good to merge once ci is green

PR #1292, opened 2024-07-23
Fix torch install instructions

atalman:
To conform with https://pytorch.org/get-started/locally/
```
pip3 install torch --index-url https://download.pytorch.org/whl/test/cpu
```


gokulavasan (approved):
lg!

PR #1294, opened 2024-07-23
Use oidc for wheel upload

atalman:
Required for uploading to aws s3.
Delete release.yml which is not used anymore


gokulavasan (approved):
lg!

PR #1297, opened 2024-07-25
Fix doc build

atalman:
As per following: https://github.com/JamesIves/github-pages-deploy-action/discussions/1431


gokulavasan (approved):
lg!

PR #1299, opened 2024-07-25
Add stateful dl import to smoke test

gokulavasan:
title

Fixes #{issue number}

### Changes

-
-


test/smoke_test/smoke_test.py
```diff
@@ -26,3 +30,6 @@ def s3_test():
     options = parser.parse_args()
     if options.s3:
         s3_test()
+    
+    if torchdata.__version__ >= "0.8.0":
```
atalman, line 34: Please throw exception here of torchdata.__version != 0.8.0 
We would need to route it later via input parameter. However this can be a followup PR

atalman (approved):
please address comment and we are good to merge

gokulavasan:
@atalman Is that change what you had in mind? Can you also confirm if it is fine to land this in main branch (with the check being about 0.8.0) and the cherry pick this to release/0.8 branch?

PR #1315, opened 2024-08-21
Make DistributedSampler stateful

### LINKED ISSUE #1269: Make DistributedSampler stateful
### 🚀 The feature

Currently RandomSampler, BatchSampler are patched here https://github.com/pytorch/data/blob/main/torchdata/stateful_dataloader/sampler.py#L134-L135 to make them stateful and work out of the box with StatefulDataLoader. 

It would be useful to consider making DistributedSampler (https://github.com/pytorch/pytorch/blob/2176ef7dfaf02dd6dbb8484a50c99d5fadf3ea0b/torch/utils/data/distributed.py#L13) also implement stateful methods and patch it in torchdata.

### Motivation, pitch

So that users can use DistributedSampler also out of the box with checkpointing capability

### Alternatives

Users would have implement the stateful interface for DistributedSampler but extending it

### Additional context

_No response_

ramanishsingh:
Fixes #1269 

### Changes

- `torchdata/stateful_dataloader/sampler.py` : Added new classes `StatefulDistributedSampler` and `_StatefulDistributedSamplerIterator`
- `test/stateful_dataloader/test_dataloader.py` new tests for `StatefulDistributedSampler`

### DIFF
--- a/test/stateful_dataloader/test_dataloader.py
+++ b/test/stateful_dataloader/test_dataloader.py
@@ -1074,6 +1074,15 @@ def row_processor(row):
 def filter_len(row):
     return len(row) == 4
 
+class MockDataset(Dataset):
+    def __init__(self, size):
+        self.size = size
+        self.data = torch.arange(size)  # Simple data that is easy to verify
+    def __len__(self):
+        return self.size
+    def __getitem__(self, idx):
+        return self.data[idx]
+
 
 @unittest.skipIf(
     TEST_WITH_TSAN,
@@ -1088,6 +1097,7 @@ class TestDataLoader(TestCase):
         self.labels = torch.randperm(50).repeat(2)
         self.dataset = TensorDataset(self.data, self.labels)
         self.persistent_workers = False
+        self.mockdataset = MockDataset(100)
 
     def _get_data_loader(self, dataset, **kwargs):
         persistent_workers = kwargs.get("persistent_workers", self.persistent_workers)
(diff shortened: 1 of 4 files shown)


andrewkho:
AI Store test can be safely ignored for now

test/stateful_dataloader/test_dataloader.py
```diff
@@ -1947,6 +1960,116 @@ def test_sampler_reproducibility(self):
...
+
+    def test_data_distribution_across_replicas(self):
+        from torchdata.stateful_dataloader.sampler import StatefulDistributedSampler
+
+        num_replicas = 5
+        all_data = []
+        for rank in range(num_replicas):
+            sampler = StatefulDistributedSampler(self.mockdataset, num_replicas=num_replicas, rank=rank, shuffle=False)
+            indices = list(iter(sampler))
+            data_sampled = [int(self.mockdataset[i].numpy().astype(int)) for i in indices]
```
andrewkho, line 2044: nit:
```suggestion
            data_sampled = [int(self.mockdataset[i].item()) for i in indices]
```

test/stateful_dataloader/test_dataloader.py
```diff
@@ -1947,6 +1960,116 @@ def test_sampler_reproducibility(self):
...
+    def test_data_order_without_shuffle(self):
+        from torchdata.stateful_dataloader.sampler import StatefulDistributedSampler
+
+        sampler = StatefulDistributedSampler(self.mockdataset, num_replicas=1, rank=0, shuffle=False)
+        indices = list(iter(sampler))
+        data_sampled = [self.mockdataset[i] for i in indices]
+        self.assertEqual(data_sampled, list(range(100)), "Data should be in sequential order when shuffle is False")
+
+    def test_data_distribution_across_replicas(self):
+        from torchdata.stateful_dataloader.sampler import StatefulDistributedSampler
```
andrewkho, line 2037: Let's move this import to top level for all cases

test/stateful_dataloader/test_dataloader.py
```diff
@@ -1947,6 +1960,116 @@ def test_sampler_reproducibility(self):
                     ls[i].append(next(its[i]))
             self.assertEqual(ls[0], ls[1])
 
+    def test_initialization_StatefulDistributedSampler(self):
```
andrewkho, line 1963: Let's move all of these tests out to a new file called `test_sampler.py`. You can update https://github.com/pytorch/data/blob/main/.github/workflows/stateful_dataloader_ci.yml to call it in an additional step

test/stateful_dataloader/test_dataloader.py
```diff
@@ -1947,6 +1960,116 @@ def test_sampler_reproducibility(self):
...
+        from torchdata.stateful_dataloader.sampler import StatefulDistributedSampler
+
+        dataset = self.dataset
+        sampler = StatefulDistributedSampler(dataset, num_replicas=10, rank=0, shuffle=False, seed=42, drop_last=False)
```
andrewkho, line 1967: For testing state_dict, let's have most of the tests set up with passing sampler + dataset to StatefulDataLoader so we can test that it works end-to-end

andrewkho: You might need to use a dummy Collate function to easily inspect elements, check the `test_state_dict.py` file for examples

torchdata/stateful_dataloader/sampler.py
```diff
@@ -125,3 +127,86 @@ def __iter__(self):
...
+        seed: int = 0,
+        drop_last: bool = False,
+    ) -> None:
+
+        super().__init__(dataset, num_replicas, rank, shuffle, seed, drop_last)
+        self.yielded = 0
+        self.next_yielded = None
+
+    def __iter__(self):
+
```
andrewkho, line 202: Is it possible to fork the `DistributedSampler.__iter__` code here instead and just update, instead of having a separate Iterator class?

torchdata/stateful_dataloader/sampler.py
```diff
@@ -125,3 +127,86 @@ def __iter__(self):
...
+            indices = indices[: self.sampler.total_size]
+        assert len(indices) == self.sampler.total_size
+
+        # subsample
+        indices = indices[self.sampler.rank : self.sampler.total_size : self.sampler.num_replicas]
+        assert len(indices) == self.sampler.num_samples
+
+        self.parent_iterator = iter(indices)
+        self.indices = list(self.parent_iterator)
+        self.current_index = 0
```
andrewkho, line 162: Is there a way to call the original code instead of forking it here?

torchdata/stateful_dataloader/sampler.py
```diff
@@ -125,3 +127,86 @@ def __iter__(self):
...
+        val = self.indices[self.current_index]
+        self.current_index += 1
+        self.sampler.yielded += 1
+        return val
+
+    def state_dict(self) -> Dict[str, Any]:
+        return self.sampler.state_dict()
+
+    def load_state_dict(self, state_dict: Dict[str, Any]) -> None:
+        self.sampler.load_state_dict(state_dict)
```
andrewkho, line 181: I don't think we need this both here and in the main sampler class, can we consolidate to have this in just one place?

andrewkho (changes requested):
Looks pretty good, but would like to simplify the code a bit and move the tests around as well

test/stateful_dataloader/test_dataloader.py
ramanishsingh: Updated.

test/stateful_dataloader/test_dataloader.py
ramanishsingh: Done.

test/stateful_dataloader/test_dataloader.py
ramanishsingh: Created here: https://github.com/pytorch/data/blob/stateful_distributedsampler/test/stateful_dataloader/test_sampler.py

Added new line here: https://github.com/pytorch/data/blob/cdc5d318bab9432383e808dd25489677dd36edf9/.github/workflows/stateful_dataloader_ci.yml#L81

test/stateful_dataloader/test_dataloader.py
ramanishsingh: New tests here: https://github.com/pytorch/data/blob/cdc5d318bab9432383e808dd25489677dd36edf9/test/stateful_dataloader/test_sampler.py#L173

torchdata/stateful_dataloader/sampler.py
ramanishsingh: https://github.com/pytorch/data/blob/cdc5d318bab9432383e808dd25489677dd36edf9/torchdata/stateful_dataloader/sampler.py#L149

torchdata/stateful_dataloader/sampler.py
```diff
@@ -125,3 +126,46 @@ def __iter__(self):
...
+        if self.next_yielded is not None:
+            self.current_index = self.next_yielded
+            self.yielded = self.next_yielded
+            self.next_yielded = None
+        if self.current_index >= len(self.indices):
+            raise StopIteration
+        val = self.indices[self.current_index]
+        self.current_index += 1
+        self.yielded += 1
+        return val
```
andrewkho, line 163: This will make a copy of the whole index list. Since DistributedSampler has a copy of the whole list anyways, the most efficient would be to randomly index into that, but we'd need to expose it or copy/paste that code into here. Instead let's just use itertools slice to avoid making the second copy of the list

```suggestion
    def __iter__(self):
        self.yielded = 0
        if self.next_yielded is not None:
            self.yielded = self.next_yielded
            self.next_yielded = None
        it = super().__iter__()
        for idx in itertools.islice(it, self.yielded, None):
            self.yielded += 1
            yield idx
```

andrewkho (commented):
Couple of suggestions, but looks great! very nice test suite.

When you're done making changes, please run the fbcode CI for media_dataloader

andrewkho (approved):
LGTM!

PR #1319, opened 2024-09-25
[Release] update CI python versions

andrewkho:
PyTorch 2.5 supports 3.9<=python<=3.12 (3.13 experimental). Updating our CI to match.

To save our CI costs, we're not including tests for python > 3.9 for any datapipes code, only for stateful dataloader 

-
-


ramanishsingh:
LGTM!
Just one quick question, do we also need to update 
https://github.com/pytorch/data/blob/cf740b890d8b5f10bd68d9a5ff2ad2218325e872/.github/workflows/_build_test_upload.yml#L227

divyanshk (approved):
Looks good to me!

ramanishsingh:
Do we need to update this line?
https://github.com/pytorch/data/blob/8b6e903e7899df8a87406642926eb0c5b3b5a807/test/test_distributed.py#L249

andrewkho:
> Do we need to update this line?
> 
> https://github.com/pytorch/data/blob/8b6e903e7899df8a87406642926eb0c5b3b5a807/test/test_distributed.py#L249
> 
> Otherwise, looks good :)

I don't think this needs to change, i'm not sure it's true that they only support 3.9+

PR #1320, opened 2024-09-25
Fix broken windows build

andrewkho:
Windows build has been broken for a while on trunk, due to new MSVC deprecation warnings being treated by errors. It turns out the submodule pointer for aws-sdk-cpp hadn't been updated in a while. This PR updates it the pointer to 1.11.311, which can build on windows. Main branch does not build on MacOS due to a bug in cmake where it doesn't set the correct MACOSX_DEPLOYMENT_TARGET.

-
-


third_party/CMakeLists.txt
```diff
@@ -54,6 +54,9 @@ else()
     else()
       list(APPEND AWSSDK_LIBRARIES ${aws_cpp_sdk_INSTALL}/lib/lib${lib}.a)
     endif()
+    if(MSVC)
+      target_compile_options(${lib} PRIVATE /W4 /WX)
```
divyanshk, line 58: Quick clarification, do we mean to make warnings surface better ? or suppress them ?

andrewkho: Ah sorry there's a compiler flag that makes warnings get treated as errors. In this case it's a deprecation warning and just trying to get it to ignore, not sure the right incantation

ramanishsingh:
LGTM!

divyanshk (approved):
LGTM!

PR #1328, opened 2024-09-26
Andrewkh/fix nightly 2

andrewkho:
This time I tested it by manually running a workflow. This branch is prohibted from actually pushing so it fails during push step, but we're able to run and see builds succeed: https://github.com/pytorch/data/actions/runs/11057077541 

-
-


divyanshk (approved):
Thanks for taking care of this, LGTM!

PR #1329, opened 2024-09-27
Cleanup DataLoader 2

andrewkho:
Starting off the cleanup by deleting DataLoader2, CI files, and a bunch of tests 

-
-


torchdata/datapipes/iter/util/prefetcher.py
divyanshk, line 1: Do other references to torchdata.datapipes in this file need to stay?

andrewkho: I think we'll just delete the whole thing in one go, i wanted to try and split it up into 2-ish PRs so we can at least have some partial rollback if necessary

divyanshk: Cool, in that case maybe we want to update the PR title?

Accepting, since tests seem to pass (except some Meta specific bots).

divyanshk (approved):
LGTM!

ramanishsingh (approved):

Issue #1334, opened 2024-10-10
[RFC] Polylithic: Enabling multi-threaded DataLoading through non-monolithic parallelism


andrewkho:
This RFC was re-created due to a problem with the original. Summary of comments from previous issue below.

### 🚀 The feature

**TL;DR \-**  We want to lean into **modular Multi-Threading/Multi-Processing** instead of the current monolithic Multi-Processing, and steer users away from the monolithic Dataset parallelism approach towards composable DataSources, and composable IterableDatasets for pre-proc operations, with parallelism configured within each operation. This will enable **multi-threaded** dataloading (with **NoGIL** support), **auto-tunable** parallelism, **torch.compilable** and GPU enabled preproc operations, more efficient loading of **mixed-modalities**, and **composable** dataloading and pre-proc graphs.

### Motivation, pitch

Working name for the project: **Polylithic** (non-monolithic)  
Where it will live: torchdata

## Multimodal DataLoading is here and torch.utils.data doesn’t support it well

Multi-Modal LLMs are here. Tasks like fine-tuning, alignment, and distillation will require multi-modal dataloading for our users. LLM training often requires reading from 10s-100s of multi-modal datasets, tokenizing them, and packing them into a “token-buffer” where tokens from individual datasets are shuffled and combined into training examples for the model. 

Audio, Image, and Video datasets may also require heavy-weight decoding operations to be performed before tokenization, and the difference in the data sizes between text, image, and video may be orders of magnitude. GPU decoding of images and video is an option for users as well, and libraries like Nvidia DALI will compile the entire pre-proc pipeline into GPU operations, minimizing the overhead of transfers between CPU and GPU memory.

torch.utils.data’s Dataset and DataLoader abstractions are extremely popular with users, however they are not well equipped to handle MultiModal DataLoading and accelerated pre-proc, because of the monolithic, black-box way in which it treated parallelism with multiprocessing; ie running GPU Preproc under multiprocessing is [not currently realistic](https://github.com/pytorch/pytorch/issues/20532). While the abstractions are extremely flexible and very easy to experiment with, users are often required to write bespoke classes to create pre-proc pipelines, handle data sharding and combine multiple datasets. Optimizing is also a challenge because of the lack of control in parallelism.

### Existing Context and definitions

Torch.utils.data contains the following abstractions today:

[...]

PR #1335, opened 2024-10-14
Andrewkh/polylithic initial commit

andrewkho:
See #1334 

Initial commit of
* BaseNode
* Mapper
* ParallelMapper (both threads and processes available, we might want to make two classes)
* Prefetcher
* Batcher
* PinMemory


torchdata/nodes/map.py
```diff
@@ -0,0 +1,132 @@
...
+        self._shutdown()
+
+    def _shutdown(self):
+        if self._started:
+            self._stop.set()
+            self._mp_stop.set()
+            for _ in range(5):
+                self._read_thread.join(timeout=0.1)
+                if not self._read_thread.is_alive():
+                    break
```
scotts, line 125: If we weren't able to successfully join with the thread after 5 tries, should we log something? I get that we don't want to block, and raising an exception would also be rude.

torchdata/nodes/pin_memory.py
```diff
@@ -0,0 +1,163 @@
...
+            self.pin_memory_thread = threading.Thread(
+                target=_pin_memory_loop,
+                args=(
+                    self.in_q,
+                    self.out_q,
+                    self.current_device,
+                    self._pin_memory_stop_event,
+                    self._pin_memory_device,
+                ),
+            )
```
scotts, line 124: Would it be possible to eliminate the `read_thread` and do all reading and pinning in the `pin_memory_thread`? My thinking here is that every thread we introduce is going to introduce potential copying overhead.

My understanding of the purpose of `read_thread` is just to get items from `self.source` into `self.in_q`. In `pin_memory_thread`, we then read items from `self.in_q`, pin them, and put them in `self.out_q`. I think we could simplify things by eliminating `read_thread` and `self.in_q`. Then data would go directly from `self.source` to `self.out_q` inside `pin_memory_thread`.

Another way to think about this: we want to introduce threads for operations that are likely to be expensive. If _both_ reading and pinning are likely to be expensive, then it makes sense to use a thread for both. But my intuition is that will be faster to combine them into one thread.

torchdata/nodes/map.py
```diff
@@ -0,0 +1,132 @@
...
+                    self.out_q,
+                    self.in_order,
+                    self.udf,
+                    self._stop if self._method == "thread" else self._mp_stop,
+                )
+                self._map_threads.append(
+                    threading.Thread(target=_apply_udf, args=args)
+                    if self._method == "thread"
+                    else mp.Process(target=_apply_udf, args=args)
+                )
```
scotts, line 78: For all of the other thread object creation, we do it in the constructor. Is there a particular reason that we defer the thread object creation for the map threads until the time we start all the threads?

torchdata/nodes/pin_memory.py
scotts: Looking at `ParallelMapper`, I think my comment above applies there as well.

torchdata/nodes/map.py
```diff
@@ -0,0 +1,132 @@
...
+        self._map_threads: List[Union[threading.Thread, mp.Process]] = []
+
+    def iterator(self) -> Iterator[T]:
+        if not self._started:
+            for worker_id in range(self.num_workers):
+                args = (
+                    worker_id,
+                    self.in_q,
+                    self.out_q,
+                    self.in_order,
```
scotts, line 70: Do I understand this correctly that all of the worker threads are getting the same `in_q` and `out_q`? That means that the parallel workers will compete with each other for work to do, and then compete with each other to submit processed work. (I don't mean this as a criticism, I'm just making sure I understand the logic correctly.)

torchdata/nodes/prefetch.py
```diff
@@ -0,0 +1,56 @@
...
+            self.thread = threading.Thread(
+                target=_populate_queue,
+                args=(self.source, self.q, self._stop_event, self.sem),
+            )
+            self.thread.start()
+
+        exception: Optional[ExceptionWrapper] = None
+        while True:
+            try:
+                item = self.q.get(block=True, timeout=0.1)
```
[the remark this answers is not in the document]
andrewkho: Agreed it's probably not required for these queue applications as they're meant to allow the task to come back and check for event flags. I'm thinking just a global var somewhere, eg https://github.com/pytorch/pytorch/blob/main/torch/utils/data/_utils/__init__.py#L20 although torch.utils.data.DataLoader does allow this to be set by the user, not sure it's actually used

torchdata/nodes/pin_memory.py
andrewkho: I think you're right on this one for PinMemory, since it's single-threaded. For ParallelMapper I think we want some way to guarantee single-threaded access to source, in case source is either not-threadsafe, or not properly sharded in the case of multiprocess workers. LMK if that doesn't make sense, can discuss more tomorrow

torchdata/nodes/map.py
andrewkho: No good reason other than it was the way the code evolved during experimentation. Let's make sure there's uniformity before landing, and also in variable naming conventions

torchdata/nodes/map.py
andrewkho: Yes that's how it's set up right now. I assume you mean in the sense of contention on the queue locks? When I last tested throughput for Imagenet, we were nowhere near hitting this limit (eg orders of magnitude lower QPS) but definitely open to changing this

scotts: Yes, the "competition" will come through potential contention on the queue locks, but that will only make a difference if the work they're competing for is very, very small. I believe this is the correct pattern for when you have a relatively coarse granularity of work: a global queue, and workers take work when they're free to do so.

torchdata/nodes/map.py
andrewkho: Just a follow-up: we'll consolidate as part of iterator creation, this is how it's handled in current torch.utils.data.DataLoader and part of this is so that IterableDataset's have a known time (iterator creation" when they will be instantiated inside a process, so they can shard themselves. We'll keep this to make sure we don't run into BC issues later

torchdata/nodes/_populate_queue.py
```diff
@@ -0,0 +1,40 @@
...
+import queue
+import threading
+from typing import Iterable
+
+from torchdata.nodes.exception_wrapper import ExceptionWrapper, StartupExceptionWrapper
+
+
+def _populate_queue(
+    source: Iterable,
+    q: queue.Queue,
```
divyanshk, line 11: This should be `Union[queue.Queue, mp.Queue]` ?

torchdata/nodes/map.py
```diff
@@ -0,0 +1,195 @@
...
+        method: Literal["thread", "process"] = "thread",
+    ):
+        self.source = source
+        self.udf = map_fn
+        self.num_workers = num_workers
+        self.in_order = in_order
+
+        self.in_q: Union[queue.Queue, mp.Queue] = queue.Queue() if method == "thread" else mp.Queue()
+        self.out_q: Union[queue.Queue, mp.Queue] = queue.Queue() if method == "thread" else mp.Queue()
+        self.sem = threading.BoundedSemaphore(value=2 * num_workers)
```
divyanshk, line 50: Confirming my understanding, here `2 * num_workers` is essentially deciding the input queue size? Is `2` here a hyper-parameter to be tuned?

torchdata/nodes/_populate_queue.py
```diff
@@ -0,0 +1,40 @@
...
+    """
+    try:
+        src_iter = iter(source)
+    except Exception:
+        e = StartupExceptionWrapper(where="in _populate_queue startup for device")
+        q.put(e)
+        return
+
+    while not stop_event.is_set():
+        if not semaphore.acquire(blocking=True, timeout=1.0):
```
scotts, line 28: I'm not sure if we're gaining anything using a semaphore to control the size of the queue, rather than just using a queue with a fixed size: https://docs.python.org/3/library/queue.html#queue.Queue

It would make sense to me if we were trying to acquire the semaphore in a non-blocking manner, and then went off and did something useful when we couldn't acquire the semaphore. But we are blocking, so I think the behavior will be the same as if we had a blocking queue?

torchdata/nodes/_populate_queue.py
andrewkho: good catch!

torchdata/nodes/map.py
andrewkho: Yes that's right, the entire value can be tuned, but we don't expose this right now

torchdata/nodes/_apply_udf.py
```diff
@@ -0,0 +1,28 @@
+import queue
+
+import torch
+
+from torch._utils import ExceptionWrapper
+
+
+def _apply_udf(worker_id, in_q, out_q, in_order, udf, stop_event):
+    torch.set_num_threads(1)
```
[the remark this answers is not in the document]
andrewkho: By default, any torch transforms can use a ton of threads, creating contention when we have a lot of workers. This setting seems to work for both threads and processes, and in testing lead to significant overall throughput improvements for imagenet. This is the right place for current architecture as the thread/process gets spun up and this should only get called once until thread/process is killed

torchdata/nodes/_populate_queue.py
andrewkho: This isn't obvious, basically the reason I went this way instead of using the sized-queue to block is that we're then holding some state in the stack implicitly, and if a user wanted to do some kind of checkpointing or state saving, this would be hard to expose to them otherwise. TBH this might not be necessary, it depends on how we're going to handle checkpointing though.

torchdata/nodes/_populate_queue.py
```diff
@@ -0,0 +1,40 @@
...
+    source: Iterable,
+    q: queue.Queue,
+    semaphore: threading.BoundedSemaphore,
+    stop_event: threading.Event,
+):
+    """Note that this is only intended to be used
+    by a single thread at once. Each instance creates its own iter for source so
+    if this is called with multiple threads, you may get duplicates if
+    source is not sharded properly.
+    """
```
divyanshk, line 19: Basic question. In the canonical `DataLoader` we have one index queue per worker process, and one output queue per worker. In the new setup, would we have one input queue and one output queue per instance of `ParallelMapper` (correct me if I confused myself here). How many `ParallelMapper`s would we have in a typical setup ? I'd assume >1 because otherwise there is only one thread doing dataset access.

andrewkho: for the torch.utils.data.DataLoader, there is one output queue shared by all workers, but otherwise you're right, there's one index queue per worker. The reason they do it that way is to give guarantees on determinism, so that work is always assigned round-robin style to the workers.

Currently we have just one index queue but all workers can pull from it, so parallelism isn't constrained, but we might want to adopt something similar to torch.utils.data.DataLoader to guarantee things like random transforms and state. We can do that in a follow up task

torchdata/nodes/_populate_queue.py
scotts: Hmm. If we're going to checkpoint the current size of the queue, then I assume we would also want to checkpoint the contents of the queue - which would also imply its size.

In general, though, I don't think we'll want to checkpoint the internal state of the runtime system. Rather, I think we'll want to checkpoint the actual state built up as a result of processing data.

To (maybe) make this a bit more clear: let's say we're reading from a file, and every operator we're processing is stateless - it's just applying a function to a data item, producing another data item. Only the input data item effects the output data item. In such a system, our checkpoints would only need to be at the source, where we're reading the file, and at the sink, to know what we've produced on the output. Everything in between can be lost; we build it back up when the source recovers, and starts reading from the correct place in the file.

We can talk more on this later - I don't think it's critical now.

scotts:
This might not be the final architecture we want with respect to where threads are. That is, in the current architecture, threads live inside of operators (nodes) directly. When threads live directly inside of operators, controlling the level of parallelism is difficult, as the number of threads is a decision local to each operator. We probably want an architecture where operators are agnostic to threads, and we have a runtime with a thread pool that can execute those operators. That allows us to more easily control the level of parallelism.

But, making changes like that should be possible with the interface we have for the nodes. Regardless, I think we want to move forward here so that we can start using the new framework with real use-cases.

scotts (approved):

divyanshk (approved):
Thanks for this amazing work! When looking at working examples, I'm trying to map it to a typical setup of torch.utils.dataloader and conform to its mental model. I guess a lot of end users would also be doing that. One thing we might have to think about is how to effectively "educate" the end users - since most might not be aware of all the smaller components inside a typical dataloader. Overall this is really promising!

PR #1346, opened 2024-10-21
Fix windows wheel build

andrewkho:
Current windows wheel build is failing on Nightly because it's trying to build S3 binaries, but we don't have any native extensions anymore. This leaves most of the code intact but just disables the envvar.

Example failing nightly:
https://github.com/pytorch/data/actions/runs/11438773492 

image

Example passing job build step (upload fails because of branch protection, but that's fine)

https://github.com/pytorch/data/actions/runs/11444540802/job/31839644282

image

### DIFF
--- a/.github/workflows/nightly_release.yml
+++ b/.github/workflows/nightly_release.yml
@@ -12,7 +12,7 @@ permissions:
 jobs:
   build_test_upload:
     if: |
-      github.repository == 'pytorch/data' && github.ref_name == 'main'
+      (github.repository == 'pytorch/data' && github.ref_name == 'main') || github.event_name == 'workflow_dispatch'
     uses: ./.github/workflows/_build_test_upload.yml
     with:
       branch: "main"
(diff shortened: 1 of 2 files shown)


.github/workflows/nightly_release.yml
```diff
@@ -12,7 +12,7 @@ permissions:
 jobs:
   build_test_upload:
     if: |
-      github.repository == 'pytorch/data' && github.ref_name == 'main'
+      (github.repository == 'pytorch/data' && github.ref_name == 'main') || github.event_name == 'workflow_dispatch'
```
andrewkho, line 15: Adding this flag to allow manual workflow_dispatches to run on feature branches, to make testing easier. Going to leave this in as it seems useful in general

ramanishsingh (approved):

andrewkho:
Merging, stateful dataloader tests don't depend on this at all

PR #1349, opened 2024-10-26
Don't subclass IterableDataset, and add two-directional adapters instead

andrewkho:
Drop the subclassing of IterableDataset, let's not pull these old weird decisions in, eg https://github.com/pytorch/pytorch/issues/120139 

Instead we add simple adapter class to cast if someone wants to stick this into a dataloader for some reason.

Also add some convenience adapters to go the other way, which take Map/Sampler or Iterable style datasets and implement the `iterator` interface for BaseNode

TestPlan: `pytest --no-header -v test/nodes/ --full-trace` passes

### DIFF
--- a/test/nodes/test_adapters.py
+++ b/test/nodes/test_adapters.py
@@ -1,0 +1,90 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
+# All rights reserved.
+#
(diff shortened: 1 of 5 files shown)


divyanshk (commented):
Nice change! LGTM!

divyanshk (approved):

PR #1353, opened 2024-11-05
Add state management to torchdata.nodes

andrewkho:
Adds state_dict/load_state_dict interface to torchdata.nodes.base_node. Update all existing node definitions to save/load state correctly.

# Interface for users (external facing)
```
BaseNode.state_dict() -> Dict[str, Any]: ...
""" Can be called at any time, on the BaseNode, and return a dictionary representing state_dict that can be
passed to load_state_dict. Calling this will return state_dict of the last result of BaseNode.__iter__()"""

BaseNode.load_state_dict(state_dict: Dict[str, Any]): ...
""" When called with the output of BaseNode.state_dict(), affects the next iterator returned by BaseNode.__iter__()
""" 
```

# Interface for developers (implementing Nodes)
```
BaseNode.iterator(initial_state: Optional[Dict[str, Any]]) -> Iterator[T]: ...
""" Resuming state is only possible at iterator instantiation. Developers must explicitly define how to handle non-null initial_state, and ensure that their generator or iterator is set up to resume correctly. input will be an output of BaseNode.get_state().

As usual, users should not call `.iterator()` directly, but instaed call `.__iter__()` instead (through `iter(...)`).
"""

BaseNode.get_state() -> Dict[str, Any]: ...
""" Developers of Nodes are required to implement and explicitly handle state management for their class.
Notably, if the node depends on a parent base node, you might do something like:
`return {"parent": self.parent.state_dict()}`

Note that users implement `get_state()` but should call `parent.state_dict()`
""" 
```

## Notable design choices: 
* the Iterator returned from BaseNode does not have state_dict/load_state_dict methods. 
  * Rationale: this greatly simplifies assumptions on where state may live, and is the lowest common denominator for defining both explicit Iterator classes, or defining generators. 
  * If we call state_dict in iterators, we can either a) force everyone to have explicit iterators, or b) check both base-node and iterator for state, which can get hairy (see eg stateful_dataloader). Consider having to reason about correctly configuring state for this scenario: `base_node.load_state_dict(sd["base"]); it = iter(base_node); it.load_state_dict(["base_iter"])`
* As a result of this, we're making the decision to assume that each BaseNode may have one active iterator at any time, which again unifies state-management for both generators and explicitly defined Iterators.
* state can only be resumed at iterator instantiation `.iterator(initial_state)`.
* We follow the same convention as StatefulDataLoader:
  * state_dict will return state of the last iterator returned by __iter__.
  * load_state_dict() will set the initial_state for the iterator returned by the next __iter__ call. 

## Testing
We'll provide a re-usable test function (or suite of tests) for developers to call in their own tests. To develop new nodes, devs will need to:
1) subclass BaseNode[T]
2) implement `.iterator(initial_state)`
3) implement `.get_state()`
4) write unit-tests that call `tests.nodes.utils.run_test_save_load_state`, in addition to testing the basic functionality.

### DIFF
--- a/torchdata/nodes/base_node.py
+++ b/torchdata/nodes/base_node.py
@@ -46,13 +46,17 @@ class BaseNode(Iterable[T]):
         raise NotImplementedError()
 
     def __iter__(self) -> BaseNodeIterator[T]:
-        if self.__it is not None and self.__it.started():
-            # Only create a new iter if the last requested one did not start/finish
-            self.__it = None
+        if self.__it is not None and not self.__it.started():
+            # Only create a new iter if the last requested one did not start
+            return self.__it
 
-        if self.__it is None:
+        if self.__initial_state is not None:
             self.__it = _EagerIter(self, self.__initial_state)
             self.__initial_state = None
+            if not self.__it.has_next():
+                self.__it = _EagerIter(self, self.__initial_state)
+        else:
+            self.__it = _EagerIter(self, self.__initial_state)
         return self.__it
 
     def state_dict(self) -> Dict[str, Any]:
@@ -76,13 +80,27 @@ class _EagerIter(BaseNodeIterator[T]):
         else:
             self._it = self.base_node.iterator(None)
 
+        self._next_val: Optional[T] = None
+
     def __next__(self) -> T:
         self._started = True
+        if self._next_val is not None:
+            val = self._next_val
+            self._next_val = None
+            return val
         try:
             return next(self._it)
         except StopIteration:
             self._finished = True
             raise
+
+    def has_next(self) -> bool:
+        if self._next_val is None:
+            try:
+                self._next_val = next(self._it)
+            except StopIteration:
+                pass
+        return self._next_val is not None
 
     def started(self) -> bool:
         return self._started
--- a/torchdata/nodes/base_node.py
+++ b/torchdata/nodes/base_node.py
@@ -45,7 +45,7 @@ class BaseNode(Iterable[T]):
         """
         raise NotImplementedError()
 
-    def __iter__(self) -> Iterator[T]:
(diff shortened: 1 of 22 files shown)


divyanshk:
Is there an example or unit test with the recursive call of `get_state()` into parent nodes?

andrewkho:
@divyanshk check out `test_prefetch` and `test_map`, these should have the recursive calls in there

torchdata/nodes/base_node.py
```diff
@@ -4,33 +4,88 @@
...
-        return _EagerIter(self)
+    def get_state(self) -> Dict[str, Any]:
+        """Return a dictionary that can be passed to iterator(...) which
+        can be used to initialize iterator at a certain state.
+        """
+        raise NotImplementedError()
+
+    def __iter__(self) -> BaseNodeIterator[T]:
+        if self.__it is not None and self.__it.started():
+            # Only create a new iter if the last requested one did not start/finish
```
divyanshk, line 50: Confirming, we return a new iterator when the previous one finished?

andrewkho: yes, or even if it just started, eg imagine this scenario:
```
ds = IterableWrapper(range(10))
it = iter(ds)
for _ in range(5):
  next(it)

list(ds) # Should result in [0, ..., 10]
```

divyanshk: Thanks. Thats basically saying that unless you save the state, or load a new 'initial state' we start over everytime you request a new iter (which is the desired behavior).

andrewkho:
Can't repro this on my mac w/Python 3.9/3.10. Somebody teach me how to get the container and ssh in so I can run it locally and get a stack-trace

divyanshk (approved):
Looks good!

PR #1357, opened 2024-11-07
Add SamplerWrapper node

andrewkho:
### Changes
* Adds a new SamplerWrapper node which will track epochs and call `set_epoch` on sampler.
* Default epoch increment behaviour can be changed by callback 

-
-


divyanshk (approved):
LGTM!

PR #1358, opened 2024-11-11
Changes default behaviour of StopIteration immediately after load_state_dict to be false

andrewkho:
Adds docstring about saving/loading state_dict at end-of-iteration (either after or just before StopIteration is thrown). 

See Docstring for details on the rationale behind this change.

TL;DR - If you load a state_dict from end-of-epoch, calling next(iter(node)) on it would throw stop iteration. This is unexpected for the (very) common case of saving state_dict at end of epoch, and loading it and expecting it to restart from the beginning of the next epoch, so the old default behaviour is to catch that stop iteration (through has_next()), and restart the iterator automatically.

The problem is that this breaks down when BaseNodes actually _need to know_ that StopIteration is going to be thrown, for example, multi-dataset samplers that need to handle end-of-iteration from one of their datasets.

We introduce a flag `restart_on_stop_iteration` in the `load_state_dict` as it's the most obvious place to set this, without making `.iterator()` implementations aware of this (they don't need to be). For this to work properly, the default MUST be False, because otherwise all child and grand-child nodes in a dag would catch it and prevent the root node from being able to handle it itself.

### DIFF
--- a/test/nodes/utils.py
+++ b/test/nodes/utils.py
@@ -4,6 +4,7 @@
 # This source code is licensed under the BSD-style license found in the
 # LICENSE file in the root directory of this source tree.
 
+import copy
 import random
 import time
 from typing import Any, Dict, Iterator, Optional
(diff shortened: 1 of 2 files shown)


divyanshk (approved):
LGTM, thank you for the quick PR

ramanishsingh (approved):

PR #1359, opened 2024-11-11
add global wrapper node, need to think of better name

andrewkho:
Although the default behaviour restart_on_stop_iteration needs to be False, there is still a very common use-case where users expect it to be true, namely: saving/restarting at the end of an epoch. 

This is our opportunity/good excuse to add a global "wrapper" node to change this behaviour, and it can be our entry point to adding other global/dag traversing behaviour as well.


torchdata/nodes/dataloader.py
```diff
@@ -0,0 +1,22 @@
+from typing import Any, Dict, Iterator, Optional
+
+from torchdata.nodes.base_node import BaseNode, T
+
+
+class DataLoader(BaseNode[T]):
```
andrewkho, line 6: Before we land this, let's think of a better name

andrewkho: Updated to Loader, ie from torchdata.nodes import Loader

divyanshk (approved):

PR #1361, opened 2024-11-12
Weighted Multi dataset mixer

divyanshk:
This PR includes a new multi-dataset mixer using weights, supporting multiple stopping criteria.


test/nodes/test_multi_dataset_weighted_sampler.py
```diff
@@ -0,0 +1,99 @@
...
+        self.assertEqual("ds3" in datasets_in_results, True)  # ds3 has maximum weight
+        self.assertEqual("ds0" in datasets_in_results, False)  # ds0 has minimum weight
+
+    def test_multi_dataset_weighted_sampler_all_dataset_exhausted(self) -> None:
+        mixer = MultiDatasetWeightedSampler(
+            self.datasets,
+            self.weights,
+            stopping_criterion=StopCriteria.ALL_DATASETS_EXHAUSTED,
+        )
+        results = list(mixer)
```
andrewkho, line 78: Do you want to add multi-epoch tests here?

test/nodes/test_multi_dataset_weighted_sampler.py
```diff
@@ -0,0 +1,99 @@
...
+
+class TestMultiDatasetWeightedSampler(TestCase):
+    def setUp(self) -> None:
+        super().setUp()
+        self._num_samples = 2
+        self._num_datasets = 4
+        self._weights_fn = lambda i: 0.1 * (i + 1)
+
+        self.datasets = {
+            f"ds{i}": IterableWrapper(DummyIterableDataset(self._num_samples, f"ds{i}"))
```
andrewkho, line 25: It'd be nice to test multiple dataset versions here, ie some generator based

andrewkho: meaning wrapping a generator, not straight generator as those are disallowed

torchdata/nodes/samplers/multi_dataset_weighted_sampler.py
```diff
@@ -0,0 +1,190 @@
...
+        if any(weight < 0 for weight in self.weights.values()):
+            raise ValueError(
+                "For multi-dataset weighted sampling, weights must be non-negative",
+            )
+
+        # Check if any weight is 0
+        if any(weight == 0 for weight in self.weights.values()):
+            raise ValueError(
+                "For multi-dataset weighted sampling, weights must be non-zero",
+            )
```
andrewkho, line 67: Let's consolidate these into a single check

torchdata/nodes/samplers/multi_dataset_weighted_sampler.py
```diff
@@ -0,0 +1,190 @@
...
+            # Force a fresh iterator from all source nodes
+            self._weighted_sampler = _WeightedSampler(
+                weights=self.weights,
+                generator=g,
+            )
+            self._datasets_exhausted = [False] * len(self.weights)
+            for k in self.dataset_names:
+                self.source_nodes[k].reset()
+
+        self._ds_iters = [iter(self.source_nodes[k]) for k in self._weighted_sampler.names]
```
andrewkho, line 93: Let's delete this since source_nodes are themselves iterators already

torchdata/nodes/samplers/multi_dataset_weighted_sampler.py
```diff
@@ -0,0 +1,190 @@
...
+            # Raise StopIteration if all datasets are exhausted
+            raise StopIteration()
+
+        idx, key = next(self._weighted_sampler)
+        try:
+            if self._datasets_exhausted[idx] and self.stopping_criterion == StopCriteria.ALL_DATASETS_EXHAUSTED:
+                # Before fetching a new item check if idx corresponds to an already
+                # exhaused dataset and StopCriteria is ALL_DATASETS_EXHAUSTED, move to next idx
+                return self.next()
+            item = next(self._ds_iters[idx])
```
andrewkho, line 106:
```suggestion
            item = next(self.source_nodes[key])
```

torchdata/nodes/samplers/multi_dataset_weighted_sampler.py
```diff
@@ -0,0 +1,190 @@
...
+                raise StopIteration()
+
+            # If StopCriteria is ALL_DATASETS_EXHAUSTED, move to next idx
+            if self.stopping_criterion == StopCriteria.ALL_DATASETS_EXHAUSTED:
+                return self.next()
+
+            # Reset the iterator and try again
+            self._ds_iters[idx].reset()
+            self._ds_iters[idx] = iter(self.source_nodes[key])
+            item = next(self._ds_iters[idx])
```
andrewkho, line 129:
```suggestion
            self.source_nodes[key].reset()
            item = next(self.source_nodes[key])
```

torchdata/nodes/samplers/multi_dataset_weighted_sampler.py
```diff
@@ -0,0 +1,190 @@
...
+
+    def get_state(self) -> Dict[str, Any]:
+        return {
+            self.GENERATOR_STATE_KEY: self._weighted_sampler.state_dict(),
+            self.DATASETS_EXHAUSTED_KEY: copy.deepcopy(self._datasets_exhausted),
+            self.DATASET_NODE_STATES_KEY: {k: self.source_nodes[k].state_dict() for k in self.dataset_names},
+        }
+
+
+class _WeightedSampler:
```
andrewkho, line 141: Would this be reused? If yes let's put it in utils

torchdata/nodes/samplers/multi_dataset_weighted_sampler.py
```diff
@@ -0,0 +1,190 @@
...
+
+import copy
+from typing import Any, Dict, Mapping, Optional
+
+import torch
+from torchdata.nodes.base_node import BaseNode, T
+from torchdata.nodes.samplers.utils import StopCriteria
+
+
+class MultiDatasetWeightedSampler(BaseNode[T]):
```
andrewkho, line 15: Maybe we should call this `MultiNodeWeightedSampler` or something similar?

test/nodes/test_multi_dataset_weighted_sampler.py
```diff
@@ -0,0 +1,99 @@
...
+
+    def test_multi_dataset_weighted_sampler_cycle_until_all_exhausted(self) -> None:
+        mixer = self.weighted_sampler_node
+        results = list(mixer)
+        datasets_in_results = {result["name"] for result in results}
+        self.assertEqual(sorted(datasets_in_results), ["ds0", "ds1", "ds2", "ds3"])
+
+    @parameterized.expand([2, 5, 7])
+    def test_save_load_state_stateful(self, midpoint: int):
+        run_test_save_load_state(self, self.weighted_sampler_node, midpoint)
```
andrewkho, line 99: Let's parameterize this over the different stop criteria

torchdata/nodes/samplers/multi_dataset_weighted_sampler.py
```diff
@@ -0,0 +1,190 @@
...
+
+    DATASET_NODE_STATES_KEY = "dataset_node_states"
+    GENERATOR_STATE_KEY = "g_state"
+    DATASETS_EXHAUSTED_KEY = "_datasets_exhausted"
+
+    def __init__(
+        self,
+        source_nodes: Mapping[str, BaseNode[T]],
+        weights: Dict[str, float],
+        stopping_criterion: str = StopCriteria.CYCLE_UNTIL_ALL_DATASETS_EXHAUSTED,
```
andrewkho, line 26: nit: let's keep consistent in naming 
```suggestion
        stop_criterion: str = StopCriterion.CYCLE_UNTIL_ALL_DATASETS_EXHAUSTED,
```

torchdata/nodes/samplers/multi_dataset_weighted_sampler.py
```diff
@@ -0,0 +1,190 @@
...
+        self._validate()
+
+    def _validate(self) -> None:
+        if self.stopping_criterion not in [
+            StopCriteria.CYCLE_UNTIL_ALL_DATASETS_EXHAUSTED,
+            StopCriteria.ALL_DATASETS_EXHAUSTED,
+            StopCriteria.FIRST_DATASET_EXHAUSTED,
+        ]:
+            raise ValueError(
+                "stop_criteria must be one of: CYCLE_UNTIL_ALL_DATASETS_EXHAUSTED, FIRST_DATASET_EXHAUSTED, ALL_DATASETS_EXHAUSTED",
```
andrewkho, line 42: keep this consistent with param name

andrewkho: It's probably also useful to log the value that we received, ie add `{stop_criteria=}` somewhere in the message

andrewkho: ditto for other validations below

torchdata/nodes/samplers/utils.py
```diff
@@ -0,0 +1,24 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
+# All rights reserved.
+#
+# This source code is licensed under the BSD-style license found in the
+# LICENSE file in the root directory of this source tree.
+
+
+class StopCriteria:
```
andrewkho, line 8: Should we have a CYCLE_FOREVER option that just continuously restarts iterators until external stop signal? ie for training for fixed number of steps

andrewkho: Can add later in a separate PR

torchdata/nodes/samplers/multi_dataset_weighted_sampler.py
```diff
@@ -0,0 +1,190 @@
...
+            self.DATASET_NODE_STATES_KEY: {k: self.source_nodes[k].state_dict() for k in self.dataset_names},
+        }
+
+
+class _WeightedSampler:
+    def __init__(
+        self,
+        weights: Dict[str, float],
+        generator,
+        batch_size: int = 1000,
```
andrewkho, line 146: Can we also add a test where this loops around?

torchdata/nodes/samplers/multi_node_weighted_sampler.py
```diff
@@ -0,0 +1,216 @@
...
+            raise ValueError(
+                f"Invalid {self.stop_criteria=}. stop_criteria must be one of: CYCLE_UNTIL_ALL_DATASETS_EXHAUSTED, FIRST_DATASET_EXHAUSTED, ALL_DATASETS_EXHAUSTED"
+            )
+
+        # Check if keys of source_nodes and weights are the same
+        if set(self.dataset_names) != set(self.weights.keys()):
+            raise ValueError(
+                f"Invalid {self.weights=}. For multi-dataset weighted sampling, keys of source_nodes and weights must be the same",
+            )
+
```
ramanishsingh, line 67: Might be a dumb ques, but do we need to check if the length of weights and no. of datasets match?

torchdata/nodes/samplers/multi_node_weighted_sampler.py
```diff
@@ -0,0 +1,216 @@
...
+        for weight in self.weights.values():
+            if not isinstance(weight, float) or weight <= 0:
```
ramanishsingh, line 69: Shall we allow `weight=0`
`torch.multinomial` allows 0 weight, just that that the sum(weights) must be >0
One situation where this might be helpful could be when a user wants to run multiple experiments with multiple weight distributions
for example in case of ds1, ds2, and ds3

exp1 = [1,0,0]
exp3 = [0,1,0]
....
exp8 = [1,1,1]

torchdata/nodes/samplers/multi_node_weighted_sampler.py
divyanshk: Good point, will add that as well!

torchdata/nodes/samplers/multi_node_weighted_sampler.py
divyanshk: Weight = 0 is not good because it means user doesnt want to give any weightage to that data - which doesnt make sense, plus for certain stopping criteria the mixer will continue looping forever because that dataset will never get exhausted

andrewkho: That's a good point about the stop criteria, we should make it foolproof and raise errors if users set something that would result in weird behaviour. But agree w/Ramanish that if there are scenarios where users might want this and it doesn't conflict with other settings then we should allow it

torchdata/nodes/samplers/multi_node_weighted_sampler.py
```diff
@@ -0,0 +1,216 @@
...
+    WEIGHTED_SAMPLER_STATE_KEY = "weighted_sampler_state"
+    DATASETS_EXHAUSTED_KEY = "datasets_exhausted"
+
+    def __init__(
+        self,
+        source_nodes: Mapping[str, BaseNode[T]],
+        weights: Dict[str, float],
+        stop_criteria: str = StopCriteria.CYCLE_UNTIL_ALL_DATASETS_EXHAUSTED,
+        rank: int | None = None,
+        world_size: int | None = None,
```
andrewkho, line 31: This isn't supported yet in Python 3.9 which we still need to support, see CI:
```
E   TypeError: unsupported operand type(s) for |: 'type' and 'NoneType'
=========================== short test summary info ============================
ERROR test/nodes/test_adapters.py - TypeError: unsupported operand type(s) for |: 'type' and 'NoneType'
ERROR test/nodes/test_multi_node_weighted_sampler.py - TypeError: unsupported operand type(s) for |: 'type' and 'NoneType'
!!!!!!!!!!!!!!!!!!! Interrupted: 2 errors during collection !!!!!!!!!!!!!!!!!!!!
============================== 2 errors in 4.32s ===============================
```

torchdata/nodes/samplers/multi_node_weighted_sampler.py
```diff
@@ -0,0 +1,216 @@
...
+        # Check if keys of source_nodes and weights are the same
+        if set(self.dataset_names) != set(self.weights.keys()):
+            raise ValueError(
+                f"Invalid {self.weights=}. For multi-dataset weighted sampling, keys of source_nodes and weights must be the same",
+            )
+
+        for weight in self.weights.values():
+            if not isinstance(weight, float) or weight <= 0:
+                raise ValueError(
+                    f"Invalid {self.weights=}. For multi-dataset weighted sampling, weights must be a 1d sequence, non-negative, and non-zero"
```
andrewkho, line 71: It'd be good for user understanding to have either a brief comment, (or include directly in the error message) about why this might unintentionally cause infinite loops

torchdata/nodes/samplers/multi_node_weighted_sampler.py
```diff
@@ -0,0 +1,216 @@
...
+
+import torch
+from torchdata.nodes.base_node import BaseNode, T
+from torchdata.nodes.samplers.utils import StopCriteria
+
+from .utils import _get_rank_seed, get_rank_and_world_size
+
+
+class MultiNodeWeightedSampler(BaseNode[T]):
+    """A node that samples from multiple datasets with weights."""
```
andrewkho, line 18: This class is not super self-explanatory. We should include a lengthy docstring here with examples about the expected behaviour of different settings. See eg MDL's doc + param strings. 

Also can follow up on this after PR but before release, let's make sure the docstring styles / param descriptions / param names etc are all consistent

torchdata/nodes/samplers/utils.py
```diff
@@ -0,0 +1,59 @@
...
+# This source code is licensed under the BSD-style license found in the
+# LICENSE file in the root directory of this source tree.
+
+import os
+
+import torch
+import torch.distributed as dist
+
+
+class StopCriteria:
```
andrewkho, line 13: This is probably worth putting in its own file TBH

torchdata/nodes/samplers/multi_node_weighted_sampler.py
```diff
@@ -0,0 +1,216 @@
...
+        # when trying next(it) on already exhausted iterator
+        if self.stop_criteria == StopCriteria.FIRST_DATASET_EXHAUSTED and any(self._datasets_exhausted.values()):
+            raise StopIteration()
+
+        key = next(self._weighted_sampler)
+        try:
+            if self._datasets_exhausted[key] and self.stop_criteria == StopCriteria.ALL_DATASETS_EXHAUSTED:
+                # Before fetching a new item check if key corresponds to an already
+                # exhaused dataset and StopCriteria is ALL_DATASETS_EXHAUSTED, move to next idx
+                return self.next()
```
andrewkho, line 118: I am pretty sure this should be 
```suggestion
                return next(self)
```

torchdata/nodes/samplers/multi_node_weighted_sampler.py
```diff
@@ -0,0 +1,216 @@
...
+                    StopCriteria.ALL_DATASETS_EXHAUSTED,
+                    StopCriteria.CYCLE_UNTIL_ALL_DATASETS_EXHAUSTED,
+                )
+                and all(self._datasets_exhausted.values())
+            ):
+                raise StopIteration()
+
+            # If StopCriteria is ALL_DATASETS_EXHAUSTED, move to next idx
+            if self.stop_criteria == StopCriteria.ALL_DATASETS_EXHAUSTED:
+                return self.next()
```
andrewkho, line 137:
```suggestion
                return next(self)
```

torchdata/nodes/samplers/multi_node_weighted_sampler.py
andrewkho: actually, can we avoid the recursion here? It becomes hard to reason about. eg if this throws StopIteration, would it get caught again in the try/except?

torchdata/nodes/samplers/multi_node_weighted_sampler.py
```diff
@@ -0,0 +1,216 @@
...
+        key = next(self._weighted_sampler)
+        try:
+            if self._datasets_exhausted[key] and self.stop_criteria == StopCriteria.ALL_DATASETS_EXHAUSTED:
+                # Before fetching a new item check if key corresponds to an already
+                # exhaused dataset and StopCriteria is ALL_DATASETS_EXHAUSTED, move to next idx
+                return self.next()
+            item = next(self.source_nodes[key])
+        except StopIteration:
+            # Mark the dataset as exhausted
+            self._datasets_exhausted[key] = True
```
andrewkho, line 122:
```suggestion
        while self._datasets_exhausted[key] and self.stop_criteria == StopCriteria.ALL_DATASETS_EXHAUSTED and not all(self._datasets_exhausted.values()):
            key = next(self._weighted_sampler)
        try:
            item = next(self.source_nodes[key])
        except StopIteration:
            # Mark the dataset as exhausted
            self._datasets_exhausted[key] = True
```

test/nodes/test_multi_node_weighted_sampler.py
```diff
@@ -0,0 +1,203 @@
...
+        self._weights_fn = lambda i: 0.1 * (i + 1)
+        self._num_epochs = 3
+
+        self.datasets = {
+            f"ds{i}": IterableWrapper(DummyIterableDataset(self._num_samples, f"ds{i}"))
+            for i in range(self._num_datasets)
+        }
+        self.weights = {f"ds{i}": self._weights_fn(i) for i in range(self._num_datasets)}
+
+    def _setUpMultiNodeWeightedSampler(
```
andrewkho, line 34: nit: Use underscores when python (except for setUp which we don't have a choice in).
```suggestion
    def _setup_multi_node_wighted_sampler(
```

test/nodes/test_multi_node_weighted_sampler.py
```diff
@@ -0,0 +1,203 @@
...
+        self, num_samples, num_datasets, weights_fn, stop_criteria
+    ) -> MultiNodeWeightedSampler:
+
+        datasets = {f"ds{i}": IterableWrapper(DummyIterableDataset(num_samples, f"ds{i}")) for i in range(num_datasets)}
+        weights = {f"ds{i}": weights_fn(i) for i in range(num_datasets)}
+        return MultiNodeWeightedSampler(datasets, weights, stop_criteria)
```
andrewkho, line 40: Let's wrap this in a Prefetcher to make sure it works properly with background threads
```suggestion
        node = MultiNodeWeightedSampler(datasets, weights, stop_criteria)
        return Prefetch(node, prefetch_factor=3)
```

andrewkho (changes requested):
Looking really good, thank you! Couple of things to take care of: biggest one is let's try to remove the recursion

torchdata/nodes/samplers/multi_node_weighted_sampler.py
ramanishsingh: Thinking more about it, I think we keep this check (weight>0) for now to not get into an infinite loop. The user can still use very small weights like 1e-9 or 1e-12 and practically get the same sampling effect. We can think about allowing zero weights later on.

torchdata/nodes/samplers/multi_node_weighted_sampler.py
```diff
@@ -0,0 +1,224 @@
...
+
+        # Raise StopIteration is StopCriteria is FIRST_DATASET_EXHAUSTED and
+        # the first dataset is exhausted. Doing this to correctly catch StopIteration
+        # when trying next(it) on already exhausted iterator
+        if self.stop_criteria == StopCriteria.FIRST_DATASET_EXHAUSTED and any(self._datasets_exhausted.values()):
+            raise StopIteration()
+
+        return
+
+    def next(self) -> T:
```
divyanshk, line 119: @andrewkho @ramanishsingh Update this method to remove the recursive call, but made the entire next() run inside a while loop which breaks after the first iteration - unless there is a redundant key in which cases to goes to the top and starts over. I am using a while loop so that we can check for early stop iteration (also moved the stop iteration checks out in a separate function). LMK if this make sense - will be adding more doc strings in next commit. Thanks.

andrewkho: This is looking good, will mostly be relying on tests to ensure this works as intended

torchdata/nodes/samplers/multi_node_weighted_sampler.py
```diff
@@ -0,0 +1,224 @@
...
+                    continue
+                item = next(self.source_nodes[key])
+            except StopIteration:
+                # Mark the dataset as exhausted
+                self._datasets_exhausted[key] = True
+
+                # Based on updated _check_for_stop_iteration, check if we should raise StopIteration
+                self._check_for_stop_iteration()
+
+                # If StopCriteria is ALL_DATASETS_EXHAUSTED, move to next key
```
andrewkho, line 139: Let's describe the else condition here (ie we should only hit "else" if "cycle_until_all_datsets_exhausted" and in that case we need to reset). Wondering if you think it would be more clear to frame the condition that way: ie 
```
if CYCLE_UNTIL_ALL_DATASESTS_EXHAUSTED:
  node.reset()
  item = next(node)
  break
else:
  continue
```

torchdata/nodes/samplers/multi_node_weighted_sampler.py
```diff
@@ -0,0 +1,224 @@
...
+                if self.stop_criteria == StopCriteria.ALL_DATASETS_EXHAUSTED:
+                    # return next(self)  # omit recursive call
+                    # key = next(self._weighted_sampler)
+                    continue
+
+                # Reset the iterator and try again
+                self.source_nodes[key].reset()
+                item = next(self.source_nodes[key])
```
andrewkho, line 147: What happens if this node throws stop iteration after reset?

torchdata/nodes/samplers/multi_node_weighted_sampler.py
```diff
@@ -0,0 +1,224 @@
...
+                # Mark the dataset as exhausted
+                self._datasets_exhausted[key] = True
+
+                # Based on updated _check_for_stop_iteration, check if we should raise StopIteration
+                self._check_for_stop_iteration()
+
+                # If StopCriteria is ALL_DATASETS_EXHAUSTED, move to next key
+                if self.stop_criteria == StopCriteria.ALL_DATASETS_EXHAUSTED:
+                    # return next(self)  # omit recursive call
+                    # key = next(self._weighted_sampler)
```
andrewkho, line 142: Delete these before landing please

torchdata/nodes/samplers/multi_node_weighted_sampler.py
```diff
@@ -0,0 +1,224 @@
...
+        while True:
+            self._check_for_stop_iteration()
+
+            # Fetch the next item's key from the weighted sampler
+            key = next(self._weighted_sampler)
+            try:
+                if self._datasets_exhausted[key] and self.stop_criteria == StopCriteria.ALL_DATASETS_EXHAUSTED:
+                    # Before fetching a new item check if key corresponds to an already
+                    # exhaused dataset and StopCriteria is ALL_DATASETS_EXHAUSTED, move to next key
+                    # return next(self)  # omit recursive call
```
andrewkho, line 129: delete before landing please

torchdata/nodes/samplers/multi_node_weighted_sampler.py
divyanshk: At line 146, we are already inside the catch block for StopIteration - i.e. dataset at `key` is at the end. Reseting it and getting a new item shouldn't return a Stop Iteration - what do you think?

To get rid of this next call here, we can "persist" the key so that the next loop call goes on correctly, but I feel that won't read very clean.

torchdata/nodes/loader.py
```diff
@@ -47,6 +47,7 @@ def __init__(
         self.root = loader.root
         self._cached_item = None
         self._num_yielded = 0
+        self._cached_state_dict: Optional[Dict[str, Any]] = None
```
andrewkho, line 50: Let's split this into another PR

torchdata/nodes/samplers/multi_node_weighted_sampler.py
andrewkho: I'm thinking of edge cases where after .reset(), it's expected to throw StopIteration right away. Eg iterating over an empty array, or the iterator is meant to be one time use.... maybe throwing is the right thing to do in this case.

torchdata/nodes/samplers/stop_criteria.py
```diff
@@ -0,0 +1,24 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
+# All rights reserved.
+#
+# This source code is licensed under the BSD-style license found in the
+# LICENSE file in the root directory of this source tree.
+
+
+class StopCriteria:
+    """
+    Stopping criteria for the dataset samplers.
```
andrewkho, line 10: When @ramanishsingh needs to use this, or when we clean up docs, let's see if we can figure out a nice way to structure this so that we don't need to repeat / maintain multiple copies of similar docstrings

andrewkho (approved):
Split out the loader.py changes and then go go gooooo

PR #1367, opened 2024-11-15
Delete conda builds

### LINKED ISSUE #1365: [Release] Delete Conda Nightly builds
### 🐛 Describe the bug

PyTorch Core is deprecating Nightly Conda builds. see https://github.com/pytorch/pytorch/issues/138506 

We need to clean ours up too, preferably before core deletes on Nov. 18 so we don't have broken builds.

cc @atalman 

### Versions

nightly

ramanishsingh:
PyTorch Core is [deprecating](https://github.com/pytorch/pytorch/issues/138506) Nightly Conda builds. 
This PR aims to delete conda builds
Fixes #1365


.github/workflows/pull_release.yml
```diff
@@ -7,6 +7,10 @@ on:
       - .github/workflows/pull_release.yml
       - .github/workflows/_build_test_upload.yml
 
+permissions:
+  contents: write
+  id-token: write
+
```
andrewkho, line 13: Do we need this change?

ramanishsingh: I was getting this error 

`The workflow is not valid. .github/workflows/pull_release.yml (Line: 11, Col: 3): Error calling workflow 'pytorch/data/.github/workflows/_build_test_upload.yml@45dc471d9a16eb4cf482f38b9167f99a13cd55d4'. The workflow is requesting 'contents: write, id-token: write', but is only allowed 'contents: read, id-token: none'.`

https://github.com/pytorch/data/actions/runs/11864790565

After I added these lines it worked. I can try to again remove these and see what happens now.

packaging/post_build_script_linux.sh
```diff
@@ -7,8 +7,8 @@ for pkg in dist/torchdata*.whl; do
     echo "PkgInfo of $pkg:"
     pkginfo $pkg
 
-    auditwheel repair $pkg --plat manylinux2014_x86_64 -w wheelhouse
```
atalman, line 10: Looks like we are mixing manylinux here and conda removal

.github/workflows/pull_release.yml
atalman: yes this maybe needed

packaging/post_build_script_linux.sh
ramanishsingh: Thanks, updated (reverted).

atalman (approved):
lgtm

PR #1369, opened 2024-11-20
Add epoch based seed to MultiNodeWeightedSampler

divyanshk:
This PR add a new method `set_epoch` to `MultiNodeWeightedSampler`. This takes in an epoch and call `reset()` to initialize a new weighted sampler based on the epoch.

This can be called on its own like so
```
mixer = Mixer(...)
for _ in range(num_epochs):
    mixer.set_epoch(i)
    results = list(mixer)
    mixer.reset()
```
or with `SamplerWrapper` like so
```
mixer = Mixer(...)
node = SamplerWrapper(...)
for _ in range(num_epochs):
    results = list(node)
    node.reset()
```
The second might be more desirable if plugging this with `Loader`.

EDIT: Adding a `EpochUpdater` node.
Usage:
```
mixer = Mixer(...)
node = EpochUpdater(...)
for _ in range(num_epochs):
    results = list(node)
    node.reset()
```

EDIT 2: Remove `EpochUpdater`. Moved the epoch update logic inside mixer class, and make epoch be a RNG. Including changes for initial state fix for Prefetcher (https://github.com/pytorch/data/pull/1377)


torchdata/nodes/samplers/multi_node_weighted_sampler.py
```diff
@@ -52,6 +53,7 @@ class MultiNodeWeightedSampler(BaseNode[T]):
     NUM_YIELDED_KEY = "num_yielded"
     WEIGHTED_SAMPLER_STATE_KEY = "weighted_sampler_state"
     DATASETS_EXHAUSTED_KEY = "datasets_exhausted"
+    EPOCH_KEY = "epoch"
```
andrewkho, line 56: nit: keep these in the same order as the docstring above

torchdata/nodes/samplers/multi_node_weighted_sampler.py
```diff
@@ -69,6 +71,8 @@ def __init__(
         self.dataset_names = list(self.source_nodes.keys())
         self._num_yielded = 0
         self.seed = seed
+        self.epoch = 0
```
andrewkho, line 74: we should give users some way to initialize/update this as well

torchdata/nodes/samplers/multi_node_weighted_sampler.py
```diff
@@ -126,11 +133,16 @@ def reset(self, initial_state: Optional[Dict[str, Any]] = None):
...
+                epoch=self.epoch,
             )
             self._datasets_exhausted = {key: False for key in self.weights.keys()}
             for k in self.dataset_names:
                 self.source_nodes[k].reset()
 
+    def set_epoch(self, epoch: int) -> None:
+        self.epoch = epoch
+        self.reset(self._initial_state)
+
```
andrewkho, line 145: Let's discuss this a bit more, i don't think we should call reset implicitly here as it will recursively reset all sub nodes

torchdata/nodes/samplers/multi_node_weighted_sampler.py
```diff
@@ -107,12 +111,15 @@ def reset(self, initial_state: Optional[Dict[str, Any]] = None):
         super().reset(initial_state)
 
         if initial_state is not None:
+            self._initial_state = initial_state
```
andrewkho, line 114: When will self._initial_state be reset? I'd prefer a solution where we don't need to hold on to this

andrewkho (changes requested):
Let's discuss this a bit more, I'm not convinced yet that the design is right

divyanshk:
Updated the PR for adding epoch based seed into the mixer.

Context: The feedback was not having to `reset` the base nodes on every `set_epoch` call for the mixer. That is valid, and the main reason I put that in was because I was trying to make things fit well with the `SamplerWrapper`, and `SamplerWrapper` doesn’t recursively call `reset` on its underlying sampler (it doesn’t need to). Let's leave the `SamplerWrapper` aside and let it serve its specific purpose.

1. I updated the PR to add a new node `EpochUpdater` which is basically SamplerWrapper` but with a reset call to its underlying node (like we would expect from any new node). 

2. Also, make the `set_epoch` update the underlying `_weighted_sampler` so that the effect of the epoch update kicks in. Without this, we need to ensure the `mixer.set_epoch()` calls always directly proceeds a `node.reset()` call - which is unintuitive as ideally we would want to define the epoch at the start of every loop, iterate, reset, repeat.

torchdata/nodes/epoch_updater.py
```diff
@@ -0,0 +1,66 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
```
andrewkho, line 1: Can we keep this in the MultiNodeWeightedSampler instead of as a separate node?  I'm not sure we're ready to break it out yet

test/nodes/test_multi_node_weighted_sampler.py
```diff
@@ -32,12 +30,11 @@ def setUp(self) -> None:
         }
         self.weights = {f"ds{i}": self._weights_fn(i) for i in range(self._num_datasets)}
 
-    def _setup_multi_node_wighted_sampler(self, num_samples, num_datasets, weights_fn, stop_criteria) -> Prefetcher:
+    def _setup_multi_node_wighted_sampler(self, num_samples, num_datasets, weights_fn, stop_criteria):
 
         datasets = {f"ds{i}": IterableWrapper(DummyIterableDataset(num_samples, f"ds{i}")) for i in range(num_datasets)}
         weights = {f"ds{i}": weights_fn(i) for i in range(num_datasets)}
-        node = MultiNodeWeightedSampler(datasets, weights, stop_criteria)
-        return Prefetcher(node, prefetch_factor=3)
```
divyanshk, line 40: @andrewkho 

I moved the the epoch update logic inside mixer. The tests for the mixer nodes itself pass. 

But one thing thats not fitting in is when we add a Prefetcher node on top. Specifically, the Prefetcher node's state is None for the underlying node (`{'root': {'snapshot': None, 'steps_since_snapshot': 0}, 'num_yielded': 0}`) right after initialization (https://github.com/pytorch/data/blob/main/test/nodes/utils.py#L98). I think that is causing the epoch update logic to fail when we try to load it back here (https://github.com/pytorch/data/blob/main/test/nodes/utils.py#L136).

torchdata/nodes/_populate_queue.py
```diff
@@ -56,18 +56,18 @@ def _put(item, block: bool = True, snapshot: Optional[Dict[str, Any]] = None):
         assert (
             isinstance(snapshot_frequency, int) and snapshot_frequency >= 0
         ), f"snapshot_frequency must be non-negative integer! Got {snapshot_frequency}"
-        src_iter = iter(source)
+        snapshot_store.append(snapshot=source.state_dict(), version=-1)
```
andrewkho, line 59: Let's split these changes into a separate PR

andrewkho: we should also add a test to catch this

torchdata/nodes/samplers/multi_node_weighted_sampler.py
```diff
@@ -63,21 +65,34 @@ def __init__(
...
         self.stop_criteria = stop_criteria
         self.dataset_names = list(self.source_nodes.keys())
         self._num_yielded = 0
+        self._started = False
+
+        # Setup epoch generator for multi-epoch seeds
         self.seed = seed
+        self._g_epoch = torch.Generator()
+        self._g_epoch.manual_seed(self.seed)
+        self._epoch_random_seed = 0
```
andrewkho, line 81: Should we use `_get_epoch_seed` here?

torchdata/nodes/samplers/utils.py
```diff
@@ -10,11 +10,15 @@
 import torch.distributed as dist
 
 
-def _get_rank_seed(seed: int, generator_rank: torch.Generator, rank: int, world_size: int) -> int:
-    generator_rank.manual_seed(seed * world_size + rank)
+def _get_rank_seed(seed: int, generator_rank: torch.Generator, rank: int, world_size: int, epoch: int) -> int:
+    generator_rank.manual_seed(seed * world_size + rank + epoch)
```
andrewkho, line 14: Will this cause create a pattern between ranks when going across epochs?

eg)
                                Epoch ->
Seeds for Rank 0/2: [b, c, d, e, f, g, a, ...]
Seeds for Rank 1/2: [a, b, c, d, e, f, g, h, ...]

andrewkho: Maybe something like this instead might work:
```suggestion
    generator_rank.manual_seed(seed * world_size + rank)
    (assuming epoch is incremented instead of randomly generated)
    for _ in range(epoch):
      __consumes_rng_state(generator_rank)
```

torchdata/nodes/samplers/utils.py
```diff
@@ -10,11 +10,15 @@
...
     return int(torch.randint(0, 2 ** 32 - 1, size=(1,), generator=generator_rank).item())
 
 
+def _get_epoch_seed(generator_epoch: torch.Generator) -> int:
+    return int(torch.randint(0, 2 ** 32 - 1, size=(1,), generator=generator_epoch).item())
```
andrewkho, line 19: I'm not totally convinced we need RNG here, why not just increment?

torchdata/nodes/samplers/multi_node_weighted_sampler.py
```diff
@@ -63,21 +65,31 @@ def __init__(
         seed: int = 0,
     ) -> None:
         super().__init__()
+
+        # Initialize basic attributes
```
andrewkho, line 69: we don't need this comment

torchdata/nodes/samplers/utils.py
```diff
@@ -10,9 +10,16 @@
...
     generator_rank.manual_seed(seed * world_size + rank)
-    return int(torch.randint(0, 2 ** 32 - 1, size=(1,), generator=generator_rank).item())
+    _random_seed: int = 0
+    for _ in range(epoch):
+        _random_seed = int(torch.randint(0, 2 ** 32 - 1, size=(1,), generator=generator_rank).item())
+    return _random_seed
+
+
+def _update_epoch_seed(epoch_seed: int) -> int:
+    return epoch_seed + 1
```
andrewkho, line 22: We don't need a function for this

torchdata/nodes/samplers/multi_node_weighted_sampler.py
```diff
@@ -63,21 +65,31 @@ def __init__(
...
         self.seed = seed
 
+        # Setup rank and world size
         if rank is None or world_size is None:
             self.rank, self.world_size = get_rank_and_world_size()
         else:
             self.rank = rank
             self.world_size = world_size
 
+        self._epoch_random_seed = _update_epoch_seed(self.seed * self.world_size + self.rank)
```
andrewkho, line 85:
```suggestion
        self._epoch = 0
```

Would this ^ still work?

torchdata/nodes/samplers/utils.py
```diff
@@ -10,9 +10,16 @@
 import torch.distributed as dist
 
 
-def _get_rank_seed(seed: int, generator_rank: torch.Generator, rank: int, world_size: int) -> int:
+def _get_rank_seed(seed: int, generator_rank: torch.Generator, rank: int, world_size: int, epoch: int) -> int:
     generator_rank.manual_seed(seed * world_size + rank)
-    return int(torch.randint(0, 2 ** 32 - 1, size=(1,), generator=generator_rank).item())
+    _random_seed: int = 0
+    for _ in range(epoch):
+        _random_seed = int(torch.randint(0, 2 ** 32 - 1, size=(1,), generator=generator_rank).item())
```
andrewkho, line 17:
```suggestion
    _random_seed = int(torch.randint(0, 2 ** 32 - 1, size=(epoch+1,), generator=generator_rank)[-1].item())
```
Is this equivalent? it would probably be faster than running torch.randint x epoch times

andrewkho (changes requested):
Couple of minor comments

torchdata/nodes/samplers/multi_node_weighted_sampler.py
```diff
@@ -63,21 +65,30 @@ def __init__(
...
+        # Setup rank and world size
         if rank is None or world_size is None:
             self.rank, self.world_size = get_rank_and_world_size()
         else:
             self.rank = rank
             self.world_size = world_size
 
+        self._epoch_random_seed = 0
+
+        # Validate input
```
andrewkho, line 86: We can drop all these comments, the code itself is fairly self-explanatory

torchdata/nodes/samplers/multi_node_weighted_sampler.py
```diff
@@ -63,21 +65,30 @@ def __init__(
...
         self.seed = seed
 
+        # Setup rank and world size
         if rank is None or world_size is None:
             self.rank, self.world_size = get_rank_and_world_size()
         else:
             self.rank = rank
             self.world_size = world_size
 
+        self._epoch_random_seed = 0
```
andrewkho, line 84: Let's just call this `_epoch` for brevity, since it's now roughly what it's trying to capture

torchdata/nodes/samplers/multi_node_weighted_sampler.py
```diff
@@ -49,9 +50,10 @@ class MultiNodeWeightedSampler(BaseNode[T]):
     """
 
     DATASET_NODE_STATES_KEY = "dataset_node_states"
+    DATASETS_EXHAUSTED_KEY = "datasets_exhausted"
+    EPOCH_RANDOM_SEED_KEY = "epoch_random_seed"
```
andrewkho, line 54: Let's just call this epoch too

torchdata/nodes/samplers/multi_node_weighted_sampler.py
```diff
@@ -63,21 +65,30 @@ def __init__(
...
             self.rank = rank
             self.world_size = world_size
 
+        self._epoch_random_seed = 0
+
+        # Validate input
         self._validate()
 
+        # Setup weighted sampler
+        self._weighted_sampler = self._get_new_weighted_sampler()
```
andrewkho, line 90: We can probably drop this too now and just have it always run during reset()

torchdata/nodes/samplers/multi_node_weighted_sampler.py
```diff
@@ -105,31 +116,35 @@ def _validate(self) -> None:
...
-            self._weighted_sampler = _WeightedSampler(
-                weights=self.weights,
-                seed=self.seed,
-                rank=self.rank,
-                world_size=self.world_size,
-            )
+
+            if self._started:
+                self._epoch_random_seed += 1
+                self._weighted_sampler = self._get_new_weighted_sampler()
```
andrewkho, line 132:
```suggestion
            if self._started:
                self._epoch_random_seed += 1
            self._weighted_sampler = self._get_new_weighted_sampler()
```
I think this works if we drop the call in `__init__`, it's a little easier to reason about

andrewkho (commented):
Looking really good, thanks! Just a couple of nits to tighten it up and should be good to go

torchdata/nodes/samplers/multi_node_weighted_sampler.py
divyanshk: We do not guarantee if this is exactly the epoch, do we still want to call it _epoch ?

torchdata/nodes/samplers/multi_node_weighted_sampler.py
```diff
@@ -49,9 +50,10 @@ class MultiNodeWeightedSampler(BaseNode[T]):
     """
 
     DATASET_NODE_STATES_KEY = "dataset_node_states"
+    DATASETS_EXHAUSTED_KEY = "datasets_exhausted"
+    EPOCH = "epoch"
```
andrewkho, line 54:
```suggestion
    EPOCH_KEY = "epoch"
```

andrewkho (approved):
Just one nit about consistent variable names, please fix before merge

torchdata/nodes/samplers/multi_node_weighted_sampler.py
divyanshk: good catch! thanks

divyanshk:
@andrewkho there seems to a a failure with a test `TestDequeSnapshotStore.test_thread_dead_error` on `macOS-latest 3.12` This one was failing in the previous run as well, but seems to work fine on my machine. Would you know more about it?

PR #1370, opened 2024-11-21
Multidataset round robin sampler

ramanishsingh:
This PR attempts to add a new sampler, Round robin sampler.

### DIFF
--- a/test/nodes/test_multi_node_round_robin_sampler.py
+++ b/test/nodes/test_multi_node_round_robin_sampler.py
@@ -16,9 +16,7 @@ from torchdata.nodes.adapters import IterableWrapper
 from torchdata.nodes.batch import Batcher
 from torchdata.nodes.loader import Loader
 from torchdata.nodes.prefetch import Prefetcher
-from torchdata.nodes.samplers.multi_node_round_robin_sampler import (
-    MultiNodeRoundRobinSampler,
-)
+from torchdata.nodes.samplers.multi_node_round_robin_sampler import MultiNodeRoundRobinSampler
 from torchdata.nodes.samplers.stop_criteria import StopCriteria
 
 from .utils import DummyIterableDataset, run_test_save_load_state
@@ -32,10 +30,7 @@ class TestMultiNodeRoundRobinSampler(TestCase):
 
     def get_equal_dataset(self, num_samples, num_datasets):
         """Returns a dictionary of datasets with the same number of samples"""
-        datasets = {
-            f"ds{i}": IterableWrapper(DummyIterableDataset(num_samples, f"ds{i}"))
-            for i in range(num_datasets)
-        }
+        datasets = {f"ds{i}": IterableWrapper(DummyIterableDataset(num_samples, f"ds{i}")) for i in range(num_datasets)}
         return datasets
 
     def get_unequal_dataset(self, num_samples, num_datasets):
@@ -44,24 +39,19 @@ class TestMultiNodeRoundRobinSampler(TestCase):
         datasets = {"ds0":[0], "ds1":[0, 1], "ds2":[0, 1, 2]}
         """
         datasets = {
-            f"ds{i}": IterableWrapper(DummyIterableDataset(num_samples + i, f"ds{i}"))
-            for i in range(num_datasets)
+            f"ds{i}": IterableWrapper(DummyIterableDataset(num_samples + i, f"ds{i}")) for i in range(num_datasets)
         }
         return datasets
 
     def test_empty_datasets(self) -> None:
         datasets = self.get_equal_dataset(0, self._num_datasets)
-        sampler = MultiNodeRoundRobinSampler(
-            datasets, StopCriteria.FIRST_DATASET_EXHAUSTED
-        )
+        sampler = MultiNodeRoundRobinSampler(datasets, StopCriteria.FIRST_DATASET_EXHAUSTED)
         for item in sampler:
             self.fail("Expected no batches as each dataset is empty.")
 
     def test_empty_datasets_batched(self) -> None:
         datasets = self.get_equal_dataset(0, self._num_datasets)
-        sampler = MultiNodeRoundRobinSampler(
-            datasets, StopCriteria.FIRST_DATASET_EXHAUSTED
-        )
+        sampler = MultiNodeRoundRobinSampler(datasets, StopCriteria.FIRST_DATASET_EXHAUSTED)
         batch_size = 3
         batcher = Batcher(sampler, batch_size=batch_size)
         for batch in batcher:
@@ -70,9 +60,7 @@ class TestMultiNodeRoundRobinSampler(TestCase):
     @parameterized.expand([4, 8, 16])
     def test_single_dataset(self, num_samples: int) -> None:
         datasets = self.get_equal_dataset(num_samples, 1)
-        sampler = MultiNodeRoundRobinSampler(
-            datasets, StopCriteria.FIRST_DATASET_EXHAUSTED
-        )
+        sampler = MultiNodeRoundRobinSampler(datasets, StopCriteria.FIRST_DATASET_EXHAUSTED)
         for num_sample, item in enumerate(sampler):
             pass
         self.assertEqual(num_sample + 1, num_samples)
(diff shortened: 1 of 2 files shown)


test/nodes/test_multi_node_round_robin_sampler.py
```diff
@@ -0,0 +1,157 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
+# All rights reserved.
+#
+# This source code is licensed under the BSD-style license found in the
+# LICENSE file in the root directory of this source tree.
+
+import collections
+import itertools
+from enum import unique
```
divyanshk, line 9: Looks like these imports are not used.

torchdata/nodes/samplers/multi_node_round_robin_sampler.py
```diff
@@ -0,0 +1,137 @@
...
+# LICENSE file in the root directory of this source tree.
+
+import copy
+from typing import Any, Dict, Mapping, Optional
+
+from torchdata.nodes.base_node import BaseNode, T
+from torchdata.nodes.samplers.stop_criteria import StopCriteria
+
+
+class MultiNodeRoundRobinSampler(BaseNode[T]):
```
divyanshk, line 14: What is the expected behavior when doing many epochs over all the datasets? Would every epoch be the same?

torchdata/nodes/samplers/multi_node_round_robin_sampler.py
```diff
@@ -0,0 +1,137 @@
...
+
+    def _validate_stop_criteria(self) -> None:
+        if self.stop_criteria not in [
+            StopCriteria.CYCLE_UNTIL_ALL_DATASETS_EXHAUSTED,
+            StopCriteria.ALL_DATASETS_EXHAUSTED,
+            StopCriteria.FIRST_DATASET_EXHAUSTED,
+        ]:
+            raise ValueError(
+                f"Invalid {self.stop_criteria=}. stop_criteria must be one of: CYCLE_UNTIL_ALL_DATASETS_EXHAUSTED,  , ALL_DATASETS_EXHAUSTED"
+            )
```
divyanshk, line 62:
```suggestion
            raise ValueError(
                f"Invalid {self.stop_criteria=}. stop_criteria must be one of: CYCLE_UNTIL_ALL_DATASETS_EXHAUSTED, ALL_DATASETS_EXHAUSTED, FIRST_DATASET_EXHAUSTED"
            )
```

torchdata/nodes/samplers/multi_node_round_robin_sampler.py
ramanishsingh: Yes, every epoch would be the same. Added a new test for that. https://github.com/pytorch/data/blob/5f6370502f38c7a4211af6e54123f060e0f7f151/test/nodes/test_multi_node_round_robin_sampler.py#L177

torchdata/nodes/samplers/multi_node_round_robin_sampler.py
```diff
@@ -0,0 +1,137 @@
...
+
+class MultiNodeRoundRobinSampler(BaseNode[T]):
+    """A node that samples from multiple datasets in a round robin fashion.
+
+    This node expects to take in a dictionary of source nodes.
+
+    The node implements the state using the following keys:
+    - CURRENT_DATASET_INDEX_KEY: The index of the current dataset.
+    - DATASETS_EXHAUSTED_KEY: A dictionary of booleans indicating whether each source node is exhausted.
+    - DATASET_NODE_STATES_KEY: A dictionary of states for each source node.
```
divyanshk, line 22: Nit: can we swap the order of  DATASETS_EXHAUSTED_KEY and DATASET_NODE_STATES_KEY to make it alphabetic.

torchdata/nodes/samplers/multi_node_round_robin_sampler.py
```diff
@@ -0,0 +1,137 @@
...
+
+    We support multiple stopping criteria:
+    - CYCLE_UNTIL_ALL_DATASETS_EXHAUSTED: Cycle through the source nodes until all datasets
+        are exhausted. This is the default behavior.
+    - FIRST_DATASET_EXHAUSTED: Stop when the first dataset is exhausted.
```
divyanshk, line 27: Say we have 5 datasets. 1st one is made up only one element. We pull from that, then iterate over the rest. On second loop the first one will give stopiteration - we mark it as exhausted. If stopping criteria is FIRST_DATASET_EXHAUSTED then we return with a StopIteration. I hope I got that right.

I was wondering if we should mention this in the docs here (and in the weighted sampler). So that folks understand that 1) we mark exhaustion on getting stop iteration 2) we continue to iterate over other source nodes even if one of the other source nodes is done giving out all it had, unless we iterate over it and get a StopIteration.

torchdata/nodes/samplers/multi_node_round_robin_sampler.py
```diff
@@ -0,0 +1,137 @@
...
+    - ALL_DATASETS_EXHAUSTED: Stop when all datasets are exhausted.
+
+    On complete exhaustion of the source nodes, the node will raise StopIteration.
+
+    Parameters:
```
divyanshk, line 32: Lets calls this `Args` instead of `Parameters`

torchdata/nodes/samplers/multi_node_round_robin_sampler.py
```diff
@@ -0,0 +1,137 @@
...
+            raise ValueError(
+                f"Invalid {self.stop_criteria=}. stop_criteria must be one of: CYCLE_UNTIL_ALL_DATASETS_EXHAUSTED, ALL_DATASETS_EXHAUSTED, FIRST_DATASET_EXHAUSTED"
+            )
+
+    def reset(self, initial_state: Optional[Dict[str, Any]] = None):
+        super().reset(initial_state)
+
+        if initial_state is not None:
+            self._datasets_exhausted = initial_state[self.DATASETS_EXHAUSTED_KEY]
+            for k in range(self.num_datasets):
```
divyanshk, line 69: I wonder if it is easier to change the ordering in the list when resumption? For the weighted sampler we have keys so if someone is messing with the keys its their problem, but seems like folks can re-order the source nodes (idk why maybe they were debugging something) on resumption and the state management would fail. What do you think?

Maybe a no-op for now, but something for later.

torchdata/nodes/samplers/multi_node_round_robin_sampler.py
```diff
@@ -0,0 +1,137 @@
...
+            self._check_for_stop_iteration()
+
+            current_iterator = self.source_nodes[self._current_dataset_index]
+            try:
+                if (
+                    self._datasets_exhausted[self._current_dataset_index]
+                    and self.stop_criteria == StopCriteria.ALL_DATASETS_EXHAUSTED
+                ):
+                    # Before fetching a new item check if the current dataset is already
+                    # exhaused and if StopCriteria is ALL_DATASETS_EXHAUSTED, move to next dataset
```
divyanshk, line 105: Let's move this comment before the `if` block, even in weighted sampler (my bad) thank you.

divyanshk (approved):
Left some minor comments. 

Looks good to me!

torchdata/nodes/samplers/multi_node_round_robin_sampler.py
ramanishsingh: > I hope I got that right.

Absolutely. 
Added some explanation as well as an example (will be available in the next commit).

torchdata/nodes/samplers/multi_node_round_robin_sampler.py
ramanishsingh: Sure. Thanks!

torchdata/nodes/samplers/multi_node_round_robin_sampler.py
ramanishsingh: That's a good point.
IMO it should be available. I have some ideas regarding that, but it will make this PR a little bit complicated. 
We can  land this one, and start a new one with that functionality.

divyanshk: Sounds good!

Issue #1337, opened 2024-10-15
[Stateful DataLoader] add utilities to inspect/modify dataloader state


andrewkho:
### 🚀 The feature

State Dict utilities to inspect state

### Motivation, pitch

We should provide some utilites for users to inspect and potentially modify the dataloader state. Currently the format of the dictionary is opaque, and we may want to make changes in the future. Instead of users inspecting and hard-coding paths, we should instead provide utilities to make it easy, and future-proof. 

Current list of desired utilities:
* Retrieve state for worker 'i'
* Retrieve and update the number of steps to skip upon resume

PR #1372, opened 2024-11-23
Rescalability via IBM dataset layers

daviswer:
Implements rescaling of checkpoints to different world sizes and numbers of workers. User specifies in advance the number of data partitions, and when saving/loading checkpoints with different total workers (must divide partition number evenly), stateful guarantees are maintained: seen data is not revisited until the next epoch.

Based off of the datasets in the corresponding IBM [torchtitan PR](https://github.com/pytorch/torchtitan/pull/376), but uses StatefulDataLoader and DCP to manage checkpointing from the master process. Sampling and Dummy datasets are included for demo purposes. It is possible that the IBM datasets can be merged into the existing node structure.

### Changes

- Add IBM rescalable datasets and checkpointing functions to `torchdata/stateful_dataloader/ibm_rescalable.py`
- Add demo script and correctness check to `examples/ibm_rescaling/rescaling_demo.py`

### DIFF
--- a/torchdata/stateful_dataloader/ibm_rescalable.py
+++ b/torchdata/stateful_dataloader/ibm_rescalable.py
@@ -375,8 +375,6 @@ class DummyDataset(_StatefulDataset):
             out = torch.rand(self.seqlen, generator=self.generator)
             out = out.mul(self.vocab).int().tolist()
             out[-1] = self.delimiter
-            if self.rank==0:
-                print(out)
             yield out
 
     def state_dict(self):
--- a/torchdata/stateful_dataloader/ibm_rescalable.py
+++ b/torchdata/stateful_dataloader/ibm_rescalable.py
@@ -587,23 +587,24 @@ def load_distributed_state_dict(
     nworkers = base["_snapshot"]["_main_snapshot"]["_num_workers"]
     rank = loader.dataset.rank
     dstate = __pop_dstate(base, device_mesh, [dtensor.placement_types.Shard(0)])  # placements)
+    inp = {"state":base, "dstate":dstate}
     # Read nondistributed state dict
-    ckp_ws = 0 if not os.path.exists(path) else len([x for x in os.listdir(path) if "__nondist_cp_" in x])
-    # Check that number of loaders matches
-    if ckp_ws == loader.dataset.worldsize:
-        state = torch.load(os.path.join(path, f"__nondist_cp_{rank}.pth"))
-        # Check that number of workers matches
-        if nworkers != state["_snapshot"]["_main_snapshot"]["_num_workers"]:
-            state = base
-    else:
-        # On mismatch, discard saved non-reshardable loader state and start fresh
-        state = base
+    ckp_ws = 0 if not os.path.exists(path) else len(os.listdir(path))
     # Read distributed state dict
     reader = checkpoint.FileSystemReader(path)
     checkpoint.load_state_dict(
-        dstate,
+        inp,
         reader,
     )
+    dstate = inp["dstate"]
+    # Check that number of loaders matches
+    if ckp_ws == loader.dataset.worldsize:
+        # Check that number of workers matches
+        if nworkers != state["_snapshot"]["_main_snapshot"]["_num_workers"]:
+            state = inp["state"]
+    else:
+        # On mismatch, discard saved non-reshardable loader state and start fresh
+        state = base
     # Get local tensors from dtensors, and slice over workers
     dstate = {k: v.to_local().chunk(nworkers) for k, v in dstate.items()}
     # Flip dict[list[tensor]] to list[dict[tensor]]
--- a/torchdata/stateful_dataloader/ibm_rescalable.py
+++ b/torchdata/stateful_dataloader/ibm_rescalable.py
@@ -7,7 +7,7 @@ from typing import Any, Callable, List, Optional, Set
 
 import torch
 from torch.distributed import checkpoint
-from torch.distributed.checkpoint import _load_state_dict_from_keys
+from torch.distributed.checkpoint.state_dict_loader import _load_state_dict_from_keys
 import torch.distributed.tensor as dtensor
 import torch.distributed as dist
 import torch.utils.data as data
--- a/torchdata/stateful_dataloader/ibm_rescalable.py
+++ b/torchdata/stateful_dataloader/ibm_rescalable.py
@@ -623,7 +623,7 @@ def load_distributed_state_dict(
     # Read distributed state dict
     reader = checkpoint.FileSystemReader(path)
     inp = _load_state_dict_from_keys(
-        keys=set("state", "dstate"),
+        keys=set(["state", "dstate"]),
         storage_reader = reader,
     )  # NOTE: assumes inp["state"] is same across all devices
     # checkpoint.load_state_dict(
--- a/torchdata/stateful_dataloader/ibm_rescalable.py
+++ b/torchdata/stateful_dataloader/ibm_rescalable.py
@@ -560,14 +560,15 @@ def save_distributed_state_dict(
     rank = loader.dataset.rank
     state = deepcopy(loader.state_dict())
     dstate = __pop_dstate(state, device_mesh, [dtensor.placement_types.Shard(0)])
+    out = {"state":state, "dstate":dstate}
     # Write distributed state dict
     writer = checkpoint.FileSystemWriter(path)
     checkpoint.save(
-        dstate,
+        out,
         writer,
     )
-    # Write nondistributed state dict
-    torch.save(state, os.path.join(path, f"__nondist_cp_{rank}.pth"))
+    # # Write nondistributed state dict
+    # torch.save(state, os.path.join(path, f"__nondist_cp_{rank}.pth"))
 
 
 def load_distributed_state_dict(
--- a/torchdata/stateful_dataloader/ibm_rescalable.py
+++ b/torchdata/stateful_dataloader/ibm_rescalable.py
@@ -491,7 +491,7 @@ class ScalableReader(_StatefulDataset):
         file_info = state_dict[self.statename("file_info")]
         print(shard_states.size(0), self.worldsize)
         if self.rank == 0:
-            print(shard_states)
+            print(shard_states.shape, shard_states)
         if shard_states.size(0) == self.worldsize:
             self.filesizes = file_info
             self.shard_states = shard_states[self.rank]
--- a/torchdata/stateful_dataloader/ibm_rescalable.py
+++ b/torchdata/stateful_dataloader/ibm_rescalable.py
@@ -490,6 +490,8 @@ class ScalableReader(_StatefulDataset):
         shard_states = state_dict[self.statename("shard_states")]
         file_info = state_dict[self.statename("file_info")]
         print(shard_states.size(0), self.worldsize)
+        if self.rank == 0:
+            print(shard_states)
         if shard_states.size(0) == self.worldsize:
             self.filesizes = file_info
             self.shard_states = shard_states[self.rank]
--- a/torchdata/stateful_dataloader/ibm_rescalable.py
+++ b/torchdata/stateful_dataloader/ibm_rescalable.py
@@ -489,6 +489,7 @@ class ScalableReader(_StatefulDataset):
         # Load back shard states (global), filesizes (all)
         shard_states = state_dict[self.statename("shard_states")]
         file_info = state_dict[self.statename("file_info")]
+        print(shard_states.size(0), self.worldsize)
         if shard_states.size(0) == self.worldsize:
             self.filesizes = file_info
             self.shard_states = shard_states[self.rank]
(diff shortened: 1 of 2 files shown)


torchdata/stateful_dataloader/ibm_rescalable.py
```diff
@@ -0,0 +1,614 @@
...
+then applying some sort of post-processing and yielding the result. Users build data processing 
+pipelines by wrapping a base `_StatefulDataset` in any number of `_WrapperDataset` layers, 
+which is then passed to the torch DataLoader. 
+
+It is likely that this can be merged into the existing Nodes structure, but we leave this for
+future work, for now.
+"""
+
+
+def _shard_partition(itemlist: List[Any], rank: int, worldsize: int) -> List[Any]:
```
andrewkho, line 44: Are tail elements just truncated?

torchdata/stateful_dataloader/ibm_rescalable.py
```diff
@@ -0,0 +1,614 @@
...
+        # Default fields
+        self.datapath = datapath
+        self.rank = rank
+        self.worldsize = worldsize
+        self.local_worldsize = -1
+
+        # Setup / loading flags
+        self.is_setup = False
+
+    def setup(self):
```
andrewkho, line 79: This can be mapped pretty easily to BaseNode.reset()

torchdata/stateful_dataloader/ibm_rescalable.py
```diff
@@ -0,0 +1,614 @@
...
+
+    def load_state_dict(self, state_dict):
+        """
+        Run setup if needed, and apply all applicable state_params from the state_dict.
+        """
+        self.setup()
+        [setattr(self, flag, state_dict[self.statename(flag)]) for flag in self.state_params]
+
+
+class _WrapperDataset(_StatefulDataset):
```
andrewkho, line 123: thinking out loud: could we do this with mixins instead of extending the type hierarchy?

andrewkho: Actually, what's the benefit of having two subclasses?

torchdata/stateful_dataloader/ibm_rescalable.py
```diff
@@ -0,0 +1,614 @@
...
+            [d.setup() for d in self.data]
+
+    def __iter__(self):
+        self.setup()
+        # Grab one item at a time, iterating over owned logical shards
+        data = [iter(d) for d in self.data]
+        while True:
+            ind = self.current_reader
+            # Read doc
+            out = next(data[ind])
```
andrewkho, line 469: How is StopIteration handled?

torchdata/stateful_dataloader/ibm_rescalable.py
```diff
@@ -0,0 +1,614 @@
...
+        state_dict.update(_StatefulDataset.state_dict(self))
+
+        # Convert to tensor form
+        out = {}
+        for k, v in state_dict.items():
+            v = torch.tensor(v)
+            if len(v.shape) == 0:
+                k = k + ".scalar"
+                v = v.unsqueeze(0)
+            out[k] = v
```
andrewkho, line 490: Is this done to satisfy DCP requirements?

torchdata/stateful_dataloader/ibm_rescalable.py
```diff
@@ -0,0 +1,614 @@
...
+
+        # Load values
+        for i in range(self.n_logicals):
+            self.data[i].load_state_dict(logical_shard_states[i])
+
+
+#### -------------------------    CHECKPOINT FUNCTIONS    ------------------------- ####
+
+
+def __pop_dstate(state, device_mesh, placements):
```
andrewkho, line 527: We should create standard utilities to get these in torchdata  https://github.com/pytorch/data/issues/1337

torchdata/stateful_dataloader/ibm_rescalable.py
```diff
@@ -0,0 +1,614 @@
...
+        data = [iter(d) for d in self.data]
+        while True:
+            ind = self.current_reader
+            # Read doc
+            out = next(data[ind])
+            # Update state
+            self.current_reader = (self.current_reader + 1) % self.n_logicals
+            yield out
+
+    def state_dict(self):
```
andrewkho, line 474:
```
{
  my_children: [c.state_dict() for c in self.children],
  scalar_state: self.scalar, # "my_string"
  my_reshardale_state: tensor.array([1, 2, 3, 4, 5]), # 2d tensor
}

question: what happens if above state_dict gets passed to DCP? 
Answer: it will fail because torch.tensor gets called on everything?

Andrew to follow up with @pradeepfn on this
```

torchdata/stateful_dataloader/ibm_rescalable.py
```diff
@@ -0,0 +1,614 @@
...
+        self.setup()
+        # Recursive fetch
+        logical_shard_states = [d.state_dict() for d in self.data]
+        assert len(logical_shard_states) > 0, f"Worker {self.rank} owns no shards???"
+        # Flip list[dict[Any]] to dict[list[Any]]
+        state_dict = {k: [d[k] for d in logical_shard_states] for k in logical_shard_states[0].keys()}
+        state_dict.update(_StatefulDataset.state_dict(self))
```
andrewkho, line 481: Does `self.current_reader` need to be stored too?

torchdata/stateful_dataloader/ibm_rescalable.py
```diff
@@ -0,0 +1,614 @@
...
+        self.g_state = self.generator.get_state().tolist()
+        return super().state_dict()
+
+    def load_state_dict(self, state_dict):
+        super().load_state_dict(state_dict)
+        # Manually set generator state
+        self.generator.set_state(torch.tensor(self.g_state, dtype=torch.uint8))
+
+
+class ScalableShardDataset(_WrapperDataset):
```
andrewkho, line 392:
```suggestion
Protocol based instead 

class ScalableShardDataset(BaseNode[T], Reshardable):
```

torchdata/stateful_dataloader/ibm_rescalable.py
andrewkho, line 469: Can we bound the number of open iterators/filepointers/etc in some way here while still maintaining re-shardability

andrewkho: Possibly run into end-of-epoch problem

andrewkho: can we remove the assumption of indexable files

torchdata/stateful_dataloader/ibm_rescalable.py
```diff
@@ -0,0 +1,614 @@
...
+    state = deepcopy(loader.state_dict())
+    dstate = __pop_dstate(state, device_mesh, [dtensor.placement_types.Shard(0)])
+    # Write distributed state dict
+    writer = checkpoint.FileSystemWriter(path)
+    checkpoint.save(
+        dstate,
+        writer,
+    )
+    # Write nondistributed state dict
+    torch.save(state, os.path.join(path, f"__nondist_cp_{rank}.pth"))
```
pradeepfn, line 570: Shall we also make this state part of the main checkpoint? We can use the torch.save serialization (output is bytestream), to store the state as part of the DCP checkpoint.

buff = io.BytesIO()
torch.save(state, buff) # or we can serialize individual keys in the state-dict. But no strong need.
buff.seek(0)

assume the unique key is 'trainer_dataloader_state_rank_k' -> "

update the dstate with new key -> value.
checkpoint.save(dstate)

torchdata/stateful_dataloader/ibm_rescalable.py
```diff
@@ -0,0 +1,614 @@
...
+    """
+    base = loader.state_dict()
+    nworkers = base["_snapshot"]["_main_snapshot"]["_num_workers"]
+    rank = loader.dataset.rank
+    dstate = __pop_dstate(base, device_mesh, [dtensor.placement_types.Shard(0)])  # placements)
+    # Read nondistributed state dict
+    ckp_ws = 0 if not os.path.exists(path) else len([x for x in os.listdir(path) if "__nondist_cp_" in x])
+    # Check that number of loaders matches
+    if ckp_ws == loader.dataset.worldsize:
+        state = torch.load(os.path.join(path, f"__nondist_cp_{rank}.pth"))
```
pradeepfn, line 593: noob q: what are we missing out, if we just set the;

data_loader_state= base

without considering the rescaling property of the training run ?

torchdata/stateful_dataloader/ibm_rescalable.py
pradeepfn: Some context on how DCP handles regular ( non-dtensor) tensor and non-tensor values found in state dict.

1\ regular tensors ( not distributed tensors) are saved against a unique fqn. The state dict can have { 'foo.bar.1' : torch.ones((2,3), dtype=float). As long as the fqns do not collide between different ranks we are good.

2\ DCP can save/load non-tensor values found in state dict as well.  That means the value can be, a string, list of ints or a ByteIO object. Under the hood DCP uses torch.save/torch.load to serialize/deserialize these non-tensor (we call them blobs) values during checkpoint save/load. E.g. UT -> https://github.com/pytorch/pytorch/blob/11bb94b7eaf272da8e1f1dfd94c3d8872247e895/test/distributed/checkpoint/test_file_system_checkpoint_cpu.py#L367


Therefore, we  can use the dataloader state-dict values as it is, to save them as part of the DCP checkpoint.

torchdata/stateful_dataloader/ibm_rescalable.py
daviswer: No, this code will distribute extra elements as evenly as possible, even if it's not perfect. Technically nothing breaks if you load into a worldsize that doesn't divide logical_shards evenly, you just end up with some shards progressing faster than others (since some devices now have an extra logical shard)

torchdata/stateful_dataloader/ibm_rescalable.py
daviswer: Yeah I thought so too!

torchdata/stateful_dataloader/ibm_rescalable.py
daviswer: The base reader is a _StatefulDataset but not a _WrapperDataset, so the distinction is meaningful, but yeah the only reason it's not mixins is because of my lack of familiarity with building mixins!

torchdata/stateful_dataloader/ibm_rescalable.py
daviswer: It's not, in this framework we just assume each iterator loops forever. Converting to a `next()` based framework would make this pretty easy to handle though.

daviswer: We need indexable files in order to support fractional file ownership between logical shards. I don't see a way around that one if we want to maintain rough epoch boundaries. We can bound the number of open files though. We'd discussed how the reshard layer makes sense to either go at the very top of the pipeline (as in here), or at the very bottom just above the base reader (as in the older version). If we move the reshard layer to the very bottom - and maybe merge it with the base reader(s) - we can allow base readers to share file pointers. Then instead of partitioning all data across shards based on file size, we can instead split every data file over all the workers. That way all logical shards on the same device will share a single file pointer, and assuming they all finish at roughly the same time, we'll end up maintaining no more than two open files per dataset per physical worker (instead of one open file per dataset per logical worker, as here). Total pulls per file goes up if we do this, but I think this is still a good tradeoff.

torchdata/stateful_dataloader/ibm_rescalable.py
```diff
@@ -0,0 +1,614 @@
...
+            self.current_reader = (self.current_reader + 1) % self.n_logicals
+            yield out
+
+    def state_dict(self):
+        self.setup()
+        # Recursive fetch
+        logical_shard_states = [d.state_dict() for d in self.data]
+        assert len(logical_shard_states) > 0, f"Worker {self.rank} owns no shards???"
+        # Flip list[dict[Any]] to dict[list[Any]]
+        state_dict = {k: [d[k] for d in logical_shard_states] for k in logical_shard_states[0].keys()}
```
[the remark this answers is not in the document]
daviswer: Yes exactly!

torchdata/stateful_dataloader/ibm_rescalable.py
daviswer: Yes, `self.current_reader` is stored here so that when we don't rescale, we'll maintain the same data ordering

torchdata/stateful_dataloader/ibm_rescalable.py
daviswer: Yes, dtensors have to have at least one dimension

torchdata/stateful_dataloader/ibm_rescalable.py
daviswer: Oh good to know! How are non-dtensor values like these handled during rescaling? I guess if all fqns are unique, they can all be broadcast over all ranks?

torchdata/stateful_dataloader/ibm_rescalable.py
daviswer: You mean if we always set `state=base`, and just drop the non-reshardable state every time? Nothing breaks, we just lose the ability to maintain the original data ordering when we're _not_ rescaling.

Issue #1374, opened 2024-11-25
[nodes] ParallelMap should fall back to Mapper when num_workers = 0


andrewkho:
### 🚀 The feature

right now users need to jump through hoops to configure inline mapping vs parallel mapping. We should offer a way to make this seamless with a config setting, instead of forcing them to call different classes

### Motivation, pitch

ParallelMapper(num_workers=0) should behave identically to Mapper. Alternatively, we have a builder that returns the correct node

PR #1375, opened 2024-11-26
Consolidate Mapper and ParallelMapper

andrewkho:
It's a pain to switch between inline vs parallel Mapping. This change enables users to switch seamlessly by creating a separate "Inline" iterator. Updates "Mapper" to be a builder function which returns ParallelMapper(..., num_workers=0)

Test Plan: Existing unit tests since this should be fully backwards compatible with no behaviour changes.

#1374
-


divyanshk (approved):
LGTM

PR #1377, opened 2024-11-26
initial state fix for pin_memory and SingleThreadedMapper

andrewkho:
Fixes some state-management edge cases.

1) calling state_dict before a pre-fetcher was not capturing a snapshot correctly. We now send a snapshot (or initialization error) as an "ack" during _populate_queue and _pin_memory startups to handle this scenario correctly. 

2) calling `loader.state_dict(); iter(state_dict)` would call `reset()` twice on the underlying nodes (once for initial state_dict), whereas running `iter(state_dict); loader.state_dict()` would only call once since `loader._it` is already instantiated. Added a flag to handle these cases.

Test plan:
Create `class StatefulRangeNode(BaseNode)` which tracks number of resets, and use these in `test_map, test_prefetch, test_pin_memory` state management tests. Those tests failed before the changes and succeed now.

### DIFF
--- a/torchdata/nodes/loader.py
+++ b/torchdata/nodes/loader.py
@@ -10,10 +10,23 @@ class Loader(Generic[T]):
         self.restart_on_stop_iteration = restart_on_stop_iteration
         self._next_iter_state_dict: Optional[Dict[str, Any]] = None
         self._it: Optional[LoaderIterator[T]] = None
+        # Tracks whether an iterator was created solely for getting a state_dict, in which case
+        # we don't want to reset the iterator. Consider these two cases, which should behave the same
+        # it = iter(loader)
+        # sd = loader.state_dict()  # No extra __iter__ call as _it already exists
+        # for _ in it: ...
+        # --------
+        # sd = loader.state_dict()  # Calls __iter__ since _it is None
+        # it = iter(loader)  # We don't want to reset the iterator here again
+        # for _ in it: ...
+        self._iter_for_state_dict: bool = False
 
     def __iter__(self):
         if self._it is None:
             self._it = LoaderIterator(self)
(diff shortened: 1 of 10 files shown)


torchdata/nodes/map.py
```diff
@@ -192,6 +192,12 @@ def __init__(
         if self.in_order:
             self._sort_thread.start()
 
+        # Try and get initial snapshot
+        while self._snapshot is None:
+            self._maybe_update_snapshot(-1)
```
divyanshk, line 197: Is it possible in any scenario _maybe_update_snapshot fails to update self._snapshot and we end up looping?

torchdata/nodes/loader.py
```diff
@@ -10,10 +10,23 @@ def __init__(self, root: BaseNode[T], restart_on_stop_iteration: bool = True):
         self.restart_on_stop_iteration = restart_on_stop_iteration
         self._next_iter_state_dict: Optional[Dict[str, Any]] = None
         self._it: Optional[LoaderIterator[T]] = None
+        # Tracks whether an iterator was created solely for getting a state_dict, in which case
```
divyanshk, line 13: Can confirm that with this change to the loader.py we do not need to tweak the test here https://github.com/pytorch/data/pull/1369/files#diff-b4616338bda203d1cf7ac17ae95afaa662502e89f5b8b9bb6c3e75e7bb6ad73aR125-R127 :-)

torchdata/nodes/map.py
andrewkho: Yes good call out, eg if the worker function hits a segfault. we can add a check here

divyanshk (approved):
Minor question, otherwise LGTM

PR #1380, opened 2024-12-03
Fix test flakiness and test for it

andrewkho:
Run each potentially flaky test 10 times, raise in worker instead of exiting to more reliably kill thread

### DIFF
--- a/test/nodes/test_snapshot_store.py
+++ b/test/nodes/test_snapshot_store.py
@@ -13,7 +13,7 @@ from torchdata.nodes.exception_wrapper import StartupExceptionWrapper
 from torchdata.nodes.snapshot_store import QueueSnapshotStore
 
 
-class TestDequeSnapshotStore(TestCase):
+class TestQueueSnapshotStore(TestCase):
     def test_snapshot_store(self) -> None:
         for _ in range(100):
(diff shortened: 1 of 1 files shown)


test/nodes/test_snapshot_store.py
```diff
@@ -55,38 +55,45 @@ def test_snapshot_store(self) -> None:
...
-        thread.start()
-        with self.assertRaisesRegex(RuntimeError, r"thread.is_alive\(\)=False"):
-            store.get_initial_snapshot(thread, QUEUE_TIMEOUT * 4.0)
-        thread.join()
+        for _ in range(10):  # Should be reliable
+            store = QueueSnapshotStore()
+            thread = threading.Thread(target=_worker_raises_after, args=(QUEUE_TIMEOUT * 3.0,))
+            thread.start()
+            with self.assertRaisesRegex(RuntimeError, r"thread.is_alive\(\)=False"):
+                store.get_initial_snapshot(thread, QUEUE_TIMEOUT * 5.0)
```
divyanshk, line 84: n00b q, why do we have different multipliers to queue_timeout here? 3 vs 5

ramanishsingh (approved):

PR #1390, opened 2024-12-09
add new torchdata.nodes doc file

divyanshk:
Clone the repo on local, created the environment, updated the `/docs` folder, then`make html` and `sphinx-serve -b build`.

Screenshot 2024-12-09 at 3 46 59 PM

Screenshot 2024-12-09 at 3 47 24 PM

PR #1394, opened 2024-12-10
Update symlink and add new list item for v0.10 in gh-pages branch

divyanshk:
This step is in preparation for torchdata v0.10 release. 

This will be landed after the binaries are pushed. Opening this PR for review. 

I have a quick question for @svekars. I have added new pages to main (https://github.com/pytorch/data/pull/1390) as well as release/0.10 branch (https://github.com/pytorch/data/pull/1392), but when looking deeper into the `gh-pages` branch I do not see them reflected immediately. Is there a way to trigger the change ?

We have our docs build process as part of the test_release workflow (config: https://github.com/pytorch/data/blob/main/.github/workflows/_build_test_upload.yml#L211) but we might want to avoid running this flow just to update the gh-pages. We will take an action item to clean up this process, but wondering if there is a way to trigger that changes for both the branches. I can see this workflow taking care of syncing (https://github.com/pytorch/data/actions/workflows/pages/pages-build-deployment) but not sure where it gets triggered from exactly.


andrewkho (approved):

svekars (approved):
LGTM

PR #1398, opened 2024-12-11
update smoke test

divyanshk:
1) update smoke test to include an import from torchdata.nodes
2) minor doc string fix


torchdata/nodes/samplers/multi_node_weighted_sampler.py
```diff
@@ -207,7 +207,7 @@ class _WeightedSampler:
     - g_rank_state: The state of the random number generator for the rank.
     - offset: The offset of the batch of indices.
 
-    Parameters:
```
andrewkho, line 210: Did I miss this in the docs update?

PR #1401, opened 2024-12-12
Andrewkh/pin test infra

andrewkho:
build_wheels_linux needs to be pinned to 2.5 otherwise it tries to pip install 2.6.0 which isn't released yet


andrewkho:
wrong base

Issue #1410, opened 2024-12-18
Fix formatting on pytorch docs


divyanshk:
### 📚 The doc issue

1) Fix unformatted `:return:` in https://pytorch.org/data/main/torchdata.nodes.html
2) Fix unformatted bullets points in https://pytorch.org/data/main/torchdata.nodes.html#torchdata.nodes.MultiNodeWeightedSampler

PR #1412, opened 2024-12-20
Fix docs formatting

divyanshk:
Fixes https://github.com/pytorch/data/issues/1410

Confirmed by running docs on local machine there bullets work fine, and Returns format correctly.
Screenshot 2024-12-20 at 12 20 46 PM
Screenshot 2024-12-20 at 12 29 42 PM


andrewkho (approved):
LGTM, although I think this works too and might be more consistent with the way Args are defined
```
Args:
  x
  y
  z

Returns:
  bool - blah blah blah
```

divyanshk:
Yup, now args and returns are formatted the same way.

Issue #1415, opened 2024-12-26
[Nodes] Add pre-batch option to parallel mapper


andrewkho:
### 🚀 The feature

Add a pre-batch option to ParallelMapper to improve performance

### Motivation, pitch

Throughput can often be improved by doing pre-batching of inputs, lowering queue overheads. Add an option (default off?) to pre-batch for improved throughput. Maybe we could even tune-this dynamically by measuring queue overhead? Need to be careful about memory overhead so maybe best to leave this off by default

PR #1416, opened 2024-12-26
Add unbatcher node

andrewkho:
Required for #1415 

### Changes
* Add Unbatcher node and unit test
-
-

### DIFF
--- a/torchdata/nodes/batch.py
+++ b/torchdata/nodes/batch.py
@@ -4,7 +4,7 @@
 # This source code is licensed under the BSD-style license found in the
 # LICENSE file in the root directory of this source tree.
 
-from typing import Any, Dict, List, Optional
+from typing import Any, Dict, List, Optional, Sequence
 
 from torchdata.nodes.base_node import BaseNode, T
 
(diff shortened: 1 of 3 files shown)


torchdata/nodes/batch.py
```diff
@@ -56,3 +56,56 @@ def next(self) -> List[T]:
...
+                self._batch = []
+                self._batch_idx = 0
+        else:
+            self.source.reset()
+            self._batch = []
+            self._cached_state_dict = None
+            self._batch_idx = 0
+
+    def next(self) -> T:
+        while self._batch_idx >= len(self._batch):
```
divyanshk, line 96: This should be `if` instead of `while` ? i.e. if the _batch_idx overshoots the current _batch, get a new _batch and reset _batch_idx to 0.

andrewkho:
Test failure seems unrelated, not sure what caused this but it seems weird to rely on a particular implementation of the RNG

torchdata/nodes/batch.py
andrewkho: @divyanshk yes i was worried about next batch of size 0 case, although I'm assuming that's unlikely but it's an edge case none the less

divyanshk (approved):
LGTM!

PR #1417, opened 2024-12-26
[Nodes] Add Prebatch setting to ParallelMapper

### LINKED ISSUE #1415: [Nodes] Add pre-batch option to parallel mapper

andrewkho:
When ParallelMapper is used for very cheap operations, the overhead of sending items over queues can quickly add up. This is a nice parameter to be able to tune.

Fixes #1415 

A few notes about the implementation:
* I chose to compose 3 nodes (Batcher, ParallelMapper, Unbatcher) into one to implement this.  This is the first time we're composing BaseNodes with other BaseNodes. This will require us to figure out graph-traversal for these options (see footnote). 
* this required us to have `_ParallelMapperIter` implement BaseNode, however getting reset to work correctly is going to be a bigger problem, so for now, just created an intermediate class with basically the current implementation of ParallelMapper, and this allows us to use torchdata.nodes composition to get things working easily.

Test Plan:
* Unit tests
* Ran a simple script to test this, output:
```
python examples/nodes/test_prebatch.py
[9999400009, 9999600004, 9999800001]
baseline: dt=3.0651697060093284s
[9999400009, 9999600004, 9999800001]
prebatch=16: dt=0.454918147996068s
[9999400009, 9999600004, 9999800001]
prebatch=256: dt=0.13740589004009962s
[9999400009, 9999600004, 9999800001]
prebatch=1024: dt=0.22711888700723648s
```

test script:
```
import time
import torchdata.nodes as tn


def run(prebatch):
    node = tn.IterableWrapper(range(100000))
    node = tn.ParallelMapper(node, map_fn=lambda x: x**2, prebatch=prebatch, method="thread", num_workers=8)
    loader = tn.Loader(node)
    x = list(loader)
    print(x[-3:])


if __name__ == "__main__":
    t0 = time.perf_counter()
    run(None)
    dt = time.perf_counter() - t0
    print(f"baseline: {dt=}s")

    for prebatch in (16, 256, 1024):
        t0 = time.perf_counter()
        run(prebatch)
        dt = time.perf_counter() - t0
        print(f"{prebatch=}: {dt=}s")
```

Footnote: Example of where this is a problem: In the ParallelMapper case here, traversing the dag with reflection (eg using instance.__dict__ and checking for baseNode instances) would generate two sinks for the source, since self.source points to it, and self._it would eventally point to it as well. One way we could handle this is with an optional "get_source/get_parent" method on BaseNode, which returns the instance of where graph traversal should begin, and in this case it would return self._it, not self.source.

### DIFF
--- a/torchdata/nodes/map.py
+++ b/torchdata/nodes/map.py
@@ -7,7 +7,7 @@
 import queue
 import threading
 import time
-from typing import Any, Callable, Dict, Iterator, List, Literal, Optional, Protocol, Sequence, TypeVar, Union
+from typing import Any, Callable, Dict, Generic, Iterator, List, Literal, Optional, Protocol, Sequence, TypeVar, Union
 
 import torch.multiprocessing as mp
 from torchdata.nodes.base_node import BaseNode, T
@@ -53,7 +53,11 @@ def Mapper(source: BaseNode[X], map_fn: Callable[[X], T]) -> "ParallelMapper[T]"
     )
 
 
-class MapOverBatch(Callable[[Sequence[X]], T]):
+Xseq = Sequence[X]
+Tseq = Sequence[T]
+
+
+class MapOverBatch(Generic[X, T]):
     def __init__(self, map_fn: Callable[[X], T]):
         self.map_fn = map_fn
 
@@ -346,7 +350,7 @@ class _ParallelMapperImpl(BaseNode[T]):
             initial_state=initial_state,
         )
 
-    def next(self):
+    def next(self) -> T:
         return next(self._it)  # type: ignore[arg-type, union-attr]
 
     def get_state(self) -> Dict[str, Any]:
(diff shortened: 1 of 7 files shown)


torchdata/nodes/map.py
```diff
@@ -272,6 +281,77 @@ def _shutdown(self):
                     t.join(timeout=QUEUE_TIMEOUT * 5)
 
 
+class _ParallelMapperImpl(BaseNode[T]):
+    """This class implements _ParallelMapperIter as a BaseNode, allowing it
```
divyanshk, line 285: Nit: This class implements _ParallelMapperIter and _InlineMapperIter as a BaseNode, ....

andrewkho: good catch!

ramanishsingh (approved):

PR #1424, opened 2025-01-20
Make Multi Node sampler cycle forever

divyanshk:
This PR adds support for continuous cycling through each base node in multi node weighted sampler. With the current setup this was each to extend. We do cycle through each node to support `StopCriteria.CYCLE_UNTIL_ALL_DATASETS_EXHAUSTED`, for `StopCriteria.CYCLE_FOREVER` we skip raising a `StopIteration`.

This should be functionally equivalent to having an infinite sampler plugged on top of each base dataset. 

The dataset order (as defined by `weights`) remains the same.

### DIFF
--- a/torchdata/nodes/samplers/stop_criteria.py
+++ b/torchdata/nodes/samplers/stop_criteria.py
@@ -17,8 +17,12 @@ class StopCriteria:
        dataset is seen exactly once. No wraparound or restart will be performed.
 
     3) FIRST_DATASET_EXHAUSTED: Stop when the first dataset is exhausted.
+
+    4) CYCLE_FOREVER: Cycle through the datasets by reinitializing each exhausted source nodes.
+       This is useful when trainer want control over certain number of steps instead of epochs.
     """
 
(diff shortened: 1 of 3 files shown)


ramanishsingh (approved):
LGTM!

torchdata/nodes/samplers/stop_criteria.py
```diff
@@ -17,8 +17,12 @@ class StopCriteria:
        dataset is seen exactly once. No wraparound or restart will be performed.
 
     3) FIRST_DATASET_EXHAUSTED: Stop when the first dataset is exhausted.
+
+    4) CYCLE_FOREVER: Cycle through the datasets by reinitializing each exhausted source nodes.
+       This is useful when trainer want control over certain number of steps instead of epochs.
```
andrewkho, line 22: Will this call set_epoch on the underlying datasets?

andrewkho: Wondering if we need to do anything special, or can advise users on how to perform that. This otherwise looks good to me

PR #1426, opened 2025-01-22
Fix MultiNodeWeightedSampler tests

ramanishsingh:
Due to the stochastic nature of the MultiNodeWeightedSampler and the way StopIteration is called when multiple datasets are involved, some test results are dependent on the OS/seed. Trying to fix seed and change testing conditions with this PR.

### DIFF
--- a/test/nodes/test_multi_node_weighted_sampler.py
+++ b/test/nodes/test_multi_node_weighted_sampler.py
@@ -24,7 +24,7 @@ class TestMultiNodeWeightedSampler(TestCase):
         self._num_datasets = 4
         self._weights_fn = lambda i: 0.1 * (i + 1)
         self._num_epochs = 3
-
+        self._seed = 42
         self.datasets = {
             f"ds{i}": IterableWrapper(DummyIterableDataset(self._num_samples, f"ds{i}"))
             for i in range(self._num_datasets)
@@ -38,11 +38,11 @@ class TestMultiNodeWeightedSampler(TestCase):
         except ImportError:
             self.fail("MultiNodeWeightedSampler or StopCriteria failed to import")
 
-    def _setup_multi_node_weighted_sampler(self, num_samples, num_datasets, weights_fn, stop_criteria) -> Prefetcher:
+    def _setup_multi_node_weighted_sampler(self, num_samples, num_datasets, weights_fn, stop_criteria, seed) -> Prefetcher:
 
         datasets = {f"ds{i}": IterableWrapper(DummyIterableDataset(num_samples, f"ds{i}")) for i in range(num_datasets)}
         weights = {f"ds{i}": weights_fn(i) for i in range(num_datasets)}
-        node = MultiNodeWeightedSampler(datasets, weights, stop_criteria)
+        node = MultiNodeWeightedSampler(datasets, weights, stop_criteria, seed=seed)
         return Prefetcher(node, prefetch_factor=3)
 
     def test_multi_node_weighted_sampler_weight_sampler_keys_mismatch(self) -> None:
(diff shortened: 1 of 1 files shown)


test/nodes/test_multi_node_weighted_sampler.py
```diff
@@ -24,7 +24,7 @@ def setUp(self) -> None:
         self._num_datasets = 4
         self._weights_fn = lambda i: 0.1 * (i + 1)
         self._num_epochs = 3
-
+        self._seed = 42
```
divyanshk, line 27: Should we make the test loop over multiple seeds?

test/nodes/test_multi_node_weighted_sampler.py
```diff
@@ -82,20 +81,21 @@ def test_multi_node_weighted_batch_sampler_zero_weights(
...
         mixer = self._setup_multi_node_weighted_sampler(
             self._num_samples,
             self._num_datasets,
             self._weights_fn,
             stop_criteria=StopCriteria.FIRST_DATASET_EXHAUSTED,
+            seed=seed,
         )
 
-        for _ in range(self._num_epochs):
+        for _ in range(3):
```
divyanshk, line 98: Should be `self._num_epochs`

test/nodes/test_multi_node_weighted_sampler.py
```diff
@@ -104,17 +104,19 @@ def test_multi_node_weighted_sampler_first_exhausted(self) -> None:
             # Check max item count for dataset is exactly _num_samples
             self.assertEqual(max(dataset_counts_in_results), self._num_samples)
 
-            # Check only one dataset has been exhausted
-            self.assertEqual(dataset_counts_in_results.count(self._num_samples), 1)
+            # Check that max items are taken from at least one dataset
+            self.assertGreaterEqual(dataset_counts_in_results.count(self._num_samples), 1)
```
divyanshk, line 108: We can also add a check that no dataset has count more than `self._num_samples`.

ramanishsingh: I think we do not need that check because we have `self.assertEqual(max(dataset_counts_in_results), self._num_samples)` in the prevous line
so that ensures no dataset item count can be greater than self._num_samples

test/nodes/test_multi_node_weighted_sampler.py
ramanishsingh: Fixed!

test/nodes/test_multi_node_weighted_sampler.py
divyanshk: Ah right!

test/nodes/test_multi_node_weighted_sampler.py
divyanshk: Synced offline, we can probably remove the _seed and add a range along with parameterized.expand to test multiple seeds for relevant unit tests.

test/nodes/test_multi_node_weighted_sampler.py
```diff
@@ -86,13 +87,15 @@ def test_multi_node_weighted_batch_sampler_zero_weights(
                 weights={f"ds{i}": 10 * i for i in range(self._num_datasets)},
             )
 
-    def test_multi_node_weighted_sampler_first_exhausted(self) -> None:
+    @parameterized.expand(range(10))
+    def test_multi_node_weighted_sampler_first_exhausted(self, seed) -> None:
         """Test MultiNodeWeightedSampler with stop criteria FIRST_DATASET_EXHAUSTED"""
         mixer = self._setup_multi_node_weighted_sampler(
```
divyanshk, line 93: one tiny nit: can you make this `node = self._setup_multi_node_weighted_sampler(` here as well as in other places we call _setup_multi_node_weighted_sampler, thanks!

divyanshk (approved):
LGTM! Thanks for the fix!

PR #1423, opened 2025-01-20
[Stateful DL] Add out of order implementation

michael-diggin:
Fixes #1414

### Changes

- Adds the changes from https://github.com/pytorch/pytorch/pull/141833 and https://github.com/pytorch/pytorch/pull/142324 to the Stateful DL
- Adds a warning log to mention that state management may not work as expected.

The tests I've added were the exact same ones from the main PyTorch repo, so the functionality of the flag is the exact same. There was a few changes required to be compatible with the few slight differences of the `_StatefulMultiProcessingDataLoaderIter`.

I have also tested some simple cases of stopping/resuming both an index and iterable ds and it seems to work correctly (the same types that are used in the new test cases). I haven't added extra tests for that into this PR just to keep the size manageable, but would add them in a follow up PR as well as checking extra edge cases.

PR #1428, opened 2025-01-28
Add tests for out of order with checkpointing

michael-diggin:
Follow up to #1423

### Changes

- Add tests for state management of both index datasets and iterable datasets when `in_order=False`
- Remove warning logs about `in_order=False`

State management with `in_order=False` actually _just works_ which is really nice.

I'm happy to keep the warning logs in for longer if we'd like to err on the side of caution. And please let me know if there are any other test cases I should add.

### DIFF
--- a/test/stateful_dataloader/test_state_dict.py
+++ b/test/stateful_dataloader/test_state_dict.py
@@ -6,6 +6,8 @@
 
 import itertools
 import json
+import math
+import time
 import unittest
 from copy import deepcopy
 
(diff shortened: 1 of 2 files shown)


andrewkho:
How can this work for out-of-order iteration? eg let's say the first sample blocks until the end of training, but you take a checkpoint in the middle, what happens then? Without replaying all of history, how will sample 0 be returned later?

michael-diggin:
> How can this work for out-of-order iteration? eg let's say the first sample blocks until the end of the epoch, but you take a checkpoint in the middle, what happens then? Without replaying all of history, how will sample 0 be returned later?

That's exactly what the tests I've added are checking - the first sample for one of the workers is slow.
For index datasets it looks like it uses the fact that it has progressed a certain number of steps since the last checkpoint (which  is part of the state dict) to skip forward, and for the iterable dataset it looks to be using the `num_yielded` field from the the state dict.
If this isn't currently expected to work I can add back in the warning logs while we fully determine if it should work as is?

andrewkho:
@michael-diggin let's leave the warnings in for now just in case, while we think through this a bit more. Thanks again!

michael-diggin:
> @michael-diggin let's leave the warnings in for now just in case, while we think through this a bit more. Thanks again!

@andrewkho no problem - added back in just now.

test/stateful_dataloader/test_state_dict.py
```diff
@@ -1632,5 +1634,100 @@ def test_mp(self):
...
+        new_dataloader.load_state_dict(state_dict)
+        for i, data in enumerate(new_dataloader):
+            output.append(data)
+
+        self.assertEqual(len(output), 10)
+        self.assertNotEqual(output, list(range(10)))
+        self.assertEqual(sorted(output), list(range(10)))
+
+    def test_out_of_order_iterable_ds(self):
+        dataset = _TestSlowIterableDataset(start=0, end=10)
```
andrewkho, line 1701: For iterable dataset the slow worker will just lead to a straggler, on resume the individual workers will resume and continue, though be limited to single-worker performance. I think I can see why this might "just work" for Iterable datasets

test/stateful_dataloader/test_state_dict.py
```diff
@@ -1632,5 +1634,100 @@ def test_mp(self):
...
+            num_workers=2,
+            in_order=False,
+        )
+
+        # worker_id = 0 gets 'stuck' on 0 and also has 2 in it's queue
+        # due to prefetch_factor being 2
+        output = []
+        for i, data in enumerate(dataloader):
+            output.append(data)
+            if i == 5:
```
andrewkho, line 1687: Can we try to set this lower than 5? I think for end = 10, slow_index=0, and num_workers = 2, i = 5 will be index 0, since 0 through 4 will come from worker 1, and worker 0 will be blocked until index 0 is released

test/stateful_dataloader/test_state_dict.py
michael-diggin: I'm not entirely sure - I've added a second case that breaks before either worker finishes, and so they both resume after the restart, which gives the same correct results.
I think maybe the fast-forwarding part of the resuming is what is allowing this to work, and since the dataset is deterministic (ie the slow samples don't change) the fast forwarding by X samples will bring it back to the same point.

test/stateful_dataloader/test_state_dict.py
michael-diggin: I'm not sure that was true - since we added the re-distribution of work. I checked this test's output and 0 was retuned at index 8, so after the break point.
I've dropped the break to 3 down however, and also added an assertion to check that 0 isn't in the output at that point as well.

andrewkho (approved):
Let's land this once CI finishes and figure out details for the next release

PR #1429, opened 2025-01-29
Update build docs 

ramanishsingh:
-
-

### DIFF
--- a/.github/workflows/_build_test_upload.yml
+++ b/.github/workflows/_build_test_upload.yml
@@ -228,7 +228,7 @@ jobs:
       - name: Install Dependencies
         run: |
           echo `python3 --version`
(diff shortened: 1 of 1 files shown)


.github/workflows/_build_test_upload.yml
```diff
@@ -228,7 +228,7 @@ jobs:
       - name: Install Dependencies
         run: |
           echo `python3 --version`
-          sudo apt-get install -y python-setuptools
```
kit1980, line 231: Just delete the whole line? You install setuptools later with pip anyway,

svekars: trying here: https://github.com/pytorch/data/pull/1430

PR #1432, opened 2025-01-30
update python version CI

ramanishsingh:
Updating CI version


divyanshk:
looks like the support for 3.13t isn't really there at the moment

PR #1439, opened 2025-02-03
Fix end of epoch StatefulDataLoader restart

### LINKED ISSUE #1437: loading a StatefulDataLoader state_dict creates a slightly different one in the dataloader, potentially dropping an epoch
### 🐛 Describe the bug

Loading a state_dict taken from a StatefulDataLoader that has just completed an epoch yields what I would describe as a "just finishing" rather than "just finished" state: the next iteration over that dataloader does nothing (as opposed to a full new epoch), before continuing as expected on the following one.

Reproduction: I run a dataloader for two epochs, save the state dict, remake it, load the state dict, and try to run for two more epochs. I find only 3 epochs have run. They *do* align with the first 3 epochs of a separate 4-epoch run, but, they are 3 instead of 4.

(Side comment - the fact that they align here despite me not setting the random seed reveals that the shuffling of the StatefulDataLoader ignores the current state of the random number generators - this behaviour could maybe be clearer in the documentation, as it is not equivalent to torch.utils.data.DataLoader in this regard)

```
from torchdata.stateful_dataloader import StatefulDataLoader

def get_dl():
    d = list(range(100))
    return StatefulDataLoader(d, batch_size=1, shuffle=True)

def run_through(dl):
    for i, b in enumerate(dl):
        if i == 0:
            print(b)

def run_for_goes(goes):
    sd = None
    c = 0
    for n in goes:
        dl = get_dl()

        if None is not sd:
            print("loading state dict:", sd)
            dl.load_state_dict(sd)
            print("recall: loaded:", sd)
            print("state dict is now:", dl.state_dict())

        for j in range(n):
            print(c, j)
            run_through(dl)
            c += 1
    
        sd = dl.state_dict()

print("===")
run_for_goes([2,2])
print("===")
run_for_goes([4])
```

expected output:
```
===
0 0
tensor([45])
1 1
tensor([33])
loading state dict: {'_index_sampler_state': {'samples_yielded': 100, 'sampler_iter_state': {'generator': tensor([1, 0, 0,  ..., 0, 0, 0], dtype=torch.uint8), 'yielded': 100}}, '_sampler_iter_state': None, '_sampler_iter_yielded': 100, '_num_yielded': 100, '_IterableDataset_len_called': None, '_shared_seed': None, 'fetcher_state': None, 'dataset_state': None, '_iterator_finished': True}
recall: loaded: {'_index_sampler_state': {'samples_yielded': 100, 'sampler_iter_state': {'generator': tensor([1, 0, 0,  ..., 0, 0, 0], dtype=torch.uint8), 'yielded': 100}}, '_sampler_iter_state': None, '_sampler_iter_yielded': 100, '_num_yielded': 100, '_IterableDataset_len_called': None, '_shared_seed': None, 'fetcher_state': None, 'dataset_state': None, '_iterator_finished': True}
state dict is now: {'_index_sampler_state': {'samples_yielded': 100, 'sampler_iter_state': {'generator': tensor([1, 0, 0,  ..., 0, 0, 0], dtype=torch.uint8), 'yielded': 100}}, '_sampler_iter_state': None, '_sampler_iter_yielded': 100, '_num_yielded': 100, '_IterableDataset_len_called': None, '_shared_seed': None, 'fetcher_state': None, 'dataset_state': None, '_iterator_finished': True}
2 0
tensor([62])
3 1
tensor([19])
===
0 0
tensor([45])
1 1
tensor([33])
2 2
tensor([62])
3 3
tensor([19])
```

obtained output:
```
===
0 0
tensor([45])
1 1
tensor([33])
loading state dict: {'_index_sampler_state': {'samples_yielded': 100, 'sampler_iter_state': {'generator': tensor([1, 0, 0,  ..., 0, 0, 0], dtype=torch.uint8), 'yielded': 100}}, '_sampler_iter_state': None, '_sampler_iter_yielded': 100, '_num_yielded': 100, '_IterableDataset_len_called': None, '_shared_seed': None, 'fetcher_state': None, 'dataset_state': None, '_iterator_finished': True}
recall: loaded: {'_index_sampler_state': {'samples_yielded': 100, 'sampler_iter_state': {'generator': tensor([1, 0, 0,  ..., 0, 0, 0], dtype=torch.uint8), 'yielded': 100}}, '_sampler_iter_state': None, '_sampler_iter_yielded': 100, '_num_yielded': 100, '_IterableDataset_len_called': None, '_shared_seed': None, 'fetcher_state': None, 'dataset_state': None, '_iterator_finished': True}
state dict is now: {'_index_sampler_state': {'samples_yielded': 0, 'sampler_iter_state': {'generator': tensor([1, 0, 0,  ..., 0, 0, 0], dtype=torch.uint8), 'yielded': 0}}, '_sampler_iter_state': None, '_sampler_iter_yielded': 100, '_num_yielded': 100, '_IterableDataset_len_called': None, '_shared_seed': None, 'fetcher_state': None, 'dataset_state': None, '_iterator_finished': True}
2 0
3 1
tensor([62])
===
0 0
tensor([45])
1 1
tensor([33])
2 2
tensor([62])
3 3
tensor([19])
```

### Versions

PyTorch version: 2.2.0
Is debug build: False

OS: macOS 15.0.1 (arm64)
GCC version: Could not collect
Clang version: 16.0.0 (clang-1600.0.26.6)
CMake version: Could not collect
Libc version: N/A

Python version: 3.11.5 (main, Sep 11 2023, 08:31:25) [Clang 14.0.6 ] (64-bit runtime)
Python platform: macOS-15.0.1-arm64-arm-64bit
Is CUDA available: False
CUDA runtime version: No CUDA
CUDA_MODULE_LOADING set to: N/A
cuDNN version: No CUDA
HIP runtime version: N/A
MIOpen runtime version: N/A
Is XNNPACK available: True

CPU:
Apple M1 Pro

[pip3] msgpack-numpy==0.4.8
[pip3] numpy==1.26.3
[pip3] pytorch-lightning==2.0.3
[pip3] torch==2.2.0
[pip3] torchaudio==2.2.0
[pip3] torchdata==0.10.1
[pip3] torchmetrics==1.1.2
[pip3] torchvision==0.15.2a0
[conda] msgpack-numpy             0.4.8                    pypi_0    pypi
[conda] numpy                     1.26.3          py311he598dae_0  
[conda] numpy-base                1.26.3          py311hfbfe69c_0  
[conda] pytorch                   2.2.0                  py3.11_0    pytorch
[conda] pytorch-lightning         2.0.3           py311hca03da5_0  
[conda] torchaudio                2.2.0                 py311_cpu    pytorch
[conda] torchdata                 0.10.1                   pypi_0    pypi
[conda] torchmetrics              1.1.2           py311hca03da5_0  
[conda] torchvision               0.15.2          cpu_py311he74fb5d_0

ramanishsingh:
Add tests to reproduce and fix #1437

### DIFF
--- a/test/stateful_dataloader/test_state_dict.py
+++ b/test/stateful_dataloader/test_state_dict.py
@@ -1314,7 +1314,7 @@ class TestDatasetIteratorStateDuplication_shard0(TestCase):
                 dataset=dataset,
                 num_workers=num_workers,
                 collate_fn=identity,
-                multiprocessing_context="forkserver" if IS_MACOS and num_workers else None,
+                multiprocessing_context=("forkserver" if IS_MACOS and num_workers else None),
             )
             it = iter(dl)
             # Fetch at least one batch from each worker
@@ -1325,7 +1325,10 @@ class TestDatasetIteratorStateDuplication_shard0(TestCase):
             if num_workers > 0:
                 for i in range(num_workers):
                     # Ensure worker state is stored only once if the dataset is also the iterator
-                    self.assertEqual(state_dict["_snapshot"]["_worker_snapshots"][f"worker_{i}"]["dataset_state"], None)
+                    self.assertEqual(
+                        state_dict["_snapshot"]["_worker_snapshots"][f"worker_{i}"]["dataset_state"],
+                        None,
+                    )
                     self.assertTrue(
                         state_dict["_snapshot"]["_worker_snapshots"][f"worker_{i}"]["fetcher_state"][
                             "dataset_iter_state"
@@ -1440,47 +1443,79 @@ class TestFastStateDictRequestRoundRobin_shard3(TestCase):
     def test_fast_state_dict_request_skip_steps(self) -> None:
         self._run_test(17, 19)
 
-class TestMultiEpochState(TestCase):
+
+class TestMultiEpochSDL_shard0(TestCase):
     def get_map_dl(self, data_size=100, num_workers=0, batch_size=1, shuffle=False):
-        dataset = DummyMapDataset(data_size, shuffle=shuffle)
+        dataset = DummyMapDataset(data_size, shuffle=False)
         return StatefulDataLoader(
             dataset=dataset,
             num_workers=num_workers,
             batch_size=batch_size,
-            multiprocessing_context=(
-                "forkserver" if IS_MACOS and num_workers else None
-            ),
+            shuffle=shuffle,
+            multiprocessing_context=("forkserver" if IS_MACOS and num_workers else None),
         )
-    
+
     def _run(self, data_size, num_workers, batch_size, shuffle=False):
-        dataloader = self.get_map_dl(data_size=data_size,num_workers=num_workers, batch_size=batch_size, shuffle=shuffle)
+        dl1 = self.get_map_dl(
+            data_size=data_size,
+            num_workers=num_workers,
+            batch_size=batch_size,
+            shuffle=shuffle,
+        )
         # Run through the dataloader for 2 epochs and count the number of items yielded
         num_items_yielded = 0
+        dl1_items = []
         for _ in range(2):
-            for _ in dataloader:
+            for batch in dl1:
+                dl1_items.append(batch)
                 num_items_yielded += 1
         # Save the state dict
-        state_dict = dataloader.state_dict()
+        state_dict = dl1.state_dict()
         # Create a new StatefulDataLoader instance and load the state dict
-        new_dataloader = self.get_map_dl(
-            num_workers=num_workers, batch_size=batch_size, shuffle=shuffle
+        new_dl1 = self.get_map_dl(
+            data_size=data_size,
+            num_workers=num_workers,
+            batch_size=batch_size,
+            shuffle=shuffle,
         )
-        new_dataloader.load_state_dict(state_dict)
+        new_dl1.load_state_dict(state_dict)
         # Run through the new dataloader for another 2 epochs and count the number of items yielded
         additional_num_items_yielded = 0
         for i in range(2):
             epoch_num_items_yielded = 0
-            for _ in new_dataloader:
+            for batch in new_dl1:
+                dl1_items.append(batch)
                 epoch_num_items_yielded += 1
             additional_num_items_yielded += epoch_num_items_yielded
         # Check that the total number of items yielded is correct
-        self.assertEqual(num_items_yielded + additional_num_items_yielded, data_size*4)
+        self.assertEqual(num_items_yielded + additional_num_items_yielded, data_size * 4)
 
+        # now run a second dataloder for 4 epochs and check if the order is same.
+        dl2 = self.get_map_dl(
+            data_size=data_size,
+            num_workers=num_workers,
+            batch_size=batch_size,
+            shuffle=shuffle,
+        )
+        dl2_items = []
+        for _ in range(4):
+            for batch in dl2:
+                dl2_items.append(batch)
+
+        self.assertEqual(dl1_items, dl2_items)
 
     def test_main_process(self):
         self._run(100, 0, 1, False)
+
     def test_multiprocess(self):
         self._run(100, 2, 1, False)
+
+    def test_main_process_shuffle(self):
+        self._run(100, 0, 1, True)
+
+    def test_multiprocess_shuffle(self):
+        self._run(100, 2, 1, True)
+
 
 class TestMultiEpochState_shard0(TestCase):
     def get_iterable_dl(self, pw, num_workers):
--- a/test/stateful_dataloader/test_state_dict.py
+++ b/test/stateful_dataloader/test_state_dict.py
@@ -83,7 +83,9 @@ class DummySampler(torch.utils.data.Sampler):
         return self.size
 
 
-class DummyIteratorIterableDataset(torch.utils.data.IterableDataset, Iterator, Stateful):
+class DummyIteratorIterableDataset(
+    torch.utils.data.IterableDataset, Iterator, Stateful
+):
     def __init__(self, samples, shuffle, include_generator):
         self.samples = samples
         self.shuffle = shuffle
@@ -139,7 +141,10 @@ class DummyIterableDataset(torch.utils.data.IterableDataset):
 class DummyMapDataset(torch.utils.data.Dataset):
     def __init__(self, size, shuffle, include_generator=True):
         self.size = size
-        self.data = [{"id": i, "strcol": f"strcol_{i}", "listcol": [i, i + 1, i + 2]} for i in range(size)]
+        self.data = [
+            {"id": i, "strcol": f"strcol_{i}", "listcol": [i, i + 1, i + 2]}
+            for i in range(size)
+        ]
         self.shuffle = shuffle
         self.include_generator = include_generator
 
@@ -202,7 +207,9 @@ class TestStatefulDataLoaderIterable_shard0(TestCase):
     def _get_dataset(self, shuffle):
         return DummyIterableDataset([0, 100, 37], shuffle=shuffle)
 
-    def _run_and_checkpoint(self, num_workers, batch_size, pw, interrupt, every_n_steps=1, shuffle=False):
+    def _run_and_checkpoint(
+        self, num_workers, batch_size, pw, interrupt, every_n_steps=1, shuffle=False
+    ):
         dataset = self._get_dataset(shuffle)
         dl = StatefulDataLoader(
             dataset=dataset,
@@ -270,7 +277,9 @@ class TestStatefulDataLoaderIterable_shard0(TestCase):
     def test_mp_every_n_steps(self):
         batch_size = 7
         for every_n_steps, interrupt in itertools.product([2, 5], [0, 1, 10]):
-            with self.subTest(every_n_steps=every_n_steps, batch_size=batch_size, interrupt=interrupt):
+            with self.subTest(
+                every_n_steps=every_n_steps, batch_size=batch_size, interrupt=interrupt
+            ):
                 self._run_and_checkpoint(
                     num_workers=3,
                     batch_size=batch_size,
@@ -291,7 +300,9 @@ class TestStatefulDataLoaderIterable_shard0(TestCase):
 
 
 class TestStatefulDataLoaderMap_shard1(TestStatefulDataLoaderIterable_shard0):
-    def _run_and_checkpoint(self, num_workers, batch_size, pw, interrupt, every_n_steps=1, shuffle=False):
+    def _run_and_checkpoint(
+        self, num_workers, batch_size, pw, interrupt, every_n_steps=1, shuffle=False
+    ):
         if num_workers == 0:
             return
         dataset = DummyMapDataset(100, shuffle=shuffle)
@@ -344,7 +355,9 @@ class TestStatefulDataLoaderMap_shard1(TestStatefulDataLoaderIterable_shard0):
 
 
 class TestStatefulSampler_shard1(TestStatefulDataLoaderIterable_shard0):
-    def _run_and_checkpoint(self, num_workers, batch_size, pw, interrupt, every_n_steps=1, shuffle=False):
+    def _run_and_checkpoint(
+        self, num_workers, batch_size, pw, interrupt, every_n_steps=1, shuffle=False
+    ):
         dataset = DummyMapDataset(100, shuffle=shuffle)
         sampler = DummySampler(len(dataset))
         dl = StatefulDataLoader(
@@ -472,7 +485,9 @@ class GeneratorSampler(torch.utils.data.Sampler):
 
 
 class TestStatefulDataLoaderGenerator_shard2(TestStatefulDataLoaderIterable_shard0):
-    def _run_and_checkpoint(self, num_workers, batch_size, pw, interrupt, every_n_steps=1, shuffle=False):
+    def _run_and_checkpoint(
+        self, num_workers, batch_size, pw, interrupt, every_n_steps=1, shuffle=False
+    ):
         dataset = GeneratorIterable([0, 100, 37])
         dl = StatefulDataLoader(
             dataset=dataset,
@@ -521,8 +536,12 @@ class TestStatefulDataLoaderGenerator_shard2(TestStatefulDataLoaderIterable_shar
         self.assertEqual(batches, exp)
 
 
-class TestStatefulDataLoaderGeneratorNoState_shard2(TestStatefulDataLoaderIterable_shard0):
-    def _run_and_checkpoint(self, num_workers, batch_size, pw, interrupt, every_n_steps=1, shuffle=False):
+class TestStatefulDataLoaderGeneratorNoState_shard2(
+    TestStatefulDataLoaderIterable_shard0
+):
+    def _run_and_checkpoint(
+        self, num_workers, batch_size, pw, interrupt, every_n_steps=1, shuffle=False
+    ):
         dataset = GeneratorIterableNoState([0, 100, 37])
         dl = StatefulDataLoader(
             dataset=dataset,
@@ -582,7 +601,9 @@ class TestSnapshotZero_shard2(TestCase):
                 collate_fn=identity,
                 snapshot_every_n_steps=every_n_steps,
                 persistent_workers=pw,
-                multiprocessing_context=("forkserver" if IS_MACOS and num_workers else None),
+                multiprocessing_context=(
+                    "forkserver" if IS_MACOS and num_workers else None
+                ),
             )
 
             it = iter(dl)
@@ -605,7 +626,9 @@ class TestSnapshotZero_shard2(TestCase):
                 collate_fn=identity,
                 snapshot_every_n_steps=every_n_steps,
                 persistent_workers=pw,
-                multiprocessing_context=("forkserver" if IS_MACOS and num_workers else None),
+                multiprocessing_context=(
+                    "forkserver" if IS_MACOS and num_workers else None
+                ),
             )
 
             it = iter(dl)
@@ -628,7 +651,9 @@ class TestSnapshotZero_shard2(TestCase):
                 collate_fn=identity,
                 snapshot_every_n_steps=every_n_steps,
                 persistent_workers=pw,
-                multiprocessing_context=("forkserver" if IS_MACOS and num_workers else None),
+                multiprocessing_context=(
+                    "forkserver" if IS_MACOS and num_workers else None
+                ),
             )
 
             it = iter(dl)
@@ -652,7 +677,9 @@ class TestSnapshotZero_shard2(TestCase):
                 collate_fn=identity,
                 snapshot_every_n_steps=every_n_steps,
                 persistent_workers=pw,
-                multiprocessing_context=("forkserver" if IS_MACOS and num_workers else None),
+                multiprocessing_context=(
+                    "forkserver" if IS_MACOS and num_workers else None
+                ),
             )
 
             it = iter(dl)
@@ -667,7 +694,9 @@ class TestSnapshotZero_shard2(TestCase):
     def test_map_iterrupted_shuffle(self):
         every_n_steps = 10
 
-        for pw, num_workers, every_n_steps in itertools.product([False, True], [0, 2], [1, 15]):
+        for pw, num_workers, every_n_steps in itertools.product(
+            [False, True], [0, 2], [1, 15]
+        ):
             dataset = DummyMapDataset(10, shuffle=True)
             dl = StatefulDataLoader(
                 dataset=dataset,
@@ -676,7 +705,9 @@ class TestSnapshotZero_shard2(TestCase):
                 collate_fn=identity,
                 snapshot_every_n_steps=every_n_steps,
                 persistent_workers=pw if num_workers > 0 else False,
-                multiprocessing_context=("forkserver" if IS_MACOS and num_workers else None),
+                multiprocessing_context=(
+                    "forkserver" if IS_MACOS and num_workers else None
+                ),
             )
 
             it = iter(dl)
@@ -712,7 +743,9 @@ class TestSnapshotEnd_shard2(TestCase):
                 snapshot_every_n_steps=every_n_steps,
                 persistent_workers=pw,
                 batch_size=bs,
-                multiprocessing_context=("forkserver" if IS_MACOS and num_workers else None),
+                multiprocessing_context=(
+                    "forkserver" if IS_MACOS and num_workers else None
+                ),
             )
             exp = list(dl)
             state_end = dl.state_dict()
@@ -728,7 +761,9 @@ class TestSnapshotEnd_shard2(TestCase):
                 snapshot_every_n_steps=every_n_steps,
                 persistent_workers=pw,
                 batch_size=bs,
-                multiprocessing_context=("forkserver" if IS_MACOS and num_workers else None),
+                multiprocessing_context=(
+                    "forkserver" if IS_MACOS and num_workers else None
+                ),
             )
             it = iter(dl)
             for _ in range(2):
@@ -750,7 +785,9 @@ class TestSnapshotEnd_shard2(TestCase):
                 snapshot_every_n_steps=every_n_steps,
                 persistent_workers=pw,
                 batch_size=bs,
-                multiprocessing_context=("forkserver" if IS_MACOS and num_workers else None),
+                multiprocessing_context=(
+                    "forkserver" if IS_MACOS and num_workers else None
+                ),
             )
             exp = list(dl)
             state_end = dl.state_dict()
@@ -766,7 +803,9 @@ class TestSnapshotEnd_shard2(TestCase):
                 snapshot_every_n_steps=every_n_steps,
                 persistent_workers=pw,
                 batch_size=bs,
-                multiprocessing_context=("forkserver" if IS_MACOS and num_workers else None),
+                multiprocessing_context=(
+                    "forkserver" if IS_MACOS and num_workers else None
+                ),
             )
             it = iter(dl)
             for _ in range(2):
@@ -791,7 +830,9 @@ class TestSnapshotEnd_shard2(TestCase):
                 persistent_workers=pw,
                 batch_size=bs,
                 generator=g,
-                multiprocessing_context=("forkserver" if IS_MACOS and num_workers else None),
+                multiprocessing_context=(
+                    "forkserver" if IS_MACOS and num_workers else None
+                ),
             )
             list(dl)
             state_end = dl.state_dict()
@@ -806,7 +847,9 @@ class TestSnapshotEnd_shard2(TestCase):
                 persistent_workers=pw,
                 batch_size=bs,
                 generator=g,
-                multiprocessing_context=("forkserver" if IS_MACOS and num_workers else None),
+                multiprocessing_context=(
+                    "forkserver" if IS_MACOS and num_workers else None
+                ),
             )
             dl.load_state_dict(state_end)
             batches = list(dl)
@@ -828,7 +871,9 @@ class TestSnapshotEnd_shard2(TestCase):
                 persistent_workers=pw,
                 batch_size=bs,
                 generator=generator,
-                multiprocessing_context=("forkserver" if IS_MACOS and num_workers else None),
+                multiprocessing_context=(
+                    "forkserver" if IS_MACOS and num_workers else None
+                ),
             )
             list(dl)
             state_end = dl.state_dict()
@@ -843,7 +888,9 @@ class TestSnapshotEnd_shard2(TestCase):
                 persistent_workers=pw,
                 batch_size=bs,
                 generator=generator,
-                multiprocessing_context=("forkserver" if IS_MACOS and num_workers else None),
+                multiprocessing_context=(
+                    "forkserver" if IS_MACOS and num_workers else None
+                ),
             )
             dl.load_state_dict(state_end)
             batches = list(dl)
@@ -863,7 +910,9 @@ class TestSnapshotEnd_shard2(TestCase):
                 snapshot_every_n_steps=every_n_steps,
                 persistent_workers=pw,
                 batch_size=bs,
-                multiprocessing_context=("forkserver" if IS_MACOS and num_workers else None),
+                multiprocessing_context=(
+                    "forkserver" if IS_MACOS and num_workers else None
+                ),
             )
             list(dl)
             state_end = dl.state_dict()
@@ -878,7 +927,9 @@ class TestSnapshotEnd_shard2(TestCase):
                 snapshot_every_n_steps=every_n_steps,
                 persistent_workers=pw,
                 batch_size=bs,
-                multiprocessing_context=("forkserver" if IS_MACOS and num_workers else None),
+                multiprocessing_context=(
+                    "forkserver" if IS_MACOS and num_workers else None
+                ),
             )
             dl.load_state_dict(state_end)
             batches = list(dl)
@@ -896,7 +947,9 @@ class TestNumWorkersMismatch_shard3(TestCase):
                 dataset=dataset,
                 num_workers=initial_num_workers,
                 collate_fn=identity,
-                multiprocessing_context=("forkserver" if IS_MACOS and initial_num_workers else None),
+                multiprocessing_context=(
+                    "forkserver" if IS_MACOS and initial_num_workers else None
+                ),
             )
             state = dl.state_dict()
 
@@ -908,7 +961,9 @@ class TestNumWorkersMismatch_shard3(TestCase):
                 dataset=dataset,
                 num_workers=num_workers,
                 collate_fn=identity,
-                multiprocessing_context=("forkserver" if IS_MACOS and num_workers else None),
+                multiprocessing_context=(
+                    "forkserver" if IS_MACOS and num_workers else None
+                ),
             )
             dl.load_state_dict(state)
             try:
@@ -994,7 +1049,9 @@ class TestFastStateDictRequest_shard3(TestCase):
 class TestJsonSerDe_shard3(TestCase):
     def _run_test_iterable(self, num_workers):
         interrupt = 4
-        dataset = DummyIterableDataset([0, 100, 37], shuffle=False, include_generator=False)
+        dataset = DummyIterableDataset(
+            [0, 100, 37], shuffle=False, include_generator=False
+        )
         dl = StatefulDataLoader(
             dataset=dataset,
             num_workers=num_workers,
@@ -1256,7 +1313,9 @@ class TestInitialState_shard0(TestCase):
 class TestStatefulDataLoaderIterable2_shard0(TestStatefulDataLoaderIterable_shard0):
     # Perform sanity test checks with the iterable dataset that is also an iterator
     def _get_dataset(self, shuffle):
-        return DummyIteratorIterableDataset(list(range(100)), shuffle=shuffle, include_generator=True)
+        return DummyIteratorIterableDataset(
+            list(range(100)), shuffle=shuffle, include_generator=True
+        )
 
 
 class TestDynamicStateIterableDataset_shard0(TestCase):
@@ -1274,7 +1333,9 @@ class TestDynamicStateIterableDataset_shard0(TestCase):
         for _ in range((num_workers + 1) * 2):
             next(it)
         state_dict = dl.state_dict()
-        worker_state = state_dict["_snapshot"]["_worker_snapshots"]["worker_0"]["fetcher_state"]["dataset_iter_state"]
+        worker_state = state_dict["_snapshot"]["_worker_snapshots"]["worker_0"][
+            "fetcher_state"
+        ]["dataset_iter_state"]
         self.assertEqual(len(worker_state), 7)
         deep_copy_state_dict = deepcopy(state_dict)
 
@@ -1284,9 +1345,9 @@ class TestDynamicStateIterableDataset_shard0(TestCase):
         next_state_dict = dl.state_dict()
         self.assertEqual(state_dict, deep_copy_state_dict)
         self.assertFalse(state_dict == next_state_dict)
-        worker_state = next_state_dict["_snapshot"]["_worker_snapshots"]["worker_0"]["fetcher_state"][
-            "dataset_iter_state"
-        ]
+        worker_state = next_state_dict["_snapshot"]["_worker_snapshots"]["worker_0"][
+            "fetcher_state"
+        ]["dataset_iter_state"]
         self.assertEqual(len(worker_state), 11)
 
         dl = StatefulDataLoader(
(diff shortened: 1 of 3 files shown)


ramanishsingh:
This does not solve the problem as it just restarts the dataloader and produces the same batches again.

gailweiss:
if its any help, while creating the issue i noticed that after loading the state dict, the resulting state dict in the dataloader is different from the one that was loaded - for example, by setting "samples_yielded" to 0 when the loaded one had 100 (see the prints in #1437 ), (and possibly more differences - I haven't checked).  looking at the code in this commit, it seems that samples_yielded is being set manually - maybe that is the root of the problem?

torchdata/stateful_dataloader/sampler.py
```diff
@@ -89,44 +107,29 @@ def load_state_dict(self, state_dict: Dict[str, Any]) -> None:
...
+        ):
+            # We skip x samples if underlying sampler is not stateful
+            for _ in range(self.samples_yielded):
+                next(self.sampler_iter)
+
+        # Skip one epoch if we were at the end of the last epoch
+        if hasattr(self.sampler, "__len__") and self.samples_yielded == len(self.sampler):
+            for _ in self.sampler_iter:
+                pass
+
```
gailweiss, line 121: hi, sorry to nitpick, but wouldn't we actually not want to skip an epoch when reloading a state dict from an unfinished epoch? i.e. IMO it makes sense to recognise the difference between a state_dict obtained via this process:
```
for b in dl:
    sd_in = dl.state_dict()  # when the for ends, sd_in will describe an "about-to-finish" state
```

vs a state_dict obtained via this process:
```
for b in dl:
    pass
sd_out = dl.state_dict()  # sd_out will describe a "just finished" state
```

I think it makes sense to have an empty epoch immediately after loading sd_in, but a full one immediately after loading sd_out.

In particular, is it possible that issue #1437 is solved just by the new line `self.samples_yielded = state_dict[self._SAMPLES_YIELDED]`?

ramanishsingh: Hi @gailweiss 
I think the code changes take care of that. In the `state_dict` ['_iterator_finished']: False or True depending on if it is just- finishing or actually-finished. 
When we load sd_in, the next epoch is still empty because `state_dict` ['_iterator_finished']: False

Is this the behavior you expect : 

```Python
from torchdata.stateful_dataloader import StatefulDataLoader


def get_dl():
    d = list(range(10))
    return StatefulDataLoader(d, batch_size=1, shuffle=True)


dl = get_dl()
for i, b in enumerate(dl):
    if i == 0:
        print(i, b)
    sd_in = dl.state_dict()
print("sd_in", sd_in)


dl = get_dl()
dl.load_state_dict(sd_in)  # load the "about-to-finish" state
batches_after_sdin_load = []
for i, b in enumerate(dl):
    batches_after_sdin_load.append(b)
    if i == 0:
        print(i, b)

print("batches_after_sdin_load", batches_after_sdin_load)
dl = get_dl()
for i, b in enumerate(dl):
    if i == 0:
        print(i, b)

sd_out = (
    dl.state_dict()
)  # when the for ends, sd_out will describe a "just-finished" state
print("sd_out", sd_out)
dl = get_dl()
dl.load_state_dict(sd_out)  # load the "about-to-finish" state
batches_after_sdout_load = []
for i, b in enumerate(dl):
    batches_after_sdout_load.append(b)
    if i == 0:
        print(i, b)

print("batches_after_sdout_load", batches_after_sdout_load)
```

Output:
```
0 tensor([5])
sd_in {'_index_sampler_state': None, '_sampler_iter_state': {'samples_yielded': 10, 'sampler_iter_state': {'generator': tensor([1, 0, 0,  ..., 0, 0, 0], dtype=torch.uint8), 'yielded': 10}}, '_sampler_iter_yielded': 10, '_num_yielded': 10, '_IterableDataset_len_called': None, '_shared_seed': None, 'fetcher_state': None, 'dataset_state': None, '_iterator_finished': False}
skipping epoch
batches_after_sdin_load []
0 tensor([5])
sd_out {'_index_sampler_state': None, '_sampler_iter_state': {'samples_yielded': 10, 'sampler_iter_state': {'generator': tensor([1, 0, 0,  ..., 0, 0, 0], dtype=torch.uint8), 'yielded': 10}}, '_sampler_iter_yielded': 10, '_num_yielded': 10, '_IterableDataset_len_called': None, '_shared_seed': None, 'fetcher_state': None, 'dataset_state': None, '_iterator_finished': True}
skipping epoch
0 tensor([2])
batches_after_sdout_load [tensor([2]), tensor([8]), tensor([1]), tensor([5]), tensor([6]), tensor([9]), tensor([3]), tensor([7]), tensor([0]), tensor([4])]
```

gailweiss: Hi, yes, that's great! Thank you and sorry for the false complaint!

test/stateful_dataloader/test_state_dict.py
```diff
@@ -1441,6 +1444,154 @@ def test_fast_state_dict_request_skip_steps(self) -> None:
         self._run_test(17, 19)
 
 
+class TestMultiEpochSDL_shard0(TestCase):
+    def get_map_dl(self, data_size=100, num_workers=0, batch_size=1, shuffle=False):
```
andrewkho, line 1448: can you test this for num_workers=0 and say num_workers=2?

torchdata/stateful_dataloader/sampler.py
```diff
@@ -89,44 +107,29 @@ def load_state_dict(self, state_dict: Dict[str, Any]) -> None:
...
 
+        if not (isinstance(self.sampler, Stateful) or isinstance(self.sampler_iter, Stateful)) and not isinstance(
+            self.sampler, _InfiniteConstantSampler
+        ):
+            # We skip x samples if underlying sampler is not stateful
+            for _ in range(self.samples_yielded):
+                next(self.sampler_iter)
+
+        # Skip one epoch if we were at the end of the last epoch
+        if hasattr(self.sampler, "__len__") and self.samples_yielded == len(self.sampler):
```
andrewkho, line 118: Do all samplers have `__len__`? This feels brittle to me

torchdata/stateful_dataloader/sampler.py
```diff
@@ -89,44 +107,29 @@ def load_state_dict(self, state_dict: Dict[str, Any]) -> None:
...
+            for _ in self.sampler_iter:
+                pass
```
andrewkho, line 120: If state_dict is saved at the end of epoch 1/start of epoch 2, why do we need to do this? What if the sampler is not stateful, would it happen twice due to lines 114? Something seems a bit off to me

test/stateful_dataloader/test_state_dict.py
ramanishsingh: Removed defaults values for `num_workers`.
Tests below run for both 0 and 2.

ramanishsingh:
@andrewkho 
Thanks.
I find that during the loading of the state dict, if the `_StatefulRandomSamplerIterator` is at its end, its `self.next_yielded` value is becoming None due to iter re-init from somewhere. 
To tackle that, I am artificially making it 0 by checking if we are at the end of an epoch and exhausting the iterator. 
I think it is less brittle than checking the length of the sampler and skipping one whole epoch. Please lmk your thoughts.

ramanishsingh:
TLDR:  After refactoring BatchSampler, the same batch sequence is repeated in the epoch following a reload due to _iterator_finished being True. Update the generator in the state_dict after each iteration to cache the latest state, ensuring RNG resumes correctly even if next_yielded is reset to 0.

Problem: 
After [breaking](https://github.com/pytorch/data/commit/45527650eeab12d3b0a001b0e17c8cb8bf77d2c1) the BatchSampler into BatchSampler and _BatchSamplerIterator, we encountered an issue where the same sequence of batches is produced in the epoch immediately following a reload, mirroring the last epoch before saving the state_dict.

Root Cause:
This issue arises because the dl state_dict is saved after the epoch completes, resulting in _iterator_finished being set to True. To illustrate, consider the epoch after reloading as epoch 3. In the state_dict of the RandomSampler (a subset of the dl state_dict), key items include self.next_yielded and the state of the generator. When a StatefulDataLoader (SDL) is instantiated with num_workers = 0 and batches are retrieved, the __iter__ method in SDL is invoked. This method utilizes next_iter_state (or the loaded_state_dict) to obtain an iterator. During this process, the generator, sampler_iter, etc., are reloaded. However, since _iterator_finished is True, the _StatefulSingleProcessDataLoaderIter that was generated is discarded, and a new one is created with state_dict=None. Consequently, we lose the RandomSampler state information because next_yielded is reset to 0, and the generator state remains at the start of epoch 2.

Proposed Solution:
While there may be more efficient solutions, one potential approach (that I have implemented) is to update the generator in the state_dict upon completing an iteration. By doing so, we cache the latest generator state, allowing us to resume RNG production from the correct point even when the RandomSampler is reset with next_yielded = 0.

ramanishsingh:
For future ref: 
https://github.com/pytorch/pytorch/blob/652880e8403b58ca44d482f200a8991b8b326e88/torch/utils/data/sampler.py#L190

In torch.utils.data RandomSampler, we are changing the state of the generator even if `self.num_samples % n==0`. A more efficient and simpler solution would be to add a check `self.num_samples % n>0` and then generate a random permutation.

torchdata/stateful_dataloader/sampler.py
```diff
@@ -89,44 +110,28 @@ def load_state_dict(self, state_dict: Dict[str, Any]) -> None:
...
+        if not (isinstance(self.sampler, Stateful) or isinstance(self.sampler_iter, Stateful)) and not isinstance(
+            self.sampler, _InfiniteConstantSampler
+        ):
+            # We skip x samples if underlying sampler is not stateful
+            for _ in range(self.samples_yielded):
+                next(self.sampler_iter)
+
+    def update_state_dict(self) -> None:
+        if isinstance(self.sampler_iter, Stateful) and hasattr(self.sampler_iter, "update_state_dict"):
+            self.sampler_iter.update_state_dict()
```
andrewkho, line 122: Leaving thoughts here:

* this introduces a new (undocumented) API for samplers (update_state_dict()) which isn't great
* If the alternative solution is to update the base sampler code, this means users bringing their own samplers won't work out of the box, which also isn't great.

torchdata/stateful_dataloader/stateful_dataloader.py
```diff
@@ -479,7 +484,11 @@ def __init__(self, loader, next_iter_state=None):
             self.load_state_dict(next_iter_state)
         else:
             self._dataset_fetcher = _DatasetKind.create_fetcher(
-                self._dataset_kind, self._dataset, self._auto_collation, self._collate_fn, self._drop_last
+                self._dataset_kind,
+                self._dataset,
+                self._auto_collation,
+                self._collate_fn,
+                self._drop_last,
```
andrewkho, line 491: nit: can we remove the whitespace changes (ie run pre-commit to format all these files) so it's easier to review?

torchdata/stateful_dataloader/sampler.py
```diff
@@ -42,13 +41,21 @@ def load_state_dict(self, state_dict: Dict[str, Any]) -> None:
         self.sampler.generator.set_state(state_dict[self._GENERATOR])
         self.next_yielded = state_dict[self._YIELDED]
 
+    def update_state_dict(self) -> None:
```
andrewkho, line 44: @ramanishsingh can you try to remove this change, and fork the random sampler code here and make it an iterator instead of a generator? It will take a bit of refactoring but should make the save/load state more explicit, easier to reason about and handle. 

If that works, check that the old RandomSampler works too as a drop-in replacement (it shoudl just fast-forward) to ensure users bringing their own (non-stateful) samplers will have things work as well

andrewkho (changes requested):

torchdata/stateful_dataloader/sampler.py
ramanishsingh: Thanks!

torchdata/stateful_dataloader/stateful_dataloader.py
ramanishsingh: I have run precommit and these whitespaces are actually added by `precommit`. If I revert back everything I get lint errors.

gailweiss:
Hi, I tried to be clever and implement a temporary workaround myself, but no luck.
In the process however, I wrote a bunch of tests that I think reflect how the stateful dataloader should behave, it currently fails several. Maybe they will be helpful? They also surface the fact that once the stateful_dataloaders do get randomly initialised (see https://github.com/pytorch/data/issues/1440 ), even if by a user just making them with different generators at first, even more things break - so it seems like loading the state dict doesnt completely wipe the dataloader's state...?

Code:
```
import torch
from copy import deepcopy


# test support functions: printing with notes, comparing dictionaries, comparing epochs,
# getting dataloaders, running tests, etc

class PrefPrint:
    def __init__(self):
        self.pref = ""
    def print(self, *a, **kw):
        print(self.pref, *a, **kw)
    
printer = PrefPrint()

def pprint(*a, **kw):
    printer.print(*a, **kw)

def same_continuation(orig, loaded, el, expect_false=False):
    l1 = [b.item() for b in orig]
    l2 = [b.item() for b in loaded]
    # print(l1, l2)
    if len(l1) != el:
        pprint(f"orig dl's epoch wrong length: {l1} (expected {el})")
        return False
    if len(l2) != el:
        pprint(f"loaded dl's epoch wrong length: {l2} (expected {el}) (orig dl's epoch was correct length)")
        return False
    res = (l1 == l2)
    if (not res) and (not expect_false):
        pprint("orig vs loaded dl epochs:", l1, l2)
    return res

def equal_state_dicts(d1, d2):
    def _comp_dicts(d1, d2, pref=">>\t"):
        if sorted(list(d1.keys())) != sorted(list(d2.keys())):
            return f"{pref} diff keys: {list(sd1.keys())}, {list(sd2.keys())}"
        for k in d1.keys():
            if isinstance(d1[k], dict):
                return _comp_dicts(d1[k], d2[k], pref=f"{pref}in {k}: ")
            elif isinstance(d1[k], torch.Tensor):
                if False in (d1[k] == d2[k]):
                    return f"{pref} diff on {k}: {d1[k].tolist()} vs {d2[k].tolist()}"
            elif d1[k] != d2[k]:
                return f"{pref} diff on {k}: {d1[k]} vs {d2[k]}"
        return ""
    res = _comp_dicts(d1, d2)
    if res:
        pprint(res)
        return False
    return True

def get_dl(from_seed=None, gen=None):
    assert None in [from_seed, gen]
    d = list(range(n_samples))
    if None is not from_seed:
        gen = torch.Generator()
        gen.manual_seed(from_seed)
    return DataLoader(d, generator=gen, batch_size=1, shuffle=True)

def run_test(f, *a, **kw):
    printer.pref = f"in {f.__name__}:"
    res = f(*a, **kw)
    printer.pref = ""
    return res

def test(f):
    tests.append(f)
    return f

tests = []
n_samples = 10 # length all the dataloaders will be in the tests

# tests themselves


# 1. 2 different inits create different shuffles
@test
def diffshuff_as_standard():
    # n = 10 -> 10! shuffles -> highly unlikely to accidentally get same epoch
    return not same_continuation(get_dl(), get_dl(), n_samples, expect_false=True)

# 2. 2 different inits from the same generator create the same shuffles
@test
def sameshuff_when_asked():
    if not same_continuation(get_dl(from_seed=1), get_dl(from_seed=1), n_samples):
        pprint("mismatch on seed 1")
        return False
    gen1, gen2 = torch.Generator(), torch.Generator()
    gen1.seed()
    gen2.set_state(gen1.get_state())
    if not same_continuation(get_dl(gen=gen1), get_dl(gen=gen2), n_samples):
        pprint("mismatch on random gen")
        return False
    return True

# 3. getting state dict and loading it after own state has changed recovers 
# previous state (ie state dict properly detached once taken)
@test
def go_back():
    dl1 = get_dl()
    sd = dl1.state_dict()
    a1 = [b.item() for b in dl1]
    dl1.load_state_dict(sd)
    a2 = [b.item() for b in dl1]
    if not a1 == a2:
        pprint("doesnt walk back steps")
        return False
    return True

# 4. loading a state dict taken from the middle of an epoch continues that epoch
@test
def resume_from_partial():
    dl1, dl2 = get_dl(), get_dl()
    a1 = []
    p = 3
    for i, b in enumerate(dl1):
        if i == p:
            sd = dl1.state_dict()
        if i > p:
            a1.append(b.item())
    dl2.load_state_dict(sd)
    a2 = [b.item() for b in dl2]
    res = a1 == a2
    if not res:
        pprint("diff continuation from partial sd")
        pprint(a1, a2)
    if not same_continuation(dl1, dl2, n_samples):
        pprint("next epoch after partial sd not aligned")
        res = False
    return res

# 5. loading a state dict taken after a dataloader has left a partial epoch resumes (like that dataloader) from that dataloader's next full epoch
@test
def resume_after_partial():
    dl1, dl2 = get_dl(), get_dl()
    a = []
    for i, b in enumerate(dl1):
        a.append(b.item())
        if i > 3:
            break
    sd = dl1.state_dict()
    dl2.load_state_dict(sd)
    return same_continuation(dl1, dl2, n_samples)


# 6. loading a state_dict taken after a full epoch has completed continues smoothly from that dataloader's next epoch 
# this highlights the main issue raised in https://github.com/pytorch/data/issues/1437
@test
def resume_between():
    g1, g2 = 2, 2
    dl1, dl2 = get_dl(), get_dl()
    _ = [[b.item() for b in dl1] for _ in range(g1)]
    sd = dl1.state_dict()
    dl2.load_state_dict(sd)
    for i in range(g2):
        if not same_continuation(dl1, dl2, n_samples):  # should be full epochs here
            pprint(f"epoch {i} after load is broken")
            return False
    return True


# 7. loading a state_dict taken partway through an epoch, specifically at the last batch (but before leaving the loop) resumes at same point (i.e., inside epoch, with 0 batches left), and then continues to the same next epoch as the original dataloader
@test
def resume_end():
    dl1, dl2 = get_dl(), get_dl()
    l1 = []
    for i, b in enumerate(dl1):
        l1.append(b.item())
        sd = dl1.state_dict()
        if i == n_samples - 1:
            break  # dl1 not finished
    pprint("dl1 first epoch is:", l1)

    dl2.load_state_dict(sd)
    l = [b.item() for b in dl2]
    if len(l) > 0:
        print("resuming finishing state does not lead to empty epoch")
        return False
    if not same_continuation(dl1, dl2, n_samples):
        pprint("resuming finishing state does not move (after finishing) to same next epoch")
        pprint("is this an off-by-one? dl1 next next epoch would be:", [b.item() for b in dl1])
        return False
    return True

# dataloader variants
from torchdata.stateful_dataloader import StatefulDataLoader as _DataLoader
import torch
from copy import deepcopy

class LoudDataLoader(_DataLoader):
        sd1 = deepcopy(state_dict)
        super(LoudDataLoader, self).load_state_dict(sd1)
        if not equal_state_dicts(sd1, self.state_dict()):
            pprint("[[ state dict after loading different from provided state dict! ]]")

# DataLoader to fix https://github.com/pytorch/data/issues/1440
class LoudDataLoader1440(LoudDataLoader):
    def __init__(self, *a, **kw):
        if None is kw.get("generator", None):
            kw["generator"] = torch.Generator()
            kw["generator"].seed()  # for some reason important for getting it going
        super().__init__(*a, **kw)



dlclasses = {"base":_DataLoader, "loud":LoudDataLoader, "loud1440":LoudDataLoader1440}
results = {}
for n, DLC in dlclasses.items():
    DataLoader = DLC
    print(f"\n================\nrunning {n} stateful dataloader tests")
    results[n] = {f.__name__ : run_test(f) for f in tests}

print("\n\n=======")
for n, r in results.items():
    print(f"\n=======\n {n} stateful dataloader test results:")
    names = [f.__name__ for f in tests]
    print("\n".join(f"{n}: \t\t[{r[n]}]" for n in names))
```

output:
```
================
running base stateful dataloader tests
in resume_after_partial: loaded dl's epoch wrong length: [8, 9, 3, 7, 4] (expected 10) (orig dl's epoch was correct length)
in resume_between: loaded dl's epoch wrong length: [] (expected 10) (orig dl's epoch was correct length)
in resume_between: epoch 0 after load is broken
in resume_end: dl1 first epoch is: [5, 6, 1, 2, 0, 8, 9, 3, 7, 4]
in resume_end: orig vs loaded dl epochs: [2, 3, 1, 8, 9, 0, 6, 7, 4, 5] [2, 8, 1, 5, 6, 9, 3, 7, 0, 4]
in resume_end: resuming finishing state does not move (after finishing) to same next epoch
in resume_end: is this an off-by-one? dl1 next next epoch would be: [2, 9, 8, 3, 6, 7, 1, 0, 4, 5]

================
running loud stateful dataloader tests
in go_back: >>	in _index_sampler_state:  diff on samples_yielded: 0 vs 10
in go_back: [[ state dict after loading different from provided state dict! ]]
in resume_from_partial: >>	in _index_sampler_state:  diff on samples_yielded: 4 vs 0
in resume_from_partial: [[ state dict after loading different from provided state dict! ]]
in resume_after_partial: >>	in _index_sampler_state:  diff on samples_yielded: 5 vs 0
in resume_after_partial: [[ state dict after loading different from provided state dict! ]]
in resume_after_partial: loaded dl's epoch wrong length: [8, 9, 3, 7, 4] (expected 10) (orig dl's epoch was correct length)
in resume_between: >>	in _index_sampler_state:  diff on samples_yielded: 10 vs 0
in resume_between: [[ state dict after loading different from provided state dict! ]]
in resume_between: loaded dl's epoch wrong length: [] (expected 10) (orig dl's epoch was correct length)
in resume_between: epoch 0 after load is broken
in resume_end: dl1 first epoch is: [5, 6, 1, 2, 0, 8, 9, 3, 7, 4]
in resume_end: >>	in _index_sampler_state:  diff on samples_yielded: 10 vs 0
in resume_end: [[ state dict after loading different from provided state dict! ]]
in resume_end: orig vs loaded dl epochs: [2, 3, 1, 8, 9, 0, 6, 7, 4, 5] [2, 8, 1, 5, 6, 9, 3, 7, 0, 4]
in resume_end: resuming finishing state does not move (after finishing) to same next epoch
in resume_end: is this an off-by-one? dl1 next next epoch would be: [2, 9, 8, 3, 6, 7, 1, 0, 4, 5]

================
running loud1440 stateful dataloader tests
in go_back: >>	in _index_sampler_state:  diff on samples_yielded: 0 vs 10
in go_back: [[ state dict after loading different from provided state dict! ]]
in go_back: doesnt walk back steps
in resume_from_partial: >>	in _index_sampler_state:  diff on samples_yielded: 4 vs 0
in resume_from_partial: [[ state dict after loading different from provided state dict! ]]
in resume_from_partial: diff continuation from partial sd
in resume_from_partial: [1, 5, 9, 8, 3, 0] [8, 9, 1, 6, 5, 3]
in resume_from_partial: orig vs loaded dl epochs: [0, 1, 7, 8, 6, 5, 3, 9, 2, 4] [5, 4, 2, 0, 7, 1, 8, 6, 3, 9]
in resume_from_partial: next epoch after partial sd not aligned
in resume_after_partial: >>	in _index_sampler_state:  diff on samples_yielded: 5 vs 0
in resume_after_partial: [[ state dict after loading different from provided state dict! ]]
in resume_after_partial: loaded dl's epoch wrong length: [1, 8, 6, 2, 3] (expected 10) (orig dl's epoch was correct length)
in resume_between: >>	in _index_sampler_state:  diff on samples_yielded: 10 vs 0
in resume_between: [[ state dict after loading different from provided state dict! ]]
in resume_between: loaded dl's epoch wrong length: [] (expected 10) (orig dl's epoch was correct length)
in resume_between: epoch 0 after load is broken
in resume_end: dl1 first epoch is: [6, 3, 5, 0, 2, 9, 7, 1, 4, 8]
in resume_end: >>	in _index_sampler_state:  diff on samples_yielded: 10 vs 0
in resume_end: [[ state dict after loading different from provided state dict! ]]
in resume_end: orig vs loaded dl epochs: [0, 6, 5, 7, 8, 2, 1, 4, 9, 3] [8, 4, 7, 3, 0, 6, 9, 2, 5, 1]
in resume_end: resuming finishing state does not move (after finishing) to same next epoch
in resume_end: is this an off-by-one? dl1 next next epoch would be: [6, 2, 7, 9, 1, 4, 3, 5, 0, 8]


=======

=======
 base stateful dataloader test results:
diffshuff_as_standard: 		[False]
sameshuff_when_asked: 		[True]
go_back: 		[True]
resume_from_partial: 		[True]
resume_after_partial: 		[False]
resume_between: 		[False]
resume_end: 		[False]

=======
 loud stateful dataloader test results:
diffshuff_as_standard: 		[False]
sameshuff_when_asked: 		[True]
go_back: 		[True]
resume_from_partial: 		[True]
resume_after_partial: 		[False]
resume_between: 		[False]
resume_end: 		[False]

=======
 loud1440 stateful dataloader test results:
diffshuff_as_standard: 		[True]
sameshuff_when_asked: 		[True]
go_back: 		[False]
resume_from_partial: 		[False]
resume_after_partial: 		[False]
resume_between: 		[False]
resume_end: 		[False]
```

ramanishsingh:
@gailweiss can you try these examples using the code on this branch? I guess the code in this branch should be working fine (except for the random generator thing, which is taken care of in #1441).

torchdata/stateful_dataloader/sampler.py
```diff
@@ -18,58 +19,162 @@ class _StatefulRandomSamplerIterator(Iterator[int], Stateful):
...
-    def __next__(self) -> int:
+    def __iter__(self):
+        return self
+
+    def __next__(self):
+        if self.replacement:
+            num_full_chunks = self.num_samples // self.chunk_size
+            remainder = self.num_samples % self.chunk_size
+            if self.chunk_index < num_full_chunks:
+                if self.perm is None or not self.perm:
```
andrewkho, line 43: Why are we doing this chunking logic? I don't think this will give the correct distribution, ie it will depend on num_chunks

torchdata/stateful_dataloader/sampler.py
```diff
@@ -18,58 +19,162 @@ class _StatefulRandomSamplerIterator(Iterator[int], Stateful):
...
+                self.yielded += 1
+                return value
+            elif remainder > 0:
+                if self.perm is None or not self.perm:
+                    self.perm = torch.randint(
+                        high=self.n,
+                        size=(remainder,),
+                        dtype=torch.int64,
+                        generator=self.sampler.generator,
+                    ).tolist()
```
andrewkho, line 65: Can we do implement with less repeated code and if/else branches? I think we need __next__ to be as lightweight as possible. 

I would suggest we generate the sequence once (ie at startup/load_state_dict or at the beginning of the function).

eg

self.seq = generate_sequence(replacement=True)
ret = self.seq[self.i]
self.i += 1
return ret

andrewkho: Is the chunking logic present in random sampler?

torchdata/stateful_dataloader/sampler.py
ramanishsingh: Hi @andrewkho , 
that's how they are doing it in https://github.com/pytorch/pytorch/blob/b18e3c01aa8bfee37078b4a06cef4361bf63b36b/torch/utils/data/sampler.py#L177
Generate 32 random numbers at once and yield from them. 
In their case `self.num_samples // 32` is essentially num_full_chunks.

gailweiss:
@ramanishsingh sure thing!

It seems better, but not all fixed. And sadly still breaks if the dataloaders are initiated with random generators. New output for same code:

```
================
running base stateful dataloader tests
in resume_after_partial: loaded dl's epoch wrong length: [8, 9, 3, 7, 4] (expected 10) (orig dl's epoch was correct length)
in resume_end: dl1 first epoch is: [5, 6, 1, 2, 0, 8, 9, 3, 7, 4]

================
running loud stateful dataloader tests
in resume_after_partial: loaded dl's epoch wrong length: [8, 9, 3, 7, 4] (expected 10) (orig dl's epoch was correct length)
in resume_end: dl1 first epoch is: [5, 6, 1, 2, 0, 8, 9, 3, 7, 4]

================
running loud1440 stateful dataloader tests
in go_back: doesnt walk back steps
in resume_from_partial: diff continuation from partial sd
in resume_from_partial: [5, 2, 6, 7, 9, 1] [9, 2, 3, 7, 8, 6]
in resume_from_partial: orig vs loaded dl epochs: [3, 5, 2, 6, 1, 4, 0, 7, 9, 8] [2, 9, 5, 6, 4, 7, 0, 1, 8, 3]
in resume_from_partial: next epoch after partial sd not aligned
in resume_after_partial: loaded dl's epoch wrong length: [8, 5, 7, 0, 9] (expected 10) (orig dl's epoch was correct length)
in resume_between: orig vs loaded dl epochs: [8, 2, 9, 3, 0, 5, 4, 7, 1, 6] [0, 5, 6, 1, 3, 4, 8, 7, 2, 9]
in resume_between: epoch 0 after load is broken
in resume_end: dl1 first epoch is: [7, 3, 8, 4, 0, 2, 6, 1, 9, 5]
in resume_end: orig vs loaded dl epochs: [6, 9, 1, 0, 3, 4, 7, 5, 8, 2] [7, 2, 4, 9, 5, 1, 6, 3, 0, 8]
in resume_end: resuming finishing state does not move (after finishing) to same next epoch
in resume_end: is this an off-by-one? dl1 next next epoch would be: [3, 6, 7, 8, 2, 1, 9, 5, 0, 4]


=======

=======
 base stateful dataloader test results:
diffshuff_as_standard: 		[False]
sameshuff_when_asked: 		[True]
go_back: 		[True]
resume_from_partial: 		[True]
resume_after_partial: 		[False]
resume_between: 		[True]
resume_end: 		[True]

=======
 loud stateful dataloader test results:
diffshuff_as_standard: 		[False]
sameshuff_when_asked: 		[True]
go_back: 		[True]
resume_from_partial: 		[True]
resume_after_partial: 		[False]
resume_between: 		[True]
resume_end: 		[True]

=======
 loud1440 stateful dataloader test results:
diffshuff_as_standard: 		[True]
sameshuff_when_asked: 		[True]
go_back: 		[False]
resume_from_partial: 		[False]
resume_after_partial: 		[False]
resume_between: 		[False]
resume_end: 		[False]
```

torchdata/stateful_dataloader/sampler.py
ramanishsingh: Yes, chunking is present in the torch.utils.data RandomSampler.

ramanishsingh:
@gailweiss 
Focusing on the base_dataloder (the randomness issue is fixed in #1441 ) I think according to you it is failing in this test
```
def resume_after_partial():
    dl1, dl2 = get_dl(), get_dl()
    a = []
    for i, b in enumerate(dl1):
        a.append(b.item())
        if i > 3:
            break
    sd = dl1.state_dict()
    dl2.load_state_dict(sd)
    return same_continuation(dl1, dl2, n_samples)
```
However, if you want the continuation behavior, you'd need to create an iterator from the dataloader.
Here's an example and output

```python
from torchdata.stateful_dataloader import StatefulDataLoader


def get_dl(num_workers=0):
    d = list(range(10))
    return StatefulDataLoader(d, batch_size=1, shuffle=True, num_workers=num_workers)


num_workers = 0
dataloader = get_dl(num_workers)
res_before_interruption = []
dl_iter = iter(dataloader)
for i, b in enumerate(dl_iter):
    res_before_interruption.append(b)
    if i == 2:
        print("Interrupting")
        break
    

res_after_interruption = []
for i, b in enumerate(dl_iter):
    res_after_interruption.append(b)
print("len(res_before_interruption)=",len(res_before_interruption),";    len(res_after_interruption)=",len(res_after_interruption))

print("res_before_interruption", res_before_interruption)
print("res_after_interruption", res_after_interruption)

print("===", "doing exp again, this time without interruption", "====")

dataloader = get_dl(num_workers)
res_wo_interruption = []
for i, b in enumerate(dataloader):
    res_wo_interruption.append(b)
print("res_wo_interruption", res_wo_interruption)
```

Output:
```
Interrupting
len(res_before_interruption)= 3 ;    len(res_after_interruption)= 7
res_before_interruption [tensor([5]), tensor([6]), tensor([1])]
res_after_interruption [tensor([2]), tensor([0]), tensor([8]), tensor([9]), tensor([3]), tensor([7]), tensor([4])]
=== doing exp again, this time without interruption ====
res_wo_interruption [tensor([5]), tensor([6]), tensor([1]), tensor([2]), tensor([0]), tensor([8]), tensor([9]), tensor([3]), tensor([7]), tensor([4])]


```

I hope this helps.

gailweiss:
I have assumed orig_dl conveys the intended behaviour - if you look at the prints from resume_after_partial, you will see that orig_dl yields 10 batches from this state. One way or the other the two have to align though - I am not manipulating orig_dl in any way between the state transfer and the test that it is equivalent to loaded_dl

ramanishsingh:
Hi @gailweiss 
I just want to confirm that we are talking about the same test. 
Is this the test?
```python
def resume_after_partial():
    dl1, dl2 = get_dl(), get_dl()
    a = []
    for i, b in enumerate(dl1):
        a.append(b.item())
        if i > 3:
            break
    sd = dl1.state_dict()
    dl2.load_state_dict(sd)
    return same_continuation(dl1, dl2, n_samples)
```
If so, this test should return that `dl1` and `dl2` are not same. 
This is because, when we do `for i, b in enumerate(dl1):` it creates the `__iter__` [method](https://github.com/pytorch/data/blob/fe6b4054e5e8b19f3dae615e852ef47515179abf/torchdata/stateful_dataloader/stateful_dataloader.py#L391) which in turn calls the [`get_iterator`](https://github.com/pytorch/data/blob/main/torchdata/stateful_dataloader/stateful_dataloader.py#L381) and it gets the iterator [here](https://github.com/pytorch/data/blob/main/torchdata/stateful_dataloader/stateful_dataloader.py#L384) which can use the state stored in `next_iter_state`.
Now even though `dl1` has yielded three batches, its `next_iter_state` is None because we haven't updated it yet. It stays as None. 
But for `dl2` we do `dl2.load_state_dict(sd)` and thus its `next_iter_state` is populated and gets the information that it has yielded 3 batches/samples. 
Thus when we call `same_continuation(dl1, dl2, n_samples)` it returns that these two dataloaders return different batches. Which should be the expected behavior.
While the `dl.state_dict()` method returns the latest state of the dataloader, it (the state) needs to be applied to the dataloader so that when the next time its iterator is called, it moves to the right pointer. 

This behavior is inline with how the normal `Dataloader` in `torch.utils.data`  and normal lists behave.
I understand it is not exactly an apples to apples comparison as the normal Dataloader isn't stateful, but it's just how the Design of the StatefulDataloader is, you need to load the state_dict such the iterator gets the right state to start from.

Behavior of DataLoader
```python
from torch.utils.data import DataLoader
d = list(range(5))
dl = DataLoader(d, batch_size=1)
for i, b in enumerate(dl):
    print(i, b)
    if i == 2:
      break
print("breaking and restarting")
for i, b in enumerate(dl):
    print(i, b)
```
Output
```
0 tensor([0])
1 tensor([1])
2 tensor([2])
breaking and restarting
0 tensor([0])
1 tensor([1])
2 tensor([2])
3 tensor([3])
4 tensor([4])
```
If you want statefulness without loading state dict, you can make an iterator (like I specified in my previous comment and example) and iterate over it. 
Please let me know if this makes sense. :)

Other than that, I believe this PR solves #1437 . I have added tests for all those cases. 

Thanks!

gailweiss:
Hi, I think we’re talking about the same test, but to me this is unexpected behaviour. In particular, to me the line: 

```
dl2.load_state_dict(dl1.state_dict())
```

(which is equivalent to the lines 

```
sd = dl1.state_dict()
dl2.load_state_dict(sd)
```
in this test, as there is no intervention between them, not even a loop break)

should always lead to a dl2 that is identical to dl1. To have a case where this is not true - regardless of what was done with dl1 beforehand - seems to me counter to the goal of a state.



Obviously I do not understand the implementation behind this, but given the tests resume_from_end and resume_between are passing now, I imagine the dataloaders do have the mechanism to differentiate between whether the state was taken from within or without a loop, so hopefully this should be possible to manage?

gailweiss:
If it helps, I think giving: 1) the iterator access to the dataloader, 2) wrapping the iterator in a yield loop, 3) listening in that loop for a GeneratorExit, and finally 4) updating the dataloader as necessary on loop entry/exit will allow maintaining this inside/outside loop state, which should in turn allow successful state transfer to other dataloaders. Example on a dummy iterator:


```
class MyIterator:
    def __init__(self, source):
        self.source = source
        self.i = -1
        self.n = len(source.d)
        
    def __next__(self):
        self.source.in_loop = True
        self.i += 1
        if self.i >= self.n:
            raise StopIteration
        return self.source.d[self.i]

    def close(self):
        self.source.in_loop = False
    

def wrapiterator(it):
    while True:
        try:
            yield next(it)
        except GeneratorExit:
            it.close()
            break
        except StopIteration:
            it.close()
            break

class MyRange:
    def __init__(self, n):
        self.d = list(range(1,n+1))
        self.in_loop = False
    def print_state(self, expected):
        print("loop state aligns with expected:", expected == self.in_loop)
    def __iter__(self):
        return wrapiterator(MyIterator(self))
```

with these classes, the code:
```

a = MyRange(3)
a.print_state(False)
for i in a:
    print(i)
    a.print_state(True)
    break
a.print_state(False)

for i in a:
    print(i)
    a.print_state(True)
a.print_state(False)
```
yields:
```
loop state aligns with expected: True
1
loop state aligns with expected: True
loop state aligns with expected: True
1
loop state aligns with expected: True
2
loop state aligns with expected: True
3
loop state aligns with expected: True
loop state aligns with expected: True
```

ramanishsingh:
Hi @gailweiss 
I understand your argument. The current behavior is just how the StatefulDataloader is designed and it is in parity with how the normal Dataloader works, and the expected behavior regarding the need to (re)load the state_dict might be different for different people. I will surely bring it up with the team, and if any changes are required, I suspect that it will be a slightly bigger refactor. Thus, it is out of scope of this current PR as this PR is aimed for #1437 , which I think it solves (please lmk if you see any issues with that). For your recent comments, after discussion with the team, I will start a new issue and tag you there. 
Thanks for your hard work!

gailweiss:
Hi, sure, I understand how this may make the work bigger, and that you may want to turn it into a different PR. I personally do hold firm that the line `dl2.load_state_dict(dl1.state_dict())` should always yield `dl2` that continues identically to `dl1` at that point. Looking at the [message](https://github.com/pytorch/data/pull/1439#issuecomment-2657982333) above, I think there has been a miscommunication at some point, so I want to be clear: for the purposes of this conversation, I don't really care what `dl1` does one way or the other, and I am sure that it does follow the regular DataLoader as desired. My only goal here is that `dl2` mimicks it correctly after receiving its state, and `resume_after_partial` surfaces a case where it does not. 

Because all these tests relate to the transfer of state from `dl1` to `dl2`, I suspect that whatever change will fix `resume_after_partial` (I suggest adding an `in_loop` flag to the state, and controlling this flag through the iterator, which can recognise both finished and broken loops by listening for `GeneratorExit` and `StopIteration`) will also be useful for fixing the original issue (end of epoch state transfer), and I think solving them all together will likely yield cleaner code than handling them independently. I note my suggestion only works if you assume no nesting of loops over the StatefulDataLoader! But my impression is that this assumption is already made for the StatefulDataLoader anyway.

andrewkho:
Hi @gailweiss thanks for your detailed experiments and thoughts! Current behaviour is as-designed ([as tested here](https://colab.research.google.com/drive/1i_DHzn2sjHhRFDYH-yc9ekf-2o34HUYn?usp=sharing)). I think you raise some interesting points though and I'd suggest creating an issue to continue the discussion as it seems like a UX/design issue and not so relevant to this current PR (which fixes the original issue you found, thank you for that!) 

As an aside, this is inherently an issue when dealing with state on Iterable objects, which is why we decided to center `torchdata.nodes` on Iterators instead of Iterables, which have much better [defined behaviour in these scenarios](https://pytorch.org/data/beta/what_is_torchdata_nodes.html#no-generator-basenodes)

gailweiss:
Thank you both for working on this, and for fixing this issue! Thinking about the difference between iterables and iterators, I appreciate the tension between what people intuitively expect when checkpointing versus how an iterable actually behaves, and how this gets confusing in the details - it's no wonder you moved to iterators directly! Still, I will consider opening another issue detailing my current thoughts and suggestions, and you can see if you want to take them in :)

ramanishsingh:
@gailweiss yes, please feel free to open the issue, and we'd be happy to discuss more.

PR #1455, opened 2025-02-25
Rescalability layer

daviswer:
Implements rescaling of checkpoints to different world sizes and numbers of workers. User specifies in advance the number of data partitions, and when saving/loading checkpoints with different total workers, stateful guarantees are maintained: seen data is not revisited until the next epoch.

Based off of the datasets in the corresponding IBM [torchtitan PR](https://github.com/pytorch/torchtitan/pull/376), but with an adjusted rescaling and iteration mechanism to support greater flexibility and robustness (removes divisibility constraints from worker and shard counts, and guarantees only one open file per physical worker regardless of number of logical shards). Uses StatefulDataLoader and DCP to manage checkpointing from the master process. An epoch completion testing script is included for demo purposes. It is possible that the IBM datasets can be merged into the existing torchdata Nodes structure.

### Changes

- Add IBM rescalable datasets and checkpointing functions to `torchdata/stateful_dataloader/ibm_rescalable.py`
- Add demo script and correctness check to `examples/ibm_rescaling/rescaling_demo.py`

### DIFF
--- a/examples/ibm_rescaling/rescaling_demo.py
+++ b/examples/ibm_rescaling/rescaling_demo.py
@@ -45,7 +45,7 @@ mesh = dist.device_mesh.init_device_mesh("cpu", [world_size])
 placement = [dist.tensor.placement_types.Shard(0)]
 
 # Build dataloader
-data = DummyDataset("data", rank, world_size, delimiter_token=-1, seed=args.seed)
+data = DummyDataset(os.path.join(args.ckpt_path, "data"), rank, world_size, delimiter_token=-1, seed=args.seed)
 # Pretend that we're sampling over multiple sub-datasets
 subdatas = ["sub_dataset", "second_subdataset", "small_subdataset"]
 [os.makedirs(os.path.join(args.ckpt_path, "data", subdata), exist_ok=True) for subdata in subdatas]
--- a/examples/ibm_rescaling/rescaling_demo.py
+++ b/examples/ibm_rescaling/rescaling_demo.py
@@ -49,15 +49,15 @@ subdatas = ["sub_dataset", "second_subdataset", "small_subdataset"]
 # Build dataloader
 data = DummyDataset(os.path.join(args.ckpt_path, "data"), rank, world_size, delimiter_token=-1, seed=args.seed)
 # Pretend that we're sampling over multiple sub-datasets
-# data = SamplingDataset(
-#     os.path.join(args.ckpt_path, "data"),
-#     data,
-#     delimiter_token=-1,
-#     datasets=subdatas,
-#     weights=[12, 17, 5],
-# )
+data = SamplingDataset(
+    os.path.join(args.ckpt_path, "data"),
+    data,
+    delimiter_token=-1,
+    datasets=subdatas,
+    weights=[12, 17, 5],
+)
 # Apply rescalability layer
-# data = ScalableShardDataset(data, n_logical_shards=args.logical_shards)
+data = ScalableShardDataset(data, n_logical_shards=args.logical_shards)
 # Statelessly convert all outputs to tensors
 data = PreprocessDataset(data, torch.tensor)
 # Wrap in StatefulDataLoader
--- a/examples/ibm_rescaling/rescaling_demo.py
+++ b/examples/ibm_rescaling/rescaling_demo.py
@@ -137,6 +137,10 @@ else:
     # Get all vals onto each rank
     vals = dist.tensor.DTensor.from_local(vals, mesh, placement).full_tensor()
 
+    # Diag save
+    os.makedirs(os.path.join(args.ckpt_path, "diag"))
+    torch.save(data.state_dict(), os.path.join(args.ckpt_path, "diag", f"loader_state_{rank}.pth"))
+
     # Perform data coverage check on rank 0 only
     if rank == 0:
         # Invert avoid to get expected vals
--- a/examples/ibm_rescaling/rescaling_demo.py
+++ b/examples/ibm_rescaling/rescaling_demo.py
@@ -79,7 +79,7 @@ if not os.path.exists(datapath):
 # Build dataloader
 data = ScalableReader(datapath, rank, world_size, ArrowHandler, -1, seed=args.seed, max_chunksize=30, n_logical_shards=args.logical_shards)
 # Pad entries to make them batch-able
-data = PreprocessDataset(data, lambda x: x + [-1]*(30-len))
+data = PreprocessDataset(data, lambda x: x + [-1]*(30-len(x)))
 # Statelessly convert all outputs to tensors
 data = PreprocessDataset(data, torch.tensor)
 # Wrap in StatefulDataLoader
--- a/examples/ibm_rescaling/rescaling_demo.py
+++ b/examples/ibm_rescaling/rescaling_demo.py
@@ -125,8 +125,8 @@ else:
     if rank == 0:
         print("Checkpoint detected!")
     load_distributed_state_dict(data, ckpt_path, mesh)
-    print("FINAL")
-    time.sleep(10000)
+    # print("FINAL")
+    # time.sleep(10000)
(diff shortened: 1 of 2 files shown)


scotts:
Thanks for the work, @daviswer! Some first-level comments:

1. Let's create some unit tests based on the example demo. I think it makes sense for them to live in test/stateful_dataloader.
2. Let's name the Python file after the main abstraction users will use from so it, so `ibm_rescalable.py` should become `scalable_reader.py`.
3. The name `_WrapperDataset` is really generic. From the code and your comments, I think a name closer to the capability it provides might be `_NestedStatefulDataset`.
4. The class `_ShardFileHandler` should probably be an [abstract base class](https://docs.python.org/3/library/abc.html). And since we anticipate that others may end up creating their own shard handlers for other formats, we should probably consider a public API, so we should drop the leading `_`. We might also want to break it out into its own file, such as `shard_handler.py`. Then all future shard handlers would go in there.

daviswer:
Thanks @scotts , I made changes 2-4 and working on unit tests now. I'll note that `_StatefulDataset` and `_NestedStatefulDataset` largely represent legacy code, gluing things together until we decide we either want to merge this into Nodes, or use these to represent stateful datasets (in which case we'll need to rework them anyways with contracts/APIs/etc. per #1456)

torchdata/stateful_dataloader/scalable_reader.py
```diff
@@ -0,0 +1,678 @@
...
+    Reader for indexable, pre-tokenized PyArrow shard files.
+    Pyarrow shard files are expected to hold multiple RecordBatches,
+    where each RecordBatch has a "tokens" field consisting of
+    a single token list (i.e. each document is a single sequence
+    under a "token" field, and the file is a list of such sequences).
+
+    A preferred format as we can load document chunks without having to ever pull
+    the entire document or shard file, allowing for graceful handling of large documents.
+    Non-standard data format, though.
+    """
```
divyanshk, line 249: I wanted to confirm my understanding of the format of the pyarrow shard files. I am imagining a very large text file made up a thousands of tokens. That file is broken in multiple PyArrow shard files. Each of those PyArrow shard files is made up of multiple RecordBatches, each with a tokens field which is a list of tokens. That means each token is a 'row' (in that sense that RecordBatches are supposed to be a batch of records/rows). Is that right ?

Additionally, why do we not consider having list of tokens as a single row in the recordbatch? What is the value addition of using recordbatches here? Thanks.

daviswer: The general assumption is that each pyarrow shard file represents a collection of documents, rather than a single large document getting split over multiple files. But yes, each file is a collection of RecordBatches, each of which contains a single 'row' of text (a single document) in the `tokens` field. We use RecordBatches because that's how pyarrow [docs](https://arrow.apache.org/docs/python/ipc.html) suggest reading/writing random-access memory-mapped files, and we put a single document per RecordBatch to minimize overhead of loading individual documents in random order.

divyanshk: Cool. Can every document (even large ones) fit in a single RecordBatch ? 
The doc string mentions "as we can load document chunks without having to ever pull the entire document", here we are still referring to a single document being loaded in a RecordBatch?

daviswer: Yes - because pyarrow files are memory-mapped, RecordBatches (and slices) are loaded lazily. Until you request the actual slice, the RecordBatch is just metadata, so it can hold a document of any size without extra overhead

divyanshk:
Sharing some points that we discussed over the call. 
1. The core work here is in the data access layer, nor particularly in the data loader. I imagine we can figure out a way to give end users (think PyTorch users with established Dataset class) a `RescalableDataset` wrapper which converts their existing Dataset into ones which can be rescaled if one decides to start and re-start a job with a different world size. The `ScalableReader` is effectively that, although we should wonder if want to make the user give more inputs (like a custom file handler) or we can configure those inside the rescalable dataset wrapper.

This can feed directly into a canonical `StatefulDataLoader`, with `{save, load}_distributed_state_dict` functionality incorporated into `StatefulDataLoader`'s state_dict / load_state_dict methods as special cases for `RescalableDataset`. At this point I don't know how feasible that is (@daviswer brought up a good point whether we want to take a dependency on DCP, vs having a generic interface for any checkpointing API to work) but this seems like a simpler interface for users to onboard to.

2. So far the implementation is solving for text-heavy AI workloads. We should also align on whether we want to extend the scope to include other styles, like for eg, a data row being an arbitrary `Dict[str, Any]`, typical map-style datasets, typical HuggingFace vision datasets, etc.

3. I need to look at some internal data-access layer APIs to ensure we don't diverge too much.

@scotts @daviswer

daviswer:
Some thoughts on point 2 @divyanshk : we could definitely separate this out into a file-interface system, plus a separate rescaling-enabled interface between nested iterable Datasets and generic indexable structures. However, we may lose some capabilities in the process. In particular, the current approach is set up to a) handle sharding of indexable structures where the total number of indices is not known in advance (i.e. many shard files containing many documents, with limited access bandwidth) and b) ensure that no more than one file per device is open/active at a time, regardless of number of files/devices. If we abstract away notions of files/items behind a generic indexable interface, it becomes harder to maintain these guarantees. 

It may be possible to still make that work but I'd have to think through the approach some more.

@scotts

examples/data_rescaling/rescaling_demo.py
```diff
@@ -0,0 +1,158 @@
...
+ckpt_path = os.path.join(args.ckpt_path, "loader_dcp_state")
+if not os.path.exists(ckpt_path) or len(os.listdir(ckpt_path)) == 0:
+    os.makedirs(ckpt_path, exist_ok=True)
+    # Iterate, assemble values to exclude
+    if rank == 0:
+        print(f"No existing checkpoint. Processing {args.n_steps} steps.")
+
+    avoid = []
+    for i, inp in enumerate(data):
+        avoid.append(inp[:,0])
```
pradeepfn, line 98: noob q: what is the purpose of 'avoid' data vector ?

torchdata/stateful_dataloader/scalable_reader.py
```diff
@@ -0,0 +1,678 @@
...
+        storage_reader = reader,
+    )  # NOTE: assumes inp["state"] is same across all devices
+    dstate = inp["dstate"]
+    # Re-pack the set of rankX args
+    # NOTE: this is the step currently breaking the no-DCP path
+    keys = list(dstate.keys())
+    ranked_state = {k:dstate.pop(k) for k in keys if "rank" in k}
+    ranked_keylist = sorted(list(ranked_state.keys()))
+    compiled_ranked = [ranked_state[k] for k in ranked_keylist]
+    dstate[ranked_keylist[0][6:]] = compiled_ranked  # Drop "rank0." prefix
```
pradeepfn, line 662: I need some clarity on the custom re-scaling state. Will sync on this offline.

torchdata/stateful_dataloader/scalable_reader.py
```diff
@@ -0,0 +1,678 @@
...
+            self.shard_states[len(shard_states):,4] = torch.iinfo(torch.int).max
+        return None
+
+
+#### -------------------------    CHECKPOINT FUNCTIONS    ------------------------- ####
+
+
+def __pop_dstate(state, device_mesh, placements, create_dtensor=False):
+    """
+    Removes worker states from the StatefulDataLoader state dict, and fuses them into a single dict
```
divyanshk, line 581: Similar to the question I asked on Slack the other day, does it even make sense to load state after rescaling if what is loaded is only partial, for eg we throw loader state here or by disregarding scalars or RNG.

daviswer: Yes partial loading makes sense - for example we might have a data buffer coupled with an RNG. On rescaling, we would want to preserve and reshard the data buffer, but don't need to hold onto the RNG state(s), since we now have a different number of them

PR #1458, opened 2025-02-26
Update docs to reflect current status

scotts:
Removes the disclaimers about what will be removed, as that has all been removed.


ramanishsingh:
Thanks for cleaning the docs. 
Let's fix the lint errors (`pre-commit run --all-files` should help) and it is gtg.

ramanishsingh (approved):

PR #1459, opened 2025-02-27
Remove files with `dataloader2` and `datapipes` references

ramanishsingh:
Some old DLv2 and datapipes examples need to be removed.


ramanishsingh:
Essentially we don't have examples for anything except nodes. Shall we keep the utils and folders now so that we can build examples later on OR just delete everything right now and build from scratch later.

divyanshk:
LGTM! Let's create the folder again. Although we do have example notebooks you submitted https://fburl.com/geybvdbe

PR #1460, opened 2025-03-01
Clean extension.py

### LINKED ISSUE #1451: pip install with BUILD_S3=TRUE broken
### 🐛 Describe the bug

In conda-forge, we're building `torchdata` using `python -m pip install` (after setting some environment variables like `USE_SYSTEM_LIBS`). However, c64801fc6b7e9d3693a0b73621a5d7ac8b4b82e5 (CC @andrewkho) removed the root `CMakeLists.txt`, which now trivially fails the cmake build for the python extension 
https://github.com/pytorch/data/blob/d15522009ac707b55957f307648040f6a8d3536c/tools/setup_helpers/extension.py#L125-L126
when using 
https://github.com/pytorch/data/blob/d15522009ac707b55957f307648040f6a8d3536c/tools/setup_helpers/extension.py#L57-L58

Given that that commit comprehensively removed `torchdata/csrc` etc. the fact that the install breaks with `BUILD_S3=TRUE` is probably an oversight.

In fact, probably a lot of that helper module can be cleaned up, since (aside from the CMake call to nowhere) none of the following variables are being used anywhere anymore
https://github.com/pytorch/data/blob/d15522009ac707b55957f307648040f6a8d3536c/tools/setup_helpers/extension.py#L41-L44

### Versions

Starting from 0.10.1

ramanishsingh:
Fixes #1451

### DIFF
--- a/tools/setup_helpers/extension.py
+++ b/tools/setup_helpers/extension.py
@@ -4,20 +4,12 @@
 # This source code is licensed under the BSD-style license found in the
 # LICENSE file in the root directory of this source tree.
 
-import distutils.sysconfig
+
 import os
-import platform
-import subprocess
-import sys
 from pathlib import Path
 
-from setuptools.command.build_ext import build_ext
 
-
-__all__ = [
-    "get_ext_modules",
-    "CMakeBuild",
-]
+__all__ = ["get_ext_modules"]
 
 
 _THIS_DIR = Path(__file__).parent.resolve()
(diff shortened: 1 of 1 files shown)


tools/setup_helpers/extension.py
```diff
@@ -38,96 +30,5 @@ def _get_build(var, default=False):
...
-        subprocess.check_call(["cmake", str(_ROOT_DIR)] + cmake_args, cwd=self.build_temp)
-        subprocess.check_call(["cmake", "--build", "."] + build_args, cwd=self.build_temp)
-
-    def get_ext_filename(self, fullname):
-        ext_filename = super().get_ext_filename(fullname)
-        ext_filename_parts = ext_filename.split(".")
-        without_abi = ext_filename_parts[:-2] + ext_filename_parts[-1:]
-        ext_filename = ".".join(without_abi)
-        return ext_filename
+    return []
```
divyanshk, line 34: I don't know about this, can we check if we need to keep `get_ext_modules` ?

ramanishsingh: Sure. Thanks!

ramanishsingh: Sure. Thanks!

PR #1454, opened 2025-02-25
Add `Filter`

keunwoochoi:
Please read through our [contribution guide](https://github.com/pytorch/data/blob/main/CONTRIBUTING.md) prior to
creating your pull request.

- If you are adding a new node, ensure you read that section in the contribution guide, as it includes requirements for
  functionality and testing.

Fixes #{issue number}

### Changes

- Adding `Filter` node and its test (for now)
- Possibly TBD after discussion - as mentioned in #1452, I'd add `Shuffler`, `Header`, `Cycler` if the maintainers are okay with it.

Issue #1452, opened 2025-02-25
Open for contribution on utility nodes like `FilterNode`, `ShuffleNode`, `HeaderNode`, `CycleNode`?


keunwoochoi:
Hi, do you think this kind of nodes would be in the scope of Torchdata? Then I'm down to open a PR to add them. 

```python

import logging
import random
from collections import deque
from typing import Any, Callable, Deque, Dict, Optional, TypeVar
from typing import Any, Dict, Optional
from torchdata.nodes import BaseNode

logger = logging.getLogger(__name__)

X = TypeVar("X")
T = TypeVar("T")
U = TypeVar("U")


class FilterNode(BaseNode[T]):
    """Node that filters items from source node based on predicate function."""

    SOURCE_KEY = "source"

    def __init__(self, source_node: BaseNode[T], filter_fn: Callable[[T], bool]):
        super().__init__()
        self.source = source_node
        self.filter_fn = filter_fn

    def reset(self, initial_state: Optional[Dict[str, Any]] = None):
        super().reset(initial_state)
        self.source.reset(initial_state.get(self.SOURCE_KEY) if initial_state else None)

    def next(self) -> T:
        while True:
            item = next(self.source)
            if self.filter_fn(item):
                return item

    def get_state(self) -> Dict[str, Any]:
        return {self.SOURCE_KEY: self.source.state_dict()}


class ShuffleNode(BaseNode[T]):
    """Node that shuffles items from source node using a buffer."""

    SOURCE_KEY = "source"

    def __init__(self, source_node: BaseNode[T], buffer_size: int, seed: Optional[int] = None):
        super().__init__()
        if buffer_size < 1:
            raise ValueError("Buffer size must be at least 1")
        self.source = source_node
        self.buffer_size = buffer_size
        self.buffer: Deque[T] = deque()
        self.rng = random.Random(seed)
        self._initial_seed = seed

    def reset(self, initial_state: Optional[Dict[str, Any]] = None):
        super().reset(initial_state)
        self.buffer.clear()

        if initial_state is not None:
            self.source.reset(initial_state.get(self.SOURCE_KEY))
            self.rng.setstate(initial_state["rng_state"])
        else:
            self.source.reset()
            if self._initial_seed is not None:
                self.rng = random.Random(self._initial_seed)

    def _fill_buffer(self) -> bool:
        """Fill buffer with items from source. Returns True if any items were added."""
        try:
            while len(self.buffer) < self.buffer_size:
[...]

PR #1461, opened 2025-03-07
Adding Cycler, Header

keunwoochoi:
creating your pull request.

- If you are adding a new node, ensure you read that section in the contribution guide, as it includes requirements for
  functionality and testing.

Following up discussion #1452 and my previous PR #1454

### Changes

- Adds `Cycler`, `Header`, `Shuffler` and their tests.


torchdata/nodes/cycler.py
```diff
@@ -0,0 +1,84 @@
...
+            # the source is empty - just propagate the StopIteration without cycling
+            if not self._has_started:
+                raise StopIteration
+
+            # Otherwise, source is exhausted after yielding some items
+            # Increment cycle count and reset
+            self._num_cycles += 1
+            self.source.reset(None)
+
+            # Try again - if it's empty, this will raise StopIteration
```
divyanshk, line 71: At this point, the source shouldn't be empty after the reset. If it was empty it would raise in line 64. 

If this makes sense, we can update the comment.

torchdata/nodes/cycler.py
```diff
@@ -0,0 +1,84 @@
...
+            return next(self.source)
+
+    def get_state(self) -> Dict[str, Any]:
+        """Get the current state of the node.
+
+        Returns:
+            A dictionary containing the state of the source node and number of cycles completed.
```
divyanshk, line 78: Nit: Can we make it "Dict[str, Any] - A dictionary containing the state of the source node and number of cycles completed."?

torchdata/nodes/cycler.py
```diff
@@ -0,0 +1,84 @@
...
+    The node keeps track of how many times it has completed a full cycle through
+    the source.
+
+    Args:
+        source_node (BaseNode[T]): The source node to cycle through.
+    """
+
+    SOURCE_KEY = "source"
+    NUM_CYCLES_KEY = "num_cycles"
+    HAS_STARTED_KEY = "has_started"
```
divyanshk, line 23: Let's also include "num_yielded".

torchdata/nodes/header.py
```diff
@@ -0,0 +1,71 @@
...
+        source_node (BaseNode[T]): The source node to pull items from.
+        n (int): The maximum number of items to yield. Must be non-negative.
+    """
+
+    SOURCE_KEY = "source"
+    NUM_YIELDED_KEY = "num_yielded"
+
+    def __init__(self, source_node: BaseNode[T], n: int):
+        super().__init__()
+        if n < 0:
```
divyanshk, line 26: Should be throw a ValueError when n  = 0 ? that is like setting up a node to not do anything, seems like wrong user input.

torchdata/nodes/shuffler.py
```diff
@@ -0,0 +1,111 @@
...
+        Returns:
+            A dictionary containing the state of the source node, buffer,
+            number of items shuffled, and random state.
+        """
+        return {
+            self.SOURCE_KEY: self.source.state_dict(),
+            self.RNG_STATE_KEY: self.rng.getstate(),
+            self.BUFFER_KEY: list(self.buffer),
+            self.NUM_SHUFFLED_KEY: self._num_shuffled,
+            self.RANDOM_STATE_KEY: self.rng.getstate(),
```
divyanshk, line 110: RANDOM_STATE_KEY is a duplicate of RNG_STATE_KEY, let's remove RANDOM_STATE_KEY

torchdata/nodes/shuffler.py
```diff
@@ -0,0 +1,111 @@
...
+        source_node (BaseNode[T]): The source node to pull items from.
+        buffer_size (int): Size of the buffer used for shuffling. Must be at least 1.
+        seed (Optional[int]): Optional seed for the random number generator.
+    """
+
+    SOURCE_KEY = "source"
+    RNG_STATE_KEY = "rng_state"
+    BUFFER_KEY = "buffer"
+    NUM_SHUFFLED_KEY = "num_shuffled"
+    RANDOM_STATE_KEY = "random_state"
```
divyanshk, line 29: Lets include num yielded in state as well.

torchdata/nodes/shuffler.py
```diff
@@ -0,0 +1,111 @@
...
+        """Get the current state of the node.
+
+        Returns:
+            A dictionary containing the state of the source node, buffer,
+            number of items shuffled, and random state.
+        """
+        return {
+            self.SOURCE_KEY: self.source.state_dict(),
+            self.RNG_STATE_KEY: self.rng.getstate(),
+            self.BUFFER_KEY: list(self.buffer),
```
divyanshk, line 108: Should we avoid keeping the entire buffer as state ? since this would be a list of node objects it might not make sense.

torchdata/nodes/shuffler.py
```diff
@@ -0,0 +1,111 @@
...
+    """Node that shuffles items from source node using a buffer.
+
+    This node maintains a buffer of items from the source node and returns them
+    in a random order. The buffer is filled to capacity initially and then
+    replenished as items are removed. The randomization is controlled by a
+    random number generator with an optional seed for reproducibility.
+
+    Args:
+        source_node (BaseNode[T]): The source node to pull items from.
+        buffer_size (int): Size of the buffer used for shuffling. Must be at least 1.
```
divyanshk, line 21: Thinking from a user POV. Would they have a strong opinion on what the `buffer_size` argument should be? Maybe not. 

I fear this would end up being set as an arbitrarily large number just to maximize shuffling capacity.

Should we see if we can make it work without buffer_size and seed as an argument? We have something like this in MultiNodeWeightedSampler https://fburl.com/xyt2tacq

keunwoochoi: from my experience, i actually believe we should definitely let user control this. the ideal shuffle buffer size depends a lot per data source in my use-cases. 
for example, if a data source is not pre-shuffled globally and only sharded (which sucks but it would a good use-case of Shuffler,) the buffer size should be as large as the number of items in each shard.

if the buffer size is too large, yes a lot of other issues can occur. but perhaps that’s up to users really and they should understand how Shuffler works? especially since i’m not really sure there’s a non-trivial method to set a value that works for majority of scenarios i can imagine.

keunwoochoi: i’m not able to open the link. can you share it from the public code repo?

torchdata/nodes/cycler.py
keunwoochoi: it is an edge-case where the source node is blank. i think that's a fair consideration (e.g., the current Iterable node also can take a blank list without any error during instantiation).

torchdata/nodes/cycler.py
```diff
@@ -0,0 +1,84 @@
...
+    def reset(self, initial_state: Optional[Dict[str, Any]] = None):
+        """Reset the node to its initial state or to the provided state.
+
+        Args:
+            initial_state: Optional state dictionary to restore from.
+        """
+        super().reset(initial_state)
+        if initial_state is not None:
+            self._num_cycles = initial_state.get(self.NUM_CYCLES_KEY, 0)
+            self._has_started = initial_state.get(self.HAS_STARTED_KEY, False)
```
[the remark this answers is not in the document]
keunwoochoi: agreed, working on it in the next commit.

torchdata/nodes/header.py
keunwoochoi: i see it as a philosophical question, just like if we're going to allow [] as a source iterator of Iterable node. i think the easier consistency we can keep is to allow `n` to be 0. it will be probably useless, it might be a mistake, but is it definitely a bug? i'm not sure. it's not a hill i'd die on though.

torchdata/nodes/shuffler.py
```diff
@@ -0,0 +1,111 @@
...
+    def _fill_buffer(self) -> bool:
+        """Fill buffer with items from source.
+
+        Returns:
+            True if any items were added to the buffer, False otherwise.
+        """
+        try:
+            while len(self.buffer) < self.buffer_size:
+                self.buffer.append(next(self.source))
+            return True
```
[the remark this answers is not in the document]
keunwoochoi: good catch! i've updated the method to explicitly handle when the buffer is already full. 

the return value now clearly indicates whether the buffer has items after the call (`True`) or is empty (`False`). this matches how it's used in `next()` where we only raise `StopIteration` if both the buffer is empty and `fill_buffer()` returns `False`.

torchdata/nodes/shuffler.py
keunwoochoi: thanks for the update. i still think the buffer size is way more (indeed, 100%) up to the user unlike the example where some nice parameter can help things work faster; without hurting affecting any of the core feature.

keunwoochoi:
hi @divyanshk , thanks for the careful review. i updated the code per the review comments. most of the requested changes are done, i think one exception is if we should allow like if `n=0` which imo we should.

divyanshk:
@keunwoochoi Thank you for the updates, will get back soon!

keunwoochoi:
@divyanshk kindly reminding of this :)

torchdata/nodes/shuffler.py
divyanshk: Let's keep a constant as a default ? I don't want most users thinking too much about this when using the `Shuffler`.

torchdata/nodes/header.py
divyanshk: Sounds good!

test/nodes/test_cycler.py
```diff
@@ -0,0 +1,136 @@
...
+            next(node)
+
+        # Verify cycles counter after one cycle
+        self.assertEqual(node._num_cycles, 0)
+
+        next(node)
+        self.assertEqual(node._num_cycles, 1)
+
+    @parameterized.expand([[0]])  # Simplified to just one test case
+    def test_save_load_state(self, midpoint: int) -> None:
```
divyanshk, line 117: Can we also include one test with the `run_test_save_load_state` utility like in the unit test of header ? also add a `Prefetcher` after `Cycler` in the test

test/nodes/test_shuffler.py
```diff
@@ -0,0 +1,168 @@
...
+        self.assertEqual(results, [])
+
+        # Verify yielded counter with empty source
+        self.assertEqual(node._num_yielded, 0)
+
+    @parameterized.expand(itertools.product([0, 3, 7]))
+    def test_save_load_state(self, midpoint: int) -> None:
+        # This test is now expected to fail since we don't save the buffer
+        # in the state, which changes the behavior after loading state
+        pass
```
divyanshk, line 88: Probably not a good idea to include the test if it is meant to fail.

Regarding restoring state, since we store `RNG_STATE_KEY` in state, we should be able to restore where we left off right? 

We should restore state else it might make the entire pipeline which has a `Shuffler` hard to use.

divyanshk:
@keunwoochoi Thanks for the reminder, and for waiting. 

This is looking good. Left a minor comment on adding a default value for shuffler buffer size, and a minor-ish comment on restoring state for `shuffler` using rng state. Thanks.

torchdata/nodes/shuffler.py
keunwoochoi: @divyanshk what would be a good constant? being someone who are not sure about having a default value, i'm prob not the best person to decide it.

keunwoochoi:
@divyanshk thanks for the review again! 

mostly done if not all except having a default value the shuffle buffer. but i'm sitting in a flight that is about to depart, i may wanna have another look and/or you can have a look too if you have some time ;)

keunwoochoi:
reviewed it again. looks good to me, asking for hopefully the last review :)

torchdata/nodes/shuffler.py
divyanshk: Let's do 1000.

torchdata/nodes/shuffler.py
```diff
@@ -0,0 +1,118 @@
...
+            True if the buffer contains any items after the call, False if the buffer is empty.
+        """
+        if len(self.buffer) >= self.buffer_size:
+            return True  # Buffer is already full
+
+        try:
+            items_added = 0
+            while len(self.buffer) < self.buffer_size:
+                self.buffer.append(next(self.source))
+                items_added += 1
```
divyanshk, line 76: Nit: items_added can be omitted.

test/nodes/test_shuffler.py
```diff
@@ -0,0 +1,231 @@
...
+        except StopIteration:
+            pass
+
+        # Combined sequence should match the expected sequence
+        combined = items_before + items_after
+
+        # The combined sequence will have fewer items than expected because
+        # we don't preserve the buffer in the state. We expect to lose
+        # approximately buffer_size items.
+        buffer_size = 5
```
divyanshk, line 124: I am missing something. Can you explain this ?

We have to ensure there is no drop in items on restoration.

divyanshk: Do we need a "fast forward" mode on loading state to get us to return the right elements back?

test/nodes/test_shuffler.py
```diff
@@ -0,0 +1,231 @@
...
+        for _ in range(midpoint):
+            try:
+                items_before.append(next(node))
+            except StopIteration:
+                break
+
+        # Save state and create a new node
+        state = node.state_dict()
+        new_source = StatefulRangeNode(n=n)
+        new_node = Shuffler(new_source, buffer_size=5, seed=42)
```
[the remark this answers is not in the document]
ramanishsingh: I agree. 
We can add a check if shuffle params are the same or not when we try to reload.

test/nodes/test_shuffler.py
ramanishsingh: Agreed. 
Intended behavior should be no difference if yielded all 20 in one run vs a run with interruptions.
A simple but inefficient solution would be to reset the source and fast forward the yielded items. 
Another one (which you might have already explored) would be to store the buffer as it is. 

These solutions both have their drawbacks. 
1. If we reset the sources and fast forward, then we are paying more cost when we restart, but we ONLY pay the cost iff we restart.
2. On the other hand, if we save the buffer we pay the cost every time we do .get_state() (some people might do it every step). 

Maybe we give both options to the user (might create confusion) and let them choose.

test/nodes/test_shuffler.py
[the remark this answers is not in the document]
keunwoochoi: yeah i'm also 50:50. in some sense, i might wish i could just resume with a different shuffle buffer size. but perhaps that wish should be realized by other, general solution e.g., by adding some method to ignore the data loader states. 

adding `BUFFER_SIZE_KEY` in the new commit.

test/nodes/test_shuffler.py
keunwoochoi: > Intended behavior should be no difference if yielded all 20 in one run vs a run with interruptions.
+1

but overall, this is a tricky problem. i can totally imagine users appreciate a non-stateful shuffler for efficiency. i'm actually down to excluding shuffler in this PR and finish other nodes first while discussing this. what do you think?

keunwoochoi:
just removed Shuffler from this PR, partly because i'm getting busier these days while i still would like to get this PR merged sooner than later.

test/nodes/test_shuffler.py
ramanishsingh: > . i'm actually down to excluding shuffler in this PR and finish other nodes first while discussing this. what do you think?
Yes, that will be easier and cleaner. Thanks for doing that!

divyanshk (approved):
Looks good to me! Thank you for this PR!

keunwoochoi:
nice! i'm not familiar with the further CI tests but please let me know if there's anything i can do.

ramanishsingh:
Some CI tests are failing due to an issue which are fixed in #1477. Your PR is fine. :)

PR #1464, opened 2025-03-15
Minor refactor in map

divyanshk:
Simple change initializing sort thread and queue only if in-order is needed.


ramanishsingh (approved):
Looks good. Good catch for efficiency.

PR #1465, opened 2025-03-19
Fixing docstrings, add thread names to help in debugging, add a missing stop check

divyanshk:
Title says it all ¯\_(ツ)_/¯


torchdata/nodes/map.py
```diff
@@ -231,10 +238,10 @@ def __iter__(self) -> Iterator[T]:
 
     def __next__(self) -> T:
         while True:
-            if self._stop.is_set():
+            if self._stop.is_set() or self._mp_stop.is_set():
```
ramanishsingh, line 241: Good catch!

ramanishsingh (approved):
LGTM! Thanks for adding more info to the threads for easier debugging.

PR #1469, opened 2025-03-25
Replace launching explicit threads with a threadpoolexecutor

divyanshk:
## Context
Current `ParallelMapper` has the following model:  

Arrows and dotted line represent the main process
 ---->----(spawn a background read_thread)---->---(start explicit transformation threads)---->---(spawn a background sort_thread)---->----

## Change
We are updating that to be:
 ---->----(spawn a background read_thread)  ---->----(start transformation threads within a threadpoolexecutor)---->----(spawn a background sort_thread) ---->----

Here we do not update the `method=process` workflow.

### Edit
Unit tests looks good on local. Tested GPU unit tests internally. 

Benchmarked thread pool vs explicit thread creation for transformation, almost identical performance in both settings.

### DIFF
--- a/test/nodes/test_map.py
+++ b/test/nodes/test_map.py
@@ -48,12 +48,12 @@ class TestMap(TestCase):
     def test_exception_handling_mapper_cuda(self):
         self._test_exception_handling_mapper(True, "thread")
 
-    def test_exception_handling_mapper_multiprocess(self):
-        self._test_exception_handling_mapper(False, "process")
+    # def test_exception_handling_mapper_multiprocess(self):
+    #     self._test_exception_handling_mapper(False, "process")
 
-    @unittest.skipIf(not TEST_CUDA, "CUDA not found")
-    def test_exception_handling_mapper_multiprocess_cuda(self):
-        self._test_exception_handling_mapper(True, "process")
+    # @unittest.skipIf(not TEST_CUDA, "CUDA not found")
+    # def test_exception_handling_mapper_multiprocess_cuda(self):
+    #     self._test_exception_handling_mapper(True, "process")
 
     def _test_map(self, in_order, method, prebatch) -> None:
         batch_size = 6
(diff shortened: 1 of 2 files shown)


scotts:
This makes a lot of sense to me - this may also help mitigate the daemon thread problem, as there are less threads that we explicitly manage.

test/nodes/test_map.py
```diff
@@ -179,3 +187,56 @@ def test_save_load_state_process(
         )
         node = Prefetcher(node, prefetch_factor=2)
         run_test_save_load_state(self, node, midpoint)
+
```
ramanishsingh, line 190: Can we add a test with multiple epochs (maybe with a `Loader`)? Just to make sure things work fine on epoch boundaries.

torchdata/nodes/map.py
```diff
@@ -186,40 +211,49 @@ def __init__(
...
+            for t in self._workers:
+                t.start()
 
         self._out_q = self._intermed_q
         if self.in_order:
             self._sort_q: queue.Queue = queue.Queue()
             self._sort_thread = threading.Thread(
                 target=_sort_worker,
-                args=(self._intermed_q, self._sort_q, self._stop),
+                args=(
```
ramanishsingh, line 247: This might be a question for the next PR. Do we want to replace all thread creations with TPE? Will it help mitigate the daemon thread problem?

divyanshk: Right. In this PR i am only replacing explicit threads for transformations, keeping the reading thread and sort thread as is.

test/nodes/test_map.py
divyanshk: Sure thing!

ramanishsingh:
Overall LGTM!
Thanks for the migration from individual threads to TPE.
Apart from better stability, do you know if there is any effect on performance ( ditto ing Scott's question).

test/nodes/test_map.py
divyanshk: @ramanishsingh Actually line 189 with `run_test_save_load_state` should handle the multiple epoch unit test with a `Loader`. Is that what you had in mind?

ramanishsingh: Yeah, that makes sense. 
That test actually covers what I was thinking. cool1

ramanishsingh (approved):

PR #1476, opened 2025-04-18
Maintain and check dataset keys in MultiNodeRoundRobinSampler

ramanishsingh:
Currently MultiNodeRoundRobinSampler takes the source nodes as a dict. We do not save the keys to the datasets in our state dict. This can lead to inconsistencies when reloading state dict (for e.g., if someone starts with 4 datasets but then only re-instantiates with 3). 

Adding following things
1. We keep the keys  to the datasets provided in MDS RR sampler and then save them in state dict. On `reset` we make sure the keys are the same and in order.
2. Also allowing the users to provide the dataset nodes in a list. This simplifies the process for them. However, it can cause issues if the users are not careful about providing the same order of datasets when re-instantiates and loading state_dict. For that, we log a warning that they need to provide the datasets in the same order.

In the next PR, will be adding an option to get dataset_key in the output item.

### DIFF
--- a/torchdata/nodes/samplers/multi_node_round_robin_sampler.py
+++ b/torchdata/nodes/samplers/multi_node_round_robin_sampler.py
@@ -68,7 +68,7 @@ class MultiNodeRoundRobinSampler(BaseNode[T]):
             )
             source_nodes = {f"ds_{i}": node for i, node in enumerate(source_nodes)}
 
-        self.dataset_keys = sorted(source_nodes.keys())
+        self.dataset_keys = list(source_nodes.keys())
         self.source_nodes = [source_nodes[k] for k in self.dataset_keys]
         self.num_datasets = len(self.source_nodes)
         self.stop_criteria = stop_criteria
--- a/torchdata/nodes/samplers/multi_node_round_robin_sampler.py
+++ b/torchdata/nodes/samplers/multi_node_round_robin_sampler.py
@@ -52,7 +52,7 @@ class MultiNodeRoundRobinSampler(BaseNode[T]):
     DATASET_KEYS = "dataset_keys"
     DATASET_NODE_STATES_KEY = "dataset_node_states"
     DATASETS_EXHAUSTED_KEY = "datasets_exhausted"
-    
+
     def __init__(
         self,
         source_nodes: Mapping[str, BaseNode[T]],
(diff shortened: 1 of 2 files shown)


torchdata/nodes/samplers/multi_node_round_robin_sampler.py
```diff
@@ -49,16 +52,24 @@ class MultiNodeRoundRobinSampler(BaseNode[T]):
...
     ) -> None:
         super().__init__()
-        self.source_nodes = [source_nodes[k] for k in source_nodes.keys()]
+        if isinstance(source_nodes, list):
+            logger.warning(
+                "source_nodes are provided as a list. Thus, if resetting the sampler with an initial_state, please make sure that the order of the source_nodes is same as in the previous state."
+            )
+            source_nodes = {f"ds_{i}": node for i, node in enumerate(source_nodes)}
+
+        self.dataset_keys = sorted(source_nodes.keys())
```
divyanshk, line 71: Just to double check, the sorting will not change affect the ordering of how the data is spit out right ? i.e. we don't want the round robin to work by going over the sorted order

divyanshk: Why do we sort the keys ? Ideally the list of keys and their ordering is also important.

torchdata/nodes/samplers/multi_node_round_robin_sampler.py
```diff
@@ -75,9 +86,16 @@ def _validate_stop_criteria(self) -> None:
                 f"Invalid {self.stop_criteria=}. stop_criteria must be one of: CYCLE_UNTIL_ALL_DATASETS_EXHAUSTED, ALL_DATASETS_EXHAUSTED, FIRST_DATASET_EXHAUSTED"
             )
 
+    def _validate_reset_dataset_keys(self, reset_keys) -> None:
```
divyanshk, line 89: If we remove the sorting of keys, then this function will also catch the case where someone flips the order of two datasets accidently.

torchdata/nodes/samplers/multi_node_round_robin_sampler.py
ramanishsingh: Thanks for the question. It raises a good point that I missed. 
In Python 3.7+, we can rely on the dict to provide us the items in the order in which they were inserted. 
I am sorting the keys, and it will change the order in some cases, which is not desirable. Fixing this.

torchdata/nodes/samplers/multi_node_round_robin_sampler.py
ramanishsingh: Thanks. Fixed in the latest commit.

divyanshk (approved):
Looks good to me!

PR #1477, opened 2025-04-18
add explicit shutdown calls

divyanshk:
@ramanishsingh helped identify some Github UI hanging in older python versions. This diff adds explicit shutdown to ensure thread pool executor doesn't stay open even after all threads are completed.


ramanishsingh:
Thanks for fixing. LGTM!

PR #1479, opened 2025-04-29
Give an option to either provide dataset or dataset_size in distributed sampler

ramanishsingh:
Currently the StatefulDistributedSampler takes the dataset as an argument, but only uses the length/size of the dataset. 
Adding an option to provide size of the dataset instead, for more flexibility.

### DIFF
--- a/torchdata/stateful_dataloader/sampler.py
+++ b/torchdata/stateful_dataloader/sampler.py
@@ -5,7 +5,10 @@
 # LICENSE file in the root directory of this source tree.
 
 import itertools
+import math
 from typing import Any, Dict, Iterator, List, Optional, Sized
+
+import torch.distributed as dist
 
 import torch.utils.data.sampler
 from torch.utils.data import Dataset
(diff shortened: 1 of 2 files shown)


torchdata/stateful_dataloader/sampler.py
```diff
@@ -179,19 +182,66 @@ def __iter__(self):
         )
 
 
-class StatefulDistributedSampler(torch.utils.data.distributed.DistributedSampler):
+class StatefulDistributedSampler(Sampler[int]):
```
divyanshk, line 185: I think we should continue subclassing `DistributedSampler` for `StatefulDistributedSampler` - it is easy to udnerstand that by just the naming and we might trigger many type checking issues in downstream code which uses `StatefulDistributedSampler` and expects a variant of `DistributedSampler`.  

Since `DistributedSampler` is a common utility in PyTorch, `StatefulDistributedSampler` should be expected to be an extension of it.

divyanshk (changes requested):
Lets sync if there is a better way to do this - maybe create a new sampler?

torchdata/stateful_dataloader/sampler.py
ramanishsingh: I decided to fork it instead of subclassing because I do not want to upstream these changes in `torch.utils.data.distributed.DistributedSampler` as it might break other users' code. 
Nevertheless, it is redundant to have Dataset as an arg when we just need the length of it.

PR #1482, opened 2025-05-02
Explicit iterator shutdown

nvski:
## Bug

Currently resetting `ParallelMap` kills threads with `del self._it`, which is not guaranteed to immediately call `self._it.__del__` and thus `._shutdown` method. It leads to an extra living read thread. This extra thread can consume extra items from `self.source` node, which will be missing from new `self._it`

I faced this issue with `lightning.pytorch` training. I could not figure out, if lightning just calls `iter` twice in a short time, or increases ref count of the old iterator object somehow, but as a matter of fact I have two running read threads, which breaks my pipeline at the first batch.

## Fix
Add explicit `self._it._shutdown()` into `reset` method. This fixes described bug

### DIFF
--- a/torchdata/nodes/map.py
+++ b/torchdata/nodes/map.py
@@ -377,6 +377,7 @@ class _ParallelMapperImpl(BaseNode[T]):
     def reset(self, initial_state: Optional[Dict[str, Any]] = None):
         super().reset(initial_state)
         if self._it is not None:
+            self._it._shutdown()
             del self._it
 
         if self.num_workers > 0:


divyanshk:
Kicking off CI - and I will also test this out! Overall this seems a good change overall.

nvski:
@divyanshk 
Setting switching from `v0.11.0` to `main` (with #1477 already merged) and setting `daemonic_reading=False` does not help

How I noticed this race condition: my pipeline expects groups of frames from the same video, and fails if frames are not consecutive. Debug job stopped at the exception before the older thread joined and I noticed extra `_populate_queue` thread which iterates through the same `self.source` concurrently

I invesigated the problem and yes, lightning calls `iter` twice in `_FitLoop.setup_data`: [first time](https://github.com/Lightning-AI/pytorch-lightning/blob/6da480d5e3304b91ac188b849ee06e369e878460/src/lightning/pytorch/loops/fit_loop.py#L252) it just checks that provided dataloader is an iterable (not fetching any data) and [then](https://github.com/Lightning-AI/pytorch-lightning/blob/6da480d5e3304b91ac188b849ee06e369e878460/src/lightning/pytorch/loops/fit_loop.py#L275) iterating once more creating their own data fetcher

There are related lightning issues and PRs

https://github.com/Lightning-AI/pytorch-lightning/issues/19427
https://github.com/Lightning-AI/pytorch-lightning/pull/20775
https://github.com/Lightning-AI/pytorch-lightning/pull/20705

It also might be a good idea to add a flag in `tn.Loader`, which will detect that no elements were fetched since the previous `iter` call and skip `reset`? Something like [this draft](https://github.com/pytorch/data/compare/main...nvski:data:patch-2). Even with race condition solved, this reiteration might cause skipping epochs or instant restarts of freshly loaded state dicts

divyanshk (approved):
This change looks good to me!

PR #1488, opened 2025-05-14
remove failing ninja-build

divyanshk:
Testing if this solves the broken CI, and what other ramifications it will have.


ramanishsingh (approved):
Thanks for fixing the failing test issue

PR #1498, opened 2025-07-17
Adds `FileLister`

keunwoochoi:
this is a utility node that list files at a specific location (local path or remote path as supported by `fsspec`). we have been using a very similar node actively through our use-cases to load files from s3 that i'd definitely call it a success. i'm opening this PR to get the maintainer's thought about adding it.

### DIFF
--- a/test/nodes/io/test_file_list.py
+++ b/test/nodes/io/test_file_list.py
@@ -314,10 +314,12 @@ def test_file_list_state_validation_index_out_of_bounds(sample_local_files):
 
     # Create a FileLister and get initial state
     node = FileLister(tmp_dir, ["**/*"])
+    # Populate file listing
+    list(node)
     initial_state = node.get_state()
 
-    # Modify the state to have invalid index
-    corrupted_state = initial_state.copy()
+    # Modify the state to have invalid index but keep num_files matching
+    corrupted_state = {**initial_state}
     corrupted_state["current_idx"] = initial_state["num_files"] + 10  # Way out of bounds
 
     # Create new node and try to restore corrupted state
@@ -342,6 +344,7 @@ def test_file_list_protocol_handling():
         # Should not add protocol prefix for local files
         file_paths = [item["data"] for item in node]
         assert all(not path.startswith("file://") for path in file_paths)
+        assert all("\\" not in path for path in file_paths)
         assert all("/" in path for path in file_paths)  # Should have forward slashes
 
 
@@ -459,12 +462,15 @@ def test_file_list_path_normalization():
         assert all(path.startswith("C:/") for path in file_paths)
 
 
-def test_file_list_state_consistency():
+def test_file_list_state_consistency(sample_local_files):
     """Test that state is consistent across multiple operations."""
     tmp_dir, _ = sample_local_files
 
     # Create a FileLister
     node = FileLister(tmp_dir, ["**/*"])
+
+    # Initialize without consuming items so state has correct num_files
+    node.reset()
 
     # Get initial state
     initial_state = node.get_state()
@@ -487,7 +493,7 @@ def test_file_list_state_consistency():
     assert updated_idx == initial_idx + 3
 
 
-def test_file_list_reset_without_state():
+def test_file_list_reset_without_state(sample_local_files):
     """Test that reset without state works correctly."""
     tmp_dir, _ = sample_local_files
 
--- a/torchdata/nodes/io/file_list.py
+++ b/torchdata/nodes/io/file_list.py
@@ -86,7 +86,7 @@ class FileLister(BaseNode[Dict]):
         self.fs = fsspec.filesystem(self.protocol, **self.fs_options)
 
         # Initialize state
-        self._file_paths = []
+        self._file_paths: List[str] = []
         self._current_idx = 0
 
     def reset(self, initial_state: Optional[Dict[str, Any]] = None):
@@ -108,7 +108,17 @@ class FileLister(BaseNode[Dict]):
                     if pattern == "**/*" or pattern == "*":
                         matches = [match for match in matches if self.fs.isfile(match)]
 
-                    file_paths_set.update(matches)
+                    # For local filesystem, ensure matches include the base path when relative
+                    normalized_matches: List[str] = []
+                    for match in matches:
+                        candidate = match
+                        if self.protocol == "file":
+                            has_scheme = "://" in match
+                            if not has_scheme and not os.path.isabs(match) and not match.startswith(self.uri_base):
+                                candidate = os.path.join(self.uri_base, match)
+                        normalized_matches.append(candidate)
+
+                    file_paths_set.update(normalized_matches)
 
                 except Exception as e:
                     logger.warning(f"Error globbing pattern {pattern}: {e}")


requirements.txt
```diff
@@ -1,2 +1,3 @@
 urllib3 >= 1.25
 requests
+fsspec
```
ramanishsingh, line 3: Is it fine to take dependency on fsspec? Is it a commonly used/supported package?

torchdata/nodes/io/file_list.py
```diff
@@ -0,0 +1,138 @@
...
+                    # Filter out directories from matches if pattern is **/* or *
+                    if pattern == "**/*" or pattern == "*":
+                        matches = [match for match in matches if self.fs.isfile(match)]
+
+                    self._file_paths.update(matches)
+
+                except Exception as e:
+                    logger.warning(f"Error globbing pattern {pattern}: {e}")
+
+            # Convert to sorted list for deterministic ordering
```
ramanishsingh, line 103: What if the file paths or num_files change between runs?
Will reset still work well? 

1. Should we also save num_files in the state to see if num_files match or not?
2. How about file_paths, will that be too expensive?

torchdata/nodes/io/file_list.py
```diff
@@ -0,0 +1,138 @@
...
+        >>> # List GCS files
+        >>> node = FileLister("gs://my-bucket/prefix", ["*.parquet"])
+        >>> for item in node:
+        ...     print(item["data"])  # prints "gs://my-bucket/prefix/file1.parquet", etc.
+    """
+
+    SOURCE_KEY = "source"
+    CURRENT_IDX_KEY = "current_idx"
+
+    def __init__(self, uri_base: str, patterns: List[str], fs_options: Optional[Dict[str, Any]] = None):
```
ramanishsingh, line 53: do you think it will be useful to have something like max_files.
That a person just wants to read x number of files because they might not know how many total are there

torchdata/nodes/io/file_list.py
```diff
@@ -0,0 +1,138 @@
...
+        # Always initialize files list
+        self._file_paths = set()
+
+        try:
+            # Use fsspec's glob for file discovery
+            for pattern in self.patterns:
+                pattern_path = os.path.join(self.uri_base, pattern)
+
+                try:
+                    matches = self.fs.glob(pattern_path)
```
ramanishsingh, line 92: This might be out of the scope of this PR, just bouncing ideas
Is there a way to limit the recursive depth of the search?

torchdata/nodes/io/file_list.py
```diff
@@ -0,0 +1,138 @@
...
+
+                except Exception as e:
+                    logger.warning(f"Error globbing pattern {pattern}: {e}")
+
+            # Convert to sorted list for deterministic ordering
+            self._file_paths = sorted(self._file_paths)
+
+            # For non-file protocols, ensure URIs are properly formatted
+            if self.protocol != "file":
+                self._file_paths = [f"{self.protocol}://{path}" for path in self._file_paths]
```
ramanishsingh, line 108: qq (havent worked with cloud storage), is it guaranteed that the path always not have the protocal prefix (s3:// etc.)? or do we need to add a check of that

torchdata/nodes/io/file_list.py
```diff
@@ -0,0 +1,138 @@
...
+
+        except Exception as e:
+            logger.error(f"Error listing files from {self.uri_base}: {e}")
+            self._file_paths = []
+
+        self._current_idx = 0
+
+        # If there's an initial state, restore position
+        if initial_state:
+            self._current_idx = initial_state.get(self.CURRENT_IDX_KEY, 0)
```
ramanishsingh, line 118: shall we do a check that if this is a valid state, i.e. _current_idx is not > len(_file_paths)

torchdata/nodes/io/file_list.py
```diff
@@ -0,0 +1,138 @@
...
+        try:
+            # Use fsspec's glob for file discovery
+            for pattern in self.patterns:
+                pattern_path = os.path.join(self.uri_base, pattern)
+
+                try:
+                    matches = self.fs.glob(pattern_path)
+
+                    # Filter out directories from matches if pattern is **/* or *
+                    if pattern == "**/*" or pattern == "*":
```
ramanishsingh, line 95: what if the pattern is "test_files/*"
will it also return directories? Do we need to filter them?

test/nodes/io/test_file_list.py
```diff
@@ -0,0 +1,258 @@
...
+    mock_fs = MagicMock()
+    mock_filesystem.return_value = mock_fs
+    mock_fs.glob.side_effect = Exception("S3 Error")
+
+    # Create FileLister for S3
+    node = FileLister("s3://my-bucket", ["*.txt"])
+
+    # Should not raise exception, but log error and return empty list
+    file_items = list(node)
+    assert len(file_items) == 0
```
ramanishsingh, line 258: Can we add some more tests
1. Check if the state is invalid (Something changed, more/less num_files)
2. Double protocol name prefix in files test (depending on: does fsspec always review s3:// when returning file names?)
3. Filtering directories

ramanishsingh (commented):
Thanks for the contribution Keunwoo
I have some questions and some suggestions. Please let me know what you think.

torchdata/nodes/io/file_list.py
keunwoochoi: for future reference: yes. 
https://filesystem-spec.readthedocs.io/en/latest/api.html

> maxdepth: int or None
> Maximum depth for '**' patterns. Applied on the first '**' found. Must be at least 1 if provided.

torchdata/nodes/io/file_list.py
keunwoochoi: it will return files only. actually the next line is the filter for dirs.

`matches = [match for match in matches if self.fs.isfile(match)]`

torchdata/nodes/io/file_list.py
keunwoochoi: it may depend on how much reproducibility you would like to have in `torchdata`. i like the intermediate idea of checking `num_files` and provide warnings. or - would you like `raise` more?

although strong `file_paths` will ensure the statefulness more strongly, i think it could be too expensive & other nodes like file reader can be supplementary to achieve the goal.

torchdata/nodes/io/file_list.py
keunwoochoi: it seems so. on s3, yes it does. just checked with google cloud:
```

>>> matches = fs.glob("gs://tfds-data/datasets/nsynth/full/2.0.0/diffs/*")
>>> matches
['tfds-data/datasets/nsynth/full/2.0.0/diffs/dataset_info.json', 'tfds-data/datasets/nsynth/full/2.0.0/diffs/diff_report.txt', 'tfds-data/datasets/nsynth/full/2.0.0/diffs/nsynth-test.counts.txt-00000-of-00001', 'tfds-data/datasets/nsynth/full/2.0.0/diffs/nsynth-test.only1.tfrecord-00000-of-00003', 'tfds-data/datasets/nsynth/full/2.0.0/diffs/nsynth-test.only1.tfrecord-00001-of-00003', 'tfds-data/datasets/nsynth/full/2.0.0/diffs/nsynth-test.only1.tfrecord-00002-of-00003', 'tfds-data/datasets/nsynth/full/2.0.0/diffs/nsynth-test.only2.tfrecord-00000-of-00003', 'tfds-data/datasets/nsynth/full/2.0.0/diffs/nsynth-test.only2.tfrecord-00001-of-00003', 'tfds-data/datasets/nsynth/full/2.0.0/diffs/nsynth-test.only2.tfrecord-00002-of-00003', 'tfds-data/datasets/nsynth/full/2.0.0/diffs/nsynth-train.counts.txt-00000-of-00001', 'tfds-data/datasets/nsynth/full/2.0.0/diffs/nsynth-train.only1.tfrecord-00000-of-00001', 'tfds-data/datasets/nsynth/full/2.0.0/diffs/nsynth-train.only2.tfrecord-00000-of-00001', 'tfds-data/datasets/nsynth/full/2.0.0/diffs/nsynth-valid.counts.txt-00000-of-00001', 'tfds-data/datasets/nsynth/full/2.0.0/diffs/nsynth-valid.only1.tfrecord-00000-of-00001', 'tfds-data/datasets/nsynth/full/2.0.0/diffs/nsynth-valid.only2.tfrecord-00000-of-00001']
```

which works well.

torchdata/nodes/io/file_list.py
```diff
@@ -0,0 +1,138 @@
...
+            raise StopIteration("No more files")
+
+        path = self._file_paths[self._current_idx]
+        self._current_idx += 1
+
+        # Create metadata based on filesystem protocol
+        metadata = {"fs_protocol": self.protocol, "source": self.protocol}
+
+        return {"data": path, "metadata": metadata}
+
```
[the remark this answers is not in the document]
keunwoochoi: looked into the `fs` methods. it seems to close the session without any explicit call (we are not opening any file or so). one thing we can do is to set `self.fs = None` but would it be any necessary? or `self._file_paths = None`, `self._current_idx = 0`? 

to be clear, `fs.close_session()` method exists but it takes arguments which would be different per the protocol.

keunwoochoi:
thanks for the reviews @ramanishsingh , i left replies.

torchdata/nodes/io/file_list.py
ramanishsingh: @keunwoochoi  thanks.

I think we should raise when num_files do not match.
My reasoning is that if we do not raise an error and there is an increase in the number of files after restart, we might be reading from a different file (after sort) and the user might not even know.

After raising the error, user will know what's happening and can change their filtering logic. 
LMK what you think

keunwoochoi:
@ramanishsingh added the requested features. seems like the test is passing now too.

PR #1499, opened 2025-07-21
draft: Adds FileReader

keunwoochoi:
similar to #1498, but for file reading.

### DIFF
--- a/torchdata/nodes/io/file_read.py
+++ b/torchdata/nodes/io/file_read.py
@@ -1,0 +1,133 @@
+import logging
+from typing import Any, Dict, Optional
+
+from smart_open import open
+from torchdata.nodes import BaseNode
+
+logger = logging.getLogger(__name__)
+
+
+class FileReader(BaseNode[Dict]):
+    """Universal node that reads file contents from any filesystem supported by smart_open.
+
+    Uses smart_open to support local files, S3, GCS, Azure, HTTP, and more filesystems.
+    Works seamlessly with any filesystem that smart_open supports.
+
+    Features:
+    - Supports any filesystem that smart_open supports (local, S3, GCS, Azure, HTTP, etc.)
+    - Handles compressed files (.gz, .bz2) transparently
+    - Maintains state for checkpointing and resumption
+    - Preserves metadata from source nodes
+    - Works with both text and binary files
+
+    Output format:
+        {
+            "data": "file contents",      # File contents as str (text mode) or bytes (binary mode)
+            "metadata": {
+                "file_path": "path/to/file.txt"  # Original file path/URI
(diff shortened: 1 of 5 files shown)


torchdata/nodes/io/file_read.py
```diff
@@ -0,0 +1,133 @@
...
+            # Include metadata from source
+            if source_metadata:
+                metadata.update(source_metadata)
+
+            return {self.DATA_KEY: content, self.METADATA_KEY: metadata}
+
+        except Exception as e:
+            logger.error(f"Error reading {file_path}: {e}")
+            raise
+
```
ramanishsingh, line 130: We should also add a shutdown method.

torchdata/nodes/io/file_read.py
```diff
@@ -0,0 +1,133 @@
...
+        try:
+            # Use smart_open to handle any file system
+            with open(
+                file_path,
+                self.mode,
+                encoding=None if "b" in self.mode else self.encoding,
+                transport_params=self.transport_params,
+            ) as f:
+                content = f.read()
```
ramanishsingh, line 116: 1. Is this use case for large files
2. Does this line load all the contents at once

if yes for 1. and No for 2., then we might want to add a chunked reader or streamer.

ramanishsingh: But I think that might be out of scope for this PR and we can work on that later.

torchdata/nodes/io/file_read.py
```diff
@@ -0,0 +1,133 @@
...
+            # Create output with metadata
+            metadata = {"file_path": file_path}
+
+            # Include metadata from source
+            if source_metadata:
+                metadata.update(source_metadata)
+
+            return {self.DATA_KEY: content, self.METADATA_KEY: metadata}
+
+        except Exception as e:
```
ramanishsingh, line 127: We can add a retry logic
what if there were some network hiccups, or some other retryable error
Is there a way to find out if the error was because of network (we can retry in that case) or some other factor (file not present, etc., so do not retry)

torchdata/nodes/io/file_read.py
```diff
@@ -0,0 +1,133 @@
+import logging
+from typing import Any, Dict, Optional
+
+from smart_open import open
```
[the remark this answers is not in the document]
keunwoochoi: good point!

torchdata/nodes/io/file_read.py
keunwoochoi: it includes large files. but i have a separate piece of code that does read text file line-by-line, without using this `FileReader`. i.e. text streamer.

agree that we those would be great to have as a separate PR.

PR #1542, opened 2026-04-16
[Nodes] MapStyleWrapper exposes ParallelMapper kwargs for parallelism

### LINKED ISSUE #1387: [Nodes] MapStyleWrapper should expose ParallelMap kwargs
### 🚀 The feature

Allow MapStyleWrapper to take parallel map kwargs

### Motivation, pitch

MapStyleWrapper is a pretty handy utility but ham-strung because it doesn't not allow parallelism. We could expose these easily by allowing `**parallel_mapper_kwargs` in the constructor, and switching `Mapper` to `ParallelMapper`, with a default of num_workers == 0.

### Alternatives

_No response_

### Additional context

_No response_

psaikaushik:
Allow MapStyleWrapper to accept **parallel_mapper_kwargs forwarded to ParallelMapper, defaulting num_workers=0 for backwards compatibility. Users can now enable parallel data loading by passing num_workers > 0 (and any other ParallelMapper kwargs such as in_order, method, etc.) without having to manually compose SamplerWrapper + ParallelMapper.

Fixes #1387

Also fixes a flaky test (test_thread_pool_executor_cancel_futures_shutdown):

_ParallelMapperIter._shutdown had no idempotency guard, so pool.shutdown() could be called twice on the same instance — once explicitly with cancel_futures=False, and again when _ParallelMapperImpl.__del__ raced in and called _shutdown(cancel_futures=True) before GC finished. Since mock.patch intercepts all ThreadPoolExecutor.shutdown calls class-wide, the second call with cancel_futures=True became the last recorded call and broke the assertion.

Fixed by adding a _pool_shutdown flag to _ParallelMapperIter.__init__ so pool.shutdown() is only ever invoked once per iterator instance.

creating your pull request.

- If you are adding a new node, ensure you read that section in the contribution guide, as it includes requirements for
  functionality and testing.

Fixes #{issue number}

### Changes

-
-

### DIFF
--- a/test/nodes/test_adapters.py
+++ b/test/nodes/test_adapters.py
@@ -119,6 +119,35 @@ class TestMapStyle(TestCase):
         run_test_save_load_state(self, node, midpoint)
 
 
+class TestMapStyleParallelMapper(TestCase):
+    @parameterized.expand([1, 4])
+    def test_parallel_workers(self, num_workers: int):
+        n = 20
+        node = MapStyleWrapper(DummyMapDataset(n), sampler=range(n), num_workers=num_workers)
+        for _ in range(2):
+            node.reset()
+            result = list(node)
+            self.assertEqual(len(result), n)
+            self.assertEqual({row["step"] for row in result}, set(range(n)))
+
+    @parameterized.expand([1, 4])
+    def test_parallel_workers_in_order(self, num_workers: int):
+        n = 20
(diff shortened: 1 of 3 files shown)


divyanshk (approved):
Both changes LGTM!

Can you add a comment in the PR summary about the change in map.py shutdown, thanks.

PR #1544, opened 2026-04-16
[Nodes] Add Compose utility for chaining callables

### LINKED ISSUE #1399: [Nodes] Create a utility to chain/compose function calls
### 🚀 The feature

Standard utility to chain function calls, similar to torchvision's Compose class

### Motivation, pitch

This is going to be really useful for folks that want to fuse multiple map-operators and we shouldn't require another dependency for something so simple and useful.

### Alternatives

_No response_

### Additional context

_No response_

psaikaushik:
Adds a Compose class that applies a sequence of callables in order, allowing users to fuse multiple mapping functions into a single Mapper or ParallelMapper node rather than chaining multiple nodes.

    node = Mapper(source, Compose([normalize, augment, to_tensor]))

Closes #1399

Fix flaky test_thread_dead_error on Windows:

test_thread_dead_error creates a thread that sleeps for QUEUE_TIMEOUT * 3.0 (0.3s) and expects get_initial_snapshot to raise with thread.is_alive()=False before its timeout of QUEUE_TIMEOUT * 5.0 (0.5s). On Windows, time.sleep has ~15ms timer resolution, so each queue.get(timeout=QUEUE_TIMEOUT) call can overshoot by ~15ms. Across 4–5 loop iterations this accumulates enough to push the total elapsed time past the 0.5s timeout before the thread's is_alive() check fires — causing the error to report thread.is_alive()=True instead. Increased the timeout to QUEUE_TIMEOUT * 10.0 (1.0s), giving 3× headroom over the thread's sleep. The loop still exits early as soon as is_alive() returns False, so there's no impact on test runtime in the normal case.

### DIFF
--- a/torchdata/nodes/transforms.py
+++ b/torchdata/nodes/transforms.py
@@ -39,5 +39,5 @@ class Compose:
         return x
 
     def __repr__(self) -> str:
-        fns_str = ",\n    ".join(repr(fn) for fn in self.fns)
-        return f"Compose([\n    {fns_str}\n])"
+        fns_str = ", ".join(repr(fn) for fn in self.fns)
+        return f"Compose([{fns_str}])"
--- a/torchdata/nodes/transforms.py
+++ b/torchdata/nodes/transforms.py
@@ -1,0 +1,43 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
+# All rights reserved.
+#
+# This source code is licensed under the BSD-style license found in the
+# LICENSE file in the root directory of this source tree.
+
+from typing import Any, Callable, Sequence
+
+
+class Compose:
+    """Composes multiple callables into a single callable that applies them sequentially.
+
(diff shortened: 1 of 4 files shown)


divyanshk (approved):

torchdata/nodes/transforms.py
```diff
@@ -0,0 +1,43 @@
...
+            raise ValueError("Compose requires at least one function, got an empty sequence.")
+        self.fns = list(fns)
+
+    def __call__(self, x: Any) -> Any:
+        for fn in self.fns:
+            x = fn(x)
+        return x
+
+    def __repr__(self) -> str:
+        fns_str = ",\n    ".join(repr(fn) for fn in self.fns)
```
divyanshk: ",\n    " seems a bit rough. maybe skip the spaces and just commas with a single space should be good

psaikaushik: Good point. Fixed it. Now it should be single line. thanks @divyanshk

PR #1546, opened 2026-04-16
Nodes/get worker info parallel mapper

### LINKED ISSUE #1388: [Nodes] Ensure get_worker_info works correctly in ParallelMapper
### 🚀 The feature

Support for torch.utils.data.IterableDataset

### Motivation, pitch

Currently IterableDataset (and possibly some map datasets) rely on torch.utils.data.get_worker_info. To ensure drop-in compatibility, we should make sure this works correctly for ParallelMapper by setting the environment up correctly. We could do this directly in ParallelMapper, or in a new IterableDatasetWrapper class.

### Alternatives

_No response_

### Additional context

_No response_

psaikaushik:
## Summary
Adds num_workers parameter to _apply_udf and sets a WorkerInfo object on each worker before the processing loop begins, so user-defined functions can identify which worker they are running in.
Introduces torchdata.nodes.get_worker_info() backed by threading.local() — always returns the correct WorkerInfo for the calling worker regardless of thread vs. process mode.
For process workers, torch.utils.data.get_worker_info() also works correctly (each process has isolated memory).
For thread workers, torchdata.nodes.get_worker_info() is the safe API; torch.utils.data.get_worker_info() may race between threads so the thread-local version is preferred.
Exports get_worker_info from torchdata.nodes.

Fixes #1388

## Test plan
 test_none_outside_worker — returns None when called from outside any worker
 test_thread_workers — all items report a valid worker_id ∈ [0, num_workers) and correct num_workers
 test_thread_workers_each_id_used — all worker ids appear across 40 items with 4 workers
 test_thread_workers_unique_seeds — each worker reports one consistent seed; seeds differ across workers
 test_process_workers_get_worker_info — torch.utils.data.get_worker_info() works in forkserver process workers
 test_process_workers_torchdata_get_worker_info — torchdata.nodes.get_worker_info() works in process workers
 test_num_workers_zero_no_worker_info — inline (num_workers=0) UDF sees None


torchdata/nodes/_apply_udf.py
```diff
@@ -7,30 +7,71 @@
...
-
+import torch.utils.data._utils.worker as _worker_module  # type: ignore[import]
 from torch._utils import ExceptionWrapper
 
 from .constants import QUEUE_TIMEOUT
 
+_thread_local = threading.local()
+
+
+def get_worker_info() -> Optional[Any]:
```
divyanshk: This shouldn't live inside _apply_udf. maybe a utils file in nodes/ ?

torchdata/nodes/_apply_udf.py
```diff
@@ -7,30 +7,71 @@
...
     """
     torch.set_num_threads(1)
+    seed = torch.initial_seed() + worker_id
+    worker_info = _worker_module.WorkerInfo(id=worker_id, num_workers=num_workers, seed=seed, dataset=None)  # type: ignore[attr-defined,arg-type]
+    # Thread-local: always returns the correct info for this worker, regardless of
+    # whether other workers (threads) have set their own worker info concurrently.
+    _thread_local.worker_info = worker_info
+    # Module-level global: correct for process workers (isolated memory); for thread
+    # workers this may race, so callers should use torchdata.nodes.get_worker_info().
+    _worker_module._worker_info = worker_info  # type: ignore[attr-defined]
```
divyanshk: I am not sure of this change. Overall it is safe, but I don't think its very clean to have this inside _apply_udf. Plus we are leaving dataset=None.

divyanshk (changes requested):

torchdata/nodes/_apply_udf.py
psaikaushik: Good point. Moved `get_worker_info()`, the thread-local storage, and the `_set_worker_info()` helper into a dedicated `torchdata/nodes/_worker_info.py`. `_apply_udf.py` now just calls `_set_worker_info(worker_id, num_workers)` and has no worker-info logic of its own.

torchdata/nodes/_apply_udf.py
psaikaushik: Agreed on the placement, addressed in the same commit by moving everything to ``_worker_info.py.`

On `dataset=None`: ParallelMapper maps over arbitrary items rather than a PyTorch Dataset object, so there is no meaningful dataset to pass. We still construct a torch.utils.data.WorkerInfo (rather than a custom dataclass) so that torch.utils.data.get_worker_info() continues to work inside process workers, keeping compatibility with existing IterableDataset code that calls it. The # type: ignore[arg-type] on that line suppresses the mypy complaint. Happy to discuss if there's a cleaner alternative you have in mind.

psaikaushik (commented):
Addressed the comments. Please take a look @divyanshk

PR #1550, opened 2026-06-08
[BUG] Prefetcher is unrecoverable from exceptions in reset()

alexdremov:
If `_SingleThreadedMapper` fails for whatever reason, `Prefetcher` ends up in inconsistent state where `self._it` is deleted.

### DIFF
--- a/torchdata/nodes/prefetch.py
+++ b/torchdata/nodes/prefetch.py
@@ -34,9 +34,11 @@ class Prefetcher(BaseNode[T]):
 
     def reset(self, initial_state: Optional[Dict[str, Any]] = None):
         super().reset(initial_state)
-        if self._it is not None:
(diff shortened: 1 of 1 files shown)


alexdremov:
@ramanishsingh @divyanshk please take a look. This is an annoying bug that does not allow to recover from problems of state restoration.

alexdremov:
@ramanishsingh @divyanshk ping

sylvesterkaczmarek (commented):
Could this recovery guard be applied to `PinMemory.reset()` and `_ParallelMapper.reset()` too? Both still `_shutdown(); del self._it` before constructing the replacement, so if construction/reset throws they leave `_it` absent and the next reset hits the same `AttributeError` this PR fixes for `Prefetcher`.

PR #1552, opened 2026-08-23
Fix Loader restoration before None items

### LINKED ISSUE #1551: Loader checkpoint restoration restarts before a None item
### Bug

`torchdata.nodes.Loader` cannot restore a checkpoint whose next item is `None` when `restart_on_stop_iteration=True` (the default). The look-ahead in `LoaderIterator.has_next()` treats `None` as meaning that no item was cached, decides the restored iterator is exhausted, and restarts it from the beginning.

### Reproduction

```python
from torchdata.nodes import IterableWrapper, Loader

loader = Loader(IterableWrapper([1, None, 2]))
iterator = iter(loader)
assert next(iterator) == 1
state = loader.state_dict()

restored = Loader(IterableWrapper([1, None, 2]))
restored.load_state_dict(state)
print(list(restored))
```

On current `main` (`e640e6f`), this prints:

```text
[1, None, 2]
```

The expected continuation is:

```text
[None, 2]
```

The restart duplicates data that was consumed before the checkpoint. Calling `LoaderIterator.has_next()` directly also returns `False` and consumes the value when the next item is `None`.

### Root cause and proposed fix

`LoaderIterator._cached_item` currently uses `None` both as a valid data value and as its empty-cache sentinel. Tracking cache occupancy separately avoids that collision without changing the public API or the look-ahead behavior for other values.

I have a focused regression test and a small fix ready.

aswanth-07:
Fixes #1551.

### Summary

`LoaderIterator` used `None` both as a valid data value and as the empty look-ahead marker. When a checkpoint resumed immediately before a `None` item, `Loader.__iter__()` called `has_next()`, misclassified the item as end-of-input, and reset the pipeline to the beginning. This duplicated all data consumed before the checkpoint.

This change tracks cache occupancy separately from the cached value. `None` can therefore be cached and returned like any other item, while the existing look-ahead and restart behavior remains unchanged for exhausted iterators.

Reset now also clears the cached prefetch state dictionary together with the cached item. This prevents `state_dict()` from reporting a position captured before a later `load_state_dict()` and iterator reset.

The regressions verify that restoring `[0, None, 2]` after consuming `0` yields exactly `[None, 2]`, and that a state request after `has_next()`, `load_state_dict()`, and `iter(loader)` reports the restored position.

### Validation

- `PYTEST_DISABLE_PLUGIN_AUTOLOAD=1 py -3.13 -m pytest -q test/nodes/test_loader.py` (5 passed)
- `py -3.13 -m pre_commit run --files torchdata/nodes/loader.py test/nodes/test_loader.py` (passed)
- Direct public-flow probe: the stale-state regression failed before the reset fix with positions 0 and 1, then passed after the fix.
- `git diff --check` (passed)
- Standalone mypy was not used as a final signal because the installed version rejects the repository's quoted `python_version` setting and this source checkout lacks the generated `torchdata.version` module.

AI assistance disclosure: AI-assisted development tools were used during investigation and implementation. The reported tests were run against the final diff.

### DIFF
--- a/test/nodes/test_loader.py
+++ b/test/nodes/test_loader.py
@@ -53,3 +53,14 @@ class TestLoader(TestCase):
         length = 10
         node = MapStyleWrapper(DummyMapDataset(length), sampler=range(length))
         self._test_loader_correct_state_dict_at_midpoint(node, length)
+
+    def test_loader_restores_before_none_item(self) -> None:
+        loader = Loader(IterableWrapper([0, None, 2]))
+        iterator = iter(loader)
+        self.assertEqual(next(iterator), 0)
+        state_dict = loader.state_dict()
+
+        restored = Loader(IterableWrapper([0, None, 2]))
+        restored.load_state_dict(state_dict)
+
+        self.assertEqual(list(restored), [None, 2])
--- a/test/nodes/test_loader.py
+++ b/test/nodes/test_loader.py
@@ -64,3 +64,15 @@ class TestLoader(TestCase):
         restored.load_state_dict(state_dict)
 
         self.assertEqual(list(restored), [None, 2])
+
+    def test_loader_reset_clears_cached_state_dict(self) -> None:
+        checkpoint_source = Loader(IterableWrapper([0, 1, 2]))
+        self.assertEqual(next(iter(checkpoint_source)), 0)
+        state_dict = checkpoint_source.state_dict()
+
+        loader = Loader(IterableWrapper([0, 1, 2]))
+        self.assertTrue(iter(loader).has_next())
+        loader.load_state_dict(state_dict)
+        iter(loader)
+
+        self.assertEqual(loader.state_dict(), state_dict)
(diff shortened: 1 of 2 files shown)


sylvesterkaczmarek (commented):
One adjacent cache survives reset: `_cached_state_dict` isn't cleared. If `has_next()` populated it, then `load_state_dict()` + `iter(loader)` resets the root and clears `_cached_item`/`_has_cached_item`, but a subsequent `state_dict()` still returns the pre-reset cached state. Should `reset()` clear `_cached_state_dict` as well?

sylvesterkaczmarek (approved):
The reset path now clears `_cached_state_dict` together with the cached item/occupancy flag, and the new regression covers `has_next()` followed by restore/reset and a subsequent `state_dict()`. This resolves the stale-checkpoint issue I raised.

aswanth-07:
Hi @divyanshk, the stale checkpoint case raised during review has now been fixed and the updated regression is approved. The branch is mergeable, and the Meta import checks are green on the current head. Could you take a final look when you have time?

PR #1554, opened 2026-09-14
Cut the nightly branch in this repo instead of from test-infra

atalman:
Our nightly branch is cut by pytorch/test-infra's trigger_nightly.yml, which runs .github/actions/trigger-nightly as pytorchbot. On 2026-09-10 pytorchbot stopped being able to write to the meta-pytorch org: in that day's run all six pytorch/* domains cut fine and all eight resolving to meta-pytorch/* failed with git exit 128. Our nightly branch has not moved since 2026-09-09T11:37:02Z and no wheels have been built since.

Move the cut here, where GITHUB_TOKEN already has the access. cut-nightly runs the same commit-tree/force-push as the test-infra action, so the nightly branch keeps being a real branch carrying main's tree under a dated commit -- unchanged for anything that consumes it.

A push made with GITHUB_TOKEN does not start a new workflow run, so the existing `push: branches: [nightly]` trigger cannot be what builds this. The build therefore runs in the same run as the cut, with ref: nightly so it builds exactly the commit just pushed. That trigger stays for manual pushes to the branch.

trigger-event is already github.event_name, and test-infra's _binary_upload.yml treats `schedule` as a first-class nightly path -- same pytorchbot-env, same gha_workflow_nightly_build_wheels OIDC role, same upload gate as a refs/heads/nightly push -- so the upload side needs no change. The matrix has said meta-pytorch/data since 202573b1cd and uploaded under that OIDC sub daily through 2026-09-09, so the role trust is already correct.

generate-matrix and build gain `if: !failure() && !cancelled()` because cut-nightly is skipped on every event except schedule, and a skipped `needs` would otherwise skip them too.


NicolasHug (approved):
stamping

